PR #11, opened 2023-09-06
Sof 1572/clean up from event system changes

KvelaGorrrrnio:
Clean up from setting up new event system. 
This includes moving most of the event system to the new module `event-system` and move separate abstractions (`interfaces`) and implementations (`services`, `factories`, `parsers`, ...) out into their own folders in each module.


KvelaGorrrrnio:
A new PR will be opened in this place.

PR #12, opened 2023-09-11
Sof 1593/trim files in core module

KvelaGorrrrnio:
- Moved event handling logic out into the new module `event-sytem`.
- Removed Zod schema files. These are merged into their respective parser classes.
- Normalized file suffix with `.service` and removed `.interface` suffix.


src/app/core/abstractions/entity-parser.service.ts
```diff
@@ -5,7 +5,7 @@ import { Part } from '../models/part'
 import { Piece } from '../models/piece'
 import { AdLibPiece } from '../models/ad-lib-piece'
 
-export abstract class EntityParser {
+export abstract class EntityParserService {
```
LindvedKrvang, line 8: I though this was the kind of naming you didn't want? Like you don't want `xHandlerService` etc?

src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -0,0 +1,75 @@
...
+import { AdLibPiece } from '../models/ad-lib-piece'
+import { BasicRundown } from '../models/basic-rundown'
+import { Part } from '../models/part'
+import { Piece } from '../models/piece'
+import { Rundown } from '../models/rundown'
+import { Segment } from '../models/segment'
+import * as zod from 'zod'
+
+
+export class ZodEntityParser implements EntityParserService {
```
LindvedKrvang, line 11: `EntityParserService` is an abstract class, so you should `extend` it right? Not `implements` it?

src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -0,0 +1,75 @@
...
+    private PIECE_PARSER = zod.object({
```
LindvedKrvang, line 12: These are instance variables. You are using are naming convention for constant values that are declared outside of the class. Please use our naming convention for instance variables or move them out to be constants.

src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -0,0 +1,75 @@
...
+        id: zod.string().nonempty(),
+        partId: zod.string().nonempty(),
+        name: zod.string().nonempty(),
+        layer: zod.string().nonempty(),
+    })
+    public parsePiece(piece: unknown): Piece {
```
LindvedKrvang, line 18: You are missing new lines before nearly all of your public methods in this file.

src/app/core/parsers/zod-entity-parser.service.ts
LindvedKrvang: Sidenote: Instance variables should be declared as the first thing of the class. Not througout the class

src/app/core/parsers/zod-rundown-event-parser.service.ts
```diff
@@ -0,0 +1,97 @@
...
+    PartSetAsNextEvent,
+    PartTakenEvent, RundownAdLibPieceInsertedEvent, RundownDeletedEvent
+} from '../models/rundown-event'
+import * as zod from 'zod'
+import { RundownEventType } from '../models/rundown-event-type'
+import { EntityParserService } from '../abstractions/entity-parser.service'
+import { AdLibPiece } from '../models/ad-lib-piece'
+
+@Injectable()
+export class ZodRundownEventParser {
```
LindvedKrvang, line 16: Same comments in this file as for `zod-entiry-parser.service`

src/app/core/abstractions/entity-parser.service.ts
KvelaGorrrrnio: Correct :D I must have accidentally have accepted the rename prompt in WebStorm :grin:
I have changed it back.

src/app/core/parsers/zod-entity-parser.service.ts
KvelaGorrrrnio: Both suggestions are applied. Both for `ZodEntityParser` and `ZodRundownEventParser`.

src/app/core/parsers/zod-entity-parser.service.ts
KvelaGorrrrnio: The only reason it is an abstract class instead of an interface is to have a runtime representation of the interface.

src/app/core/parsers/zod-entity-parser.service.ts
KvelaGorrrrnio: There are now newlines between all methods in both files.

src/app/core/parsers/zod-rundown-event-parser.service.ts
KvelaGorrrrnio: Same solutions as for `zod-entity-parser.service` are applied.

src/app/core/parsers/zod-entity-parser.service.ts
LindvedKrvang: Doesn't matter why it's an abstract class. It's still an abstract class ;)

LindvedKrvang (approved):
I don't like that you still `implements` the abstract class instead of `extends` it.
Doesn't matter why it's an abstract class. At the end of the day it's and abstract class and should be treated as such.

RasmusAlbrektsen (approved):

PR #13, opened 2023-09-11
Sof 1593/change unsubscribe lambdas to subscription objects

KvelaGorrrrnio:
- Converts `Unsubscribe` lambda to `EventSubscription` object.
- Corrects context binding for parser method usage.
- Corrected handling of deactivation of rundown in BasicRundownStateService.

### DIFF
--- a/src/app/core/services/basic-rundown-state.service.ts
+++ b/src/app/core/services/basic-rundown-state.service.ts
@@ -68,7 +68,7 @@ export class BasicRundownStateService implements OnDestroy {
   private deactivateBasicRundownFromEvent(event: RundownEvent): void {
     this.basicRundowns = this.basicRundowns.map(basicRundown => {
       if (basicRundown.id === event.rundownId) {
-        basicRundown.isActive = true
+        basicRundown.isActive = false
       }
       return basicRundown
     })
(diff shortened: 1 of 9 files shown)


LindvedKrvang (approved):

andr9528 (approved):

src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -11,28 +11,28 @@ import { RundownEvent } from '../models/rundown-event'
...
 
-  private registerEventConsumers(): void {
-    const unsubscribeFromConnectionStatusEvents = this.registerConnectionStatusConsumers()
-    const unsubscribesFromRundownEvents = this.registerBasicRundownEventConsumers()
-    this.unsubscribeFromEvents = () => [
-        ...unsubscribesFromRundownEvents,
-        ...unsubscribeFromConnectionStatusEvents
-      ].forEach(unsubscribe => unsubscribe())
+  private subscribeToEvents(): void {
+    const connectionStatusSubscriptions = this.subscribeToConnectionStatus()
```
RasmusAlbrektsen, line 27: Should these not be singular? Don't you just subscribe a single time to a pipeline of events? The name indicates that the constant holds multiple subscriptions.  Same question for the one underneath, and `subscribeToEvents()` in the `rundown-state.service.ts`.

RasmusAlbrektsen (approved):
Looks good, just a single question :-)

src/app/core/services/basic-rundown-state.service.ts
KvelaGorrrrnio: You are correct that currently `subscribeToConnectionStatus` returns a list of subscriptions with a length of 1. `subscribeToRundownEvents` returns a list with more than one subscription – one for each type of event that it subscribes to. One could wrap this in an EventSubscription object, to have a single object to unsubscribe from all subscriptions with a "single" call.

PR #14, opened 2023-09-11
Sof 1593/show notification on connection lost

KvelaGorrrrnio:
Displays a snackbar on connection loss and on re-established connection.

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -1,6 +1,8 @@
 import { TestBed } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
 import { AppComponent } from './app.component';
+import { ConnectionErrorService } from './shared/services/connection-error.service'
+import { instance, mock } from '@typestrong/ts-mockito'
 
 describe('AppComponent', () => {
   beforeEach(async () => {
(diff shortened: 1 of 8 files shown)


src/app/app.component.spec.ts
```diff
@@ -9,14 +11,31 @@ describe('AppComponent', () => {
...
+  it('should create the app', async () => {
+    await configureTestBed()
+    const fixture = TestBed.createComponent(AppComponent)
+    const app = fixture.componentInstance
     expect(app).toBeTruthy();
-  });
-});
+  })
+})
+async function configureTestBed(): Promise<void> {
```
LindvedKrvang, line 26: Would be nice with a new line :)

src/app/core/services/connection-status-observer.service.ts
```diff
@@ -25,6 +25,10 @@ export class ConnectionStatusObserver implements OnDestroy {
         })
     }
 
+    public subscribeToClosed(consumer: () => void): EventSubscription {
```
LindvedKrvang, line 28: This might just be my interpreatation of `consumer`, but I understand "consumer" as the class that wants to consume/use the EventObserver and that the Consumer has an action to be executed. I don't interpret "consumer" as the action/callback to be executed.
What you are parsing here is that "action"/callback, so in my mind `consumer` is not the right word.

You could call the parameter for `callback`, but that is still vauge. Since you method is called `subscribeToClosed` (which postfixing "event" to wouldn't hurt), you could call the parameter for `onClosed`. 
Then your method would look like this:
```
public subscribeToClosedEvent(onClosed: () => void): EventSubScription {
  return this.eventObserver.subscribe(ConnectionEventType.CLOSED, onClosed)
}
```
This makes it quite clear for the "consumer" of `ConnectionStatusObserver` what the consumer needs to parse along to this method.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

andr9528 (approved):

src/app/core/services/connection-status-observer.service.ts
KvelaGorrrrnio: I think that it might be a difference in which paradigm we were "raised". Personally, I see a function as an object for composition and hereby a worthy "consumer" of events. I can get behind the `onClosed` naming scheme.
I have applied it for `RundownEventObserver` and `ConnectionStatusObserver`.

src/app/app.component.spec.ts
KvelaGorrrrnio: Newline added.

LindvedKrvang (approved):

PR #15, opened 2023-09-11
Display message if no rundowns are available or if they are loading

KvelaGorrrrnio:
The Rundown Overview component now displays a message indicating that no rundowns are available – if it us unable to retrieve rundowns or if there simply are none.
It also listens for when the BasicRundownStateService is loading/fetching data from remote in order to keep the user informed.

The styling basically a centered heading and a description for both scenarios.

### DIFF
--- a/src/app/core/services/basic-rundown-state.service.ts
+++ b/src/app/core/services/basic-rundown-state.service.ts
@@ -13,12 +13,16 @@ export class BasicRundownStateService implements OnDestroy {
   private basicRundowns: BasicRundown[] = []
   private eventSubscriptions: EventSubscription[]
 
+  private readonly isLoadingSubject: BehaviorSubject<boolean>
+  private isLoading = true
+
   constructor(
       private readonly basicRundownService: BasicRundownService,
       private readonly rundownEventObserver: RundownEventObserver,
       private readonly connectionStatusObserver: ConnectionStatusObserver
   ) {
     this.basicRundownsSubject = new BehaviorSubject<BasicRundown[]>(this.basicRundowns)
+    this.isLoadingSubject = new BehaviorSubject<boolean>(this.isLoading)
     this.subscribeToEvents()
     this.resetBasicRundownSubject()
   }
@@ -39,10 +43,17 @@ export class BasicRundownStateService implements OnDestroy {
   }
 
   private resetBasicRundownSubject(): void {
+    this.setIsLoading(true)
     this.fetchBasicRundowns()
         .then(basicRundowns => this.basicRundowns = basicRundowns)
         .then(() => this.basicRundownsSubject.next(this.basicRundowns))
         .catch(error => console.error('[error]', 'Encountered error while fetching basic rundowns:', error))
+        .finally(() => this.setIsLoading(false))
+  }
+
+  private setIsLoading(isLoading: boolean): void {
+    this.isLoading = isLoading
+    this.isLoadingSubject.next(this.isLoading)
   }
 
   private subscribeToRundownEvents(): EventSubscription[] {
@@ -82,6 +93,10 @@ export class BasicRundownStateService implements OnDestroy {
 
   public subscribeToBasicRundowns(consumer: (basicRundowns: BasicRundown[]) => void): SubscriptionLike {
     return this.basicRundownsSubject.subscribe(consumer)
+  }
+
+  public subscribeToLoading(consumer: (isLoading: boolean) => void): SubscriptionLike {
+    return this.isLoadingSubject.subscribe(consumer)
   }
 
   private fetchBasicRundowns(): Promise<BasicRundown[]> {
--- a/src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
+++ b/src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
@@ -42,7 +42,7 @@
   }
 }
 
-.c-rundown-overview-title {
+.c-rundown-overview__title {
   margin: 1em;
   font-size: 3em;
   font-weight: lighter;
--- a/src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
+++ b/src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
@@ -42,6 +42,13 @@
   }
 }
 
+.c-rundown-overview {
+  &__no-rundowns,
+  &__loading-rundowns {
+    text-align: center;
+  }
+}
+
 .c-rundown-overview__title {
   margin: 1em;
   font-size: 3em;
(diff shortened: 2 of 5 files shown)


src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -84,6 +95,10 @@ export class BasicRundownStateService implements OnDestroy {
     return this.basicRundownsSubject.subscribe(consumer)
   }
 
+  public subscribeToLoading(consumer: (isLoading: boolean) => void): SubscriptionLike {
```
LindvedKrvang, line 98: I think it would be a good idea that you follow the same naming convention that you did for all your other "consumer" parameters. So this parameter would be called `onLoad`

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
```diff
@@ -42,7 +42,14 @@
...
 
-.c-rundown-overview-title {
+.c-rundown-overview {
+  &__no-rundowns,
+  &__loading-rundowns {
+    text-align: center;
+  }
+}
+
+.c-rundown-overview__title {
```
LindvedKrvang, line 52: I must admit, I don't think we gain much value from using BEMIT.
`c-rundown-overview__` is redundant to add since Angular allready uses ViewEncapsulation to scope the CSS.
All CSS classes defined in `rundown-overview.component.css` will only be applied in html files that has been told to use this specific file. Currently, that is only the `rundown-overview.component.html`.

So we add a lot of "clutter" to our CSS names that we don't gain a benefit from.
https://angular.io/guide/view-encapsulation

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.ts
```diff
@@ -7,15 +7,17 @@ import { Rundown } from '../../../core/models/rundown';
...
   templateUrl: './rundown-overview.component.html',
   styleUrls: ['./rundown-overview.component.scss']
 })
-export class RundownOverviewComponent implements OnInit {
-
-  public basicRundowns: BasicRundown[]
+export class RundownOverviewComponent implements OnInit, OnDestroy {
+  public basicRundowns: BasicRundown[] = []
+  public isLoading: boolean = true
+  private subscriptions: SubscriptionLike[] = []
```
LindvedKrvang, line 20: I have noticed you have also used `SubscriptionLike` somewhere else in another PR.
What's the reason for not using `Subscription`?

If it's because you just want the unsubscripe functionality and nothing else, then what's the reason for not using `Unsubscribable` which is where `SubscriptionLike` gets its unsubscribe from?

I'm in favour of just using `Subscription`. It keeps it "logical" simple :)

src/app/core/services/basic-rundown-state.service.ts
LindvedKrvang: A second read through and I realise that your Subject doesn't only emit on load, so my naming suggestion would not be accurate.
I do however, still prefer that we don't call it `consumer` :)

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
RasmusAlbrektsen: We need to make a decision for what we do here before we go too far in one direction, or stray in to many different ones. 
I can agree that the BEMIT convention might not have as much purpose in Angular as it would in other scenarios.

andr9528 (approved):

src/app/core/services/basic-rundown-state.service.ts
KvelaGorrrrnio: I also thought about changing it so `onX`. I just couldn't find a proper name, and thought it was a bit overkill to split it into two subscriptions.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.ts
KvelaGorrrrnio: `Subscription` is an implementation. That would mean that everything that we try to put in the subscriptions must implement all public and private methods and properties that `Subscription` has. The reason I went with SubscriptionLike is that it has the `closed: boolean` property. Such that the consumer of the subscription can see if the connection is live. If we don't care about that, then I am for changing it to use `Unsubscribable`.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
KvelaGorrrrnio: I do see your point. And it might not be that it is as needed in Angular as it is for projects that do not use Shadow DOM. But the structure in the class names still give useful information. By looking at this single line, I have a lot of information about what the class is describing. I know that we are in the scope of the Rundown overview component and that we are targeting the title. I don't even have to look at the filename.

When debugging and going through the DOM, then these classes are also useful, in the same sense that the context is given. If the class was simply `.title` and I were to search for that in the entire project, there would be a lot to filter out before finding the correct one. So again, it informs the structure and context to the reader.

Another argument is that if we use BEMIT throughout our products, then we will have a structure to follow when writing SCSS/CSS. As you mentioned, `the right tool for the right job`, which is correct, but as I see it, the pros of having a structure/standard for how we do styling outweighs the possible redundancy it might bring for the minority of our products.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.ts
LindvedKrvang: But you are populating it with actual implemetentations of  `Subscription`.

> That would mean that everything that we try to put in the subscriptions must implement all public and private methods and properties that Subscription has. 

If you find yourself manually creating your own subscriptions, then you are doing something wrong. The `Subscription` is created for us by `Subject.subscribe()`.

Semantically, it makes more sense to work on Subscriptions than something that is like Subscriptions.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
LindvedKrvang: I do like the fact that we can go from one project to another and then have a good idea of what to expect.
We just need to be carefull with that argument because that quickly becomes a slippery slope to use something that is not a good choice for a given project, because it's a good choice for three different projects.

RasmusAlbrektsen: I like the point about specifically naming separation within the file as well as the ease of finding things when going through the DOM.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.ts
KvelaGorrrrnio: We are still using `ManagedSubscription`, which is currently implementing SubscriptionLike and not Subscription.

PR #16, opened 2023-09-14
Sof 1529/Create Rundown View

RasmusAlbrektsen:
_**Note:** Is depending on https://github.com/tv2/sofie-server/pull/6._
This PR creates the general layout for the Rundown view and it's components. The rundown view is by no means done. At the bottom we have written down some of the tasks that should be focused on next! 

At the moment this PR includes:
- Visualized Segments, Parts and Pieces with Sofie-styling
- A timeline that can be animated
- An On air playhead
- A Next-cursor that is made from a border that shows where the next cursor can be put. 
- A Piece layer service that can be used to determine which layers are required for a part.

Current status is:
Skærmbillede 2023-09-20 kl  14 02 57

Some of the outstanding tasks are:
Pieces
- Piece colors, have we accounted for all scenarios (also, the way they are chosen should be optimized)
- DVE colors, at the moment we do not account for how the colors depict which kind of DVE we employ.
- Live speak (should we call this “Voice over” instead?“), when we have a piece with both video and sound, these colors are not accounted for, like Mosart_ur, and we instead just put ont he Live speak color.
- “Piece-layer type” should be renamed to something more appropriate.

Parts
- Invalid parts are shown in the UI as an empty part and can be taken.
- Autonexted parts shows wrong duration.

Segment
- Editorial line
- LIVE color on segments that have live parts. 
- Last part next/autonext indicator
- More view modes that are controller by specific components designed for that purpose.

Rundown
- At the moment, the screen does not shift focus towards the On Air-playhead. If the playhead moves out of the screen, we should move the window focus towards the segment that has the playhead on it.
- Make the rundown view reset on activate, currently, activating the rundown does not reset the segments.

Other:
- Extract colors out into (s)css variables.
- No text select for rundown view.


src/app/rundown/components/rundown/rundown.component.html
```diff
@@ -8,7 +8,7 @@
       <button *ngIf="rundown?.isActive" (click)="deactivateRundown()">Deactivate Rundown</button>
     </div>
     <div class="segments">
-      <sofie-segment *ngFor="let segment of rundown?.segments" [segment]="segment" [isRundownActive]="rundown?.isActive ?? false" (setNextEvent)="setNext($event)"></sofie-segment>
+      <sofie-segment *ngFor="let segment of rundown?.segments" [segment]="segment" [isOnAir]="segment.isOnAir" [isRundownActive]="rundown?.isActive ?? false" (setNextEvent)="setNext($event)"></sofie-segment>
```
LindvedKrvang, line 11: This seems redundant:
`[segment]="segment" [isOnAir]="segment.isOnAir"`
since you are passing the same value twice.

I assume this has something to do with you wanting to only listen on `isOnAir` changes?

src/app/rundown/components/part/part.component.ts
```diff
@@ -23,20 +27,70 @@ import {animate, state, style, transition, trigger} from '@angular/animations';
...
+  public time: number
+
   @Input()
   public part: Part
 
+  @Input()
+  public pieceLayers: string[]
+
+  @Input()
+  public pixelsPerSecond: number
```
LindvedKrvang, line 49: This is starting to become a lot of inputs for a component.

Same applies for some of the other components

src/app/rundown/components/segment/segment.component.html
```diff
@@ -1,4 +1,44 @@
...
+        {{ pieceLayer }}
+      </div>
+    </div>
+  </div>
+
+  <!-- TODO: Should this be a separate component, such that we can easily swap it out. E.g. if we want to follow playhead or show full overview. -->
+  <div class="c-segment__timeline">
+    <sofie-timeline [ngClass]="'c-segment__timeline-markers'" [time]="timeReference" [pixelsPerSecond]="30"></sofie-timeline>
+    <div class="c-segment__timeline-parts">
+      <div class="c-segment__timeline-parts-wrapper">
```
LindvedKrvang, line 26: Do you really need both these divs?
The first does literally nothing but add padding.

src/app/rundown/components/segment/segment.component.html
```diff
@@ -1,4 +1,44 @@
...
+  </div>
+
+  <div class="c-segment__controls">
+    <div class="c-segment__controls-toolbar">Hello</div>
+    <div class="c-segment__piece-layers">
+      <div class="c-segment__piece-layer" *ngFor="let pieceLayer of pieceLayers">
+        {{ pieceLayer }}
+      </div>
+    </div>
+  </div>
```
LindvedKrvang, line 20: Do all these need to be divs?
Seems like `<div class="c-segment__controls-toolbar">Hello</div>` could be `<span class="c-segment__controls-toolbar">Hello</span>`.
and `<div class="c-segment__piece-layers">` seems to do nothing, so why is it there?

There seem to be a tendency to use way too many divs in general.

src/app/rundown/components/segment/segment.component.ts
```diff
@@ -1,25 +1,124 @@
...
+    const usedPieceLayers = this.pieceLayerService.getPieceLayersForParts(this.segment.parts)
+    const availablePieceLayersInOrder: string [] = [
+      'OVERLAY',
+      'PGM',
+      'JINGLE',
+      'MUSIK',
+      'MANUS',
+      'ADLIB',
+      'SEC',
+      'AUX',
```
LindvedKrvang, line 56: Since we currently don't plan on making these configurable, we should move them to a const or an enum.

src/app/shared/services/piece-layer.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    public getPieceLayersForParts(parts: Part[]): string[] {
+        const pieceLayers: string[] = []
+        parts.forEach(part => {
+            part.pieces.concat(part.adLibPieces).forEach(piece => {
+                const layer = this.getPieceLayer(piece)
+                if (!pieceLayers.includes(layer) && layer !== 'UNKNOWN') {
+                    pieceLayers.push(layer)
+                }
+            })
+        })
```
LindvedKrvang, line 14: This isn't tested but this should return the same without the nested `foreach` loop:
```
const duplicatedLayers: string[] = parts.flatMap(part => {
            return part.pieces.concat(part.adLibPieces)
                .map(piece => this.getPieceLayer(piece))
                .filter(layer => layer !== 'UNKNOWN')
        })
const set: Set<string> = new Set(duplicatedLayers)
return Array.from(set)
```
We should try to limit the use of `foreach` and use some of the other methods available on Array instead. It's a big flag that the `foreach` are nested.

If you always want the returned array to contain unique entries, then consider just changing the return type to a Set.

LindvedKrvang (commented):
Since this is still a work in progress there are some stuff I deliberately didn't comment on.

However, it seems you have a tendency to use a lot of unnecessary divs. This should be looked in to.

src/app/rundown/components/segment/segment.component.ts
andr9528: I vote for an Enum :)

src/app/shared/pipes/timestamp.pipe.ts
```diff
@@ -0,0 +1,18 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'timestamp'
+})
+export class TimestampPipe implements PipeTransform {
+  transform(timestampInSeconds: number,): string {
```
andr9528, line 7: I know this method is public, due to there being no access modifier, and it being the implementation of the interface, but could we please add access modifiers explicitely?

src/app/shared/services/piece-layer.service.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        return pieceLayers
+    }
+
+    //TODO: Does not account for the fact that a JINGLE should be PGM when in the KLAR ON AIR (if this is desired behaviour?)
+    public getPieceLayer(piece: Piece): string {
```
andr9528, line 20: I can already tell that ones we add the `complexity` linting rule, with a max of 10, that this method is going to error out.
Is there some way it could be split down into smaller methods, to reduce the cyclomatic complexity?

src/app/rundown/components/rundown/rundown.component.html
KvelaGorrrrnio: This was a "hack" to let the component know about changes when a segment is putOnAir or is takenOffAir.
Finding a better data patch strategy might remove the need for it. I'll investigate how we can go about this.

src/app/shared/services/piece-layer.service.ts
KvelaGorrrrnio: I can tell that the cyclomatic complexity of this method is 7, so it is within the desired threshold.
And I can contribute with the knowledge that this entire method will be rewritten completely when `piece.type` is being populated in the backend.

src/app/rundown/components/segment/segment.component.html
KvelaGorrrrnio: `.c-segment__controls-toolbar` should be a blocking element and will contain zoom/scale controls.
`.c-segment__piece-layers` is there to regulate the space where the layers are confined in the same way as `ol` or `ul` do for `li`.

src/app/rundown/components/part/part.component.html
```diff
@@ -1,17 +1,20 @@
...
   </span>
-  {{part.id}}
-  <button *ngIf="isRundownActive && !part.isNext && !part.isOnAir" (click)="setPartAsNext()">Set Part as Next</button>
-  <div class="piece-container">
-    <sofie-piece *ngFor="let piece of part.pieces" [piece]="piece"></sofie-piece>
-    <sofie-ad-lib-piece *ngFor="let adLibPiece of part.adLibPieces" [adLibPiece]="adLibPiece"></sofie-ad-lib-piece>
+  <button *ngIf="isRundownActive && !part.isNext && !part.isOnAir" (click)="setPartAsNext()">Set Part as Next</button>-->
+  <div class="c-part__piece-layers">
+    <div class="c-part__piece-layer" [style.width]="partWidthInPixels" *ngFor="let pieces of layeredPieces">
+      <sofie-piece *ngFor="let piece of pieces" [maxDuration]="duration" [piece]="piece" [pixelsPerSecond]="pixelsPerSecond"></sofie-piece>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: `ngX` attributes are now the first ones on elements.

src/app/rundown/components/part/part.component.ts
KvelaGorrrrnio: Yes.

src/app/rundown/components/segment/segment.component.html
KvelaGorrrrnio: `.c-segment__timeline-parts-wrapper` was intended for organizing the part content, where `.-segment__timeline-parts` was the scroll-container.

Since we don't scroll at the moment, the wrapper div is removed.

src/app/rundown/components/segment/segment.component.ts
KvelaGorrrrnio: This has been extracted out into an enum and a method for setting the order.

src/app/shared/services/piece-layer.service.ts
KvelaGorrrrnio: This has been rewritten to use a functional approach:
```typescript
    public getPieceLayersForParts(parts: Part[]): PieceLayer[] {
        const pieces: Piece[] = parts.flatMap(part => part.pieces.concat(part.adLibPieces))
        const pieceLayers: PieceLayer[] = pieces.map(piece => this.getPieceLayer(piece))
        return [...new Set(pieceLayers)]
    }
```

src/app/shared/pipes/timestamp.pipe.ts
KvelaGorrrrnio: Access modifier is made explicit and the extra comma has been removed.

src/app/core/models/part.ts
```diff
@@ -45,4 +60,19 @@ export class Part {
   public insertAdLibPiece(adLibPiece: AdLibPiece): void {
     this.adLibPieces.push(adLibPiece)
   }
+
+  public getDuration(): number {
+    const expectedDuration = this.expectedDuration ?? 4000
```
andr9528, line 65: Any particular reason for it to default to 4000 (frames?, ms?, seconds?)?

src/app/core/models/rundown.ts
```diff
@@ -26,40 +26,63 @@ export class Rundown {
...
+    segmentToComeOnAir.putOnAir(rundownCursor.partId, timestamp)
+  }
+
+  private takeOnAirSegmentOffAirIfIsNextSegment(nextSegmentId: string, timestamp: number): void {
+    const activeSegment = this.segments.find(segment => segment.isOnAir)
+    if (!activeSegment || activeSegment.id === nextSegmentId) {
+      return
+    }
+    activeSegment.takeOffAir(timestamp)
   }
```
andr9528, line 71: Something about what this method does compared to its name doesn't line up in my head.
It says it takes on-air segments off-air, if it is the next segment. However if it is the next segment it shouldn't be taken off-air, as the method correctly doesn't.

Am i misunderstanding something, or is the name missing a `Not`, i.e the name actually being `takeOnAirSegmentOffAirIfIsNotNextSegment`. 
Thats a rather long name for a pretty small method. 
Perhabs a complely new name need to be found, or the method refactored slightly?

andr9528: Oh and also adding a type to `activeSegment` would be nice. 
It is after all checked for `undefined`, due to comming from a `.find`, which can return `undefined`.

src/app/core/models/rundown.ts
```diff
@@ -26,40 +26,63 @@ export class Rundown {
...
+    if (this.isOffAirSegmentSetAsNext(newNextSegment)) {
+      newNextSegment!.reset()
+    }
+    oldNextSegment?.removeAsNextSegment()
+    newNextSegment?.setAsNextSegment(rundownCursor.partId)
+  }
+
+  private isOffAirSegmentSetAsNext(segment: Segment | undefined): boolean {
+    return segment === undefined || !segment.isOnAir
   }
```
andr9528, line 86: Am i understanding this correctly?
If the `segment` supplied is `undefined` then `undefined` is set as next segment, meaning we are on the last segment?

src/app/rundown/components/piece/piece.component.ts
```diff
@@ -11,5 +11,82 @@ export class PieceComponent {
...
+  @Input()
+  public pixelsPerSecond: number
+
+  @Input()
+  public partDuration: number
+
+  @Input()
+  public expectedPartDuration: number
+
+  public DEFAULT_PIECE_DURATION: number = 4000
```
andr9528, line 23: Would be nice to have some indication for what unit is used. 
Is it 4000 frames? ms? seconds?

As soon as it has been added to just one of the constants/variables uses in the calculations, then the unit can be inferred for the rest.

src/app/rundown/components/rundown/rundown.component.html
```diff
@@ -8,18 +8,23 @@
...
+<!--  <div>-->
+<!--    <div class="infinite-pieces">-->
+<!--      <span class=subtitle>Infinite Pieces</span>-->
+<!--      <sofie-piece *ngFor="let infinitePiece of rundown?.getInfinitePieces()" [piece]="infinitePiece"></sofie-piece>-->
+<!--    </div>-->
+<!--    <div class="ad-lib-pieces">-->
+<!--      <span class=subtitle>Adlib Pieces</span>-->
+<!--      <sofie-ad-lib-piece-identifier *ngFor="let identifier of adLibPieceIdentifiers" [adLibPieceIdentifier]="identifier" [rundownId]="rundown?.id ?? ''"></sofie-ad-lib-piece-identifier>-->
+<!--    </div>-->
+<!--  </div>-->
```
andr9528, line 28: Is this going to be used in the future, or just leftover code?
If suppose to be used later, then maybe add a comment detailing that?

src/app/core/models/part.ts
KvelaGorrrrnio: I have extracted the value out into the constant `DEFAULT_PART_DURATION`. Regarding the units, I think it makes sense that the default unit is milliseconds.

src/app/rundown/components/timeline-playhead/timeline-playhead.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  @Input()
+  public time: number
+
+  @Input()
+  public pixelsPerSecond: number
+
+  @HostBinding('style.left.px')
+  get left() {
+    return Math.floor(this.time * this.pixelsPerSecond / 1000)
+    // return '50px'
```
andr9528, line 19: Leftover code?

src/app/core/models/rundown.ts
KvelaGorrrrnio: Good observation. You are correct. I have changed the method name to `takeOnAirSegmentOffAirIfNotSetAsNext`.

src/app/rundown/components/timeline-playhead/timeline-playhead.component.ts
KvelaGorrrrnio: Correct. It is removed now.

src/app/core/models/rundown.ts
KvelaGorrrrnio: I have removed the method. Instead `!newNextSegment?.isOnAir` is used instead.

src/app/shared/pipes/timestamp.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+  public transform(timestampInSeconds: number): string {
+    const seconds: number = timestampInSeconds % 60
+    let minutes: number = Math.floor(timestampInSeconds / 60)
+    let hours: number = 0
+    if (minutes >= 60) {
+      hours = Math.floor(minutes / 60)
+      minutes = minutes % 60
+    }
+
+    return `${hours > 0 ? `${hours}:` : ''}${minutes < 10 ? `0${minutes}` : minutes}:${seconds < 10 ? `0${seconds}` : seconds}`
```
andr9528, line 16: This line can be simplified by doing the following.

```typescript
public transform(timestampInSeconds: number): string {
   ...
   return `${hours > 0 ? `${hours}:` : ''}${this.zeroPadStart(minutes)}:${this.zeroPadStart(seconds)}`
}

private zeroPadStart(nr: number): string {
   return nr.toString().padStart(2, '0')
 }
```

src/app/rundown/components/piece/piece.component.ts
andr9528: I take that the unit is the same as for a `Part`, i.e ms?
If so, fell free to mark this one resolved.

src/app/shared/pipes/timestamp.pipe.ts
KvelaGorrrrnio: Jeg har opdateret klassen til at bruge `.padStart`, men også selve strukturen for formatteringen.

andr9528: What are you saying? :D

src/app/rundown/components/rundown/rundown.component.html
andr9528: Great to see it gone.

src/app/shared/pipes/timestamp.pipe.ts
KvelaGorrrrnio: Sorry :grin: I have updated the class to use `.padStart`, but also how the formatting is applied.

src/app/rundown/components/piece/piece.component.ts
KvelaGorrrrnio: I have taken your advice and suffixed `_IN_MS` for the Part constant. This one has been removed since it wasn't necessary.

src/app/shared/services/piece-layer.service.ts
LindvedKrvang: Followup question: 
Why not let the return type be `Set<PieceLayer>` instead of `PieceLayer[]` if it's a requirement that there are no duplicates?

src/app/rundown/components/piece/piece.component.ts
```diff
@@ -11,5 +11,76 @@ export class PieceComponent {
...
+
+  public getPieceTypeModifierClass(): string {
+    const pieceType = this.getPieceType()
+    return `piece--${pieceType}`
+  }
+
+  // TODO: Use either piece.type or source layer types for determining style.
+  public getPieceType(): string {
+    const layerType = this.piece.layer.replace(/^studio\d+_/, '')
+    switch (layerType) {
```
LindvedKrvang, line 50: Piece.type was intented to be used for this

src/app/rundown/components/piece/piece.component.ts
```diff
@@ -11,5 +11,76 @@ export class PieceComponent {
...
+      case 'selected_voiceover':
+        return 'live-speak'
+      case 'local':
+        return 'local'
+      case 'dve':
+      case 'dve_adlib':
+        return 'dve'
+      default:
+        return 'unknown'
+    }
```
LindvedKrvang, line 84: This needs to be something else than magic strings.
Also, I don't think sourceLayers should be used for this.

src/app/rundown/components/rundown/rundown.component.html
```diff
@@ -8,17 +8,12 @@
       <button *ngIf="rundown?.isActive" (click)="deactivateRundown()">Deactivate Rundown</button>
```
LindvedKrvang, line 8: Would be nice if you changed this button grouping to:
```
<button *ngIf="!rundown.isActive; else activeRundownButtons" (click)="activeRundown()">...</button>
<ng-template #activeRundownButtons>
  <button> (click)="takeNext()">...</button>
  ...
</ng-template>
```

src/app/rundown/components/segment/segment.component.html
```diff
@@ -1,14 +1,29 @@
...
-  </span>
-  <div class="parts">
-    <sofie-part *ngFor="let part of segment.parts" [part]="part" [isRundownActive]="isRundownActive" (partSelectedAsNextEvent)="emitSetNextEvent($event)"></sofie-part>
+<div class="c-segment" [class.c-segment--on-air]="segment.isOnAir" [class.c-segment--set-as-next]="segment.isNext">
+  <div class="c-segment__header">
+    <div class="c-segment__header-toolbar">
+      <span>123</span>
+      <span style="margin-left: 50px;">00:30</span>
+    </div>
+    <div class="c-segment__title">
```
LindvedKrvang, line 7: Any reason for this not being a span?

src/app/rundown/components/segment/segment.component.ts
```diff
@@ -1,25 +1,103 @@
...
   @Input()
-  public isRundownActive: boolean
+  public isOnAir: boolean
 
   @Input()
   public segment: Segment
 
   @Output()
-  setNextEvent: EventEmitter<{segmentId: string, partId: string}> = new EventEmitter()
+  public setNextEvent: EventEmitter<{segmentId: string, partId: string}> = new EventEmitter()
```
LindvedKrvang, line 29: This comment also applies for `setNextEvent` on Part.

All this do is propagate up to the RundownComponent so it can do a call to `RundownService`, where it doesn't need to know the response.

Why not just call the RundownService directly from the Part component?
We have a RundownStateService to handle the state, so we can allow our "dumb" component to make this call for us.
Saves a lot of boilerplate code ;)

src/app/rundown/components/segment/segment.component.ts
```diff
@@ -1,25 +1,103 @@
...
+    const pieceLayersInOrder: PieceLayer[] = this.pieceLayerService.getPieceLayersInOrder()
+    const usedPieceLayers: PieceLayer[] = this.pieceLayerService.getPieceLayersForParts(this.segment.parts)
+    return pieceLayersInOrder.filter(layer => usedPieceLayers.includes(layer))
+  }
+
+  public ngOnChanges(changes: SimpleChanges): void {
+    this.pieceLayers = this.getUsedPieceLayersInOrder()
+
+    const isOnAirChange: SimpleChange | undefined = changes['isOnAir']
+    if (isOnAirChange?.currentValue === false) {
```
LindvedKrvang, line 52: This doesn't work `!isOnAirChange?.currentValue`?

src/app/rundown/components/timeline-markers/timeline-markers.component.ts
```diff
@@ -0,0 +1,190 @@
...
+    const timeOffsetInPixels = this.pixelsPerSecond * ((this.time / 1000) % this.secondsPerSection)
+    for (let i = 0; i < subsectionCount; i++) {
+      this.drawSubsectionIfNotCollidingWithSection(i, subsectionWidth, timeOffsetInPixels)
+    }
+  }
+
+  private drawSubsectionIfNotCollidingWithSection(subsectionIndex: number, subsectionWidth: number, timeOffsetInPixels: number): void {
+    if (subsectionIndex % this.subsectionsPerSection === 0) {
+      return
+    }
```
LindvedKrvang, line 84: You can get rid of the "if" part of this method by moving the responsibility of checking the if outside the method. Then the method only have one responsibility: Draw the section.

LindvedKrvang (approved):
Looks good.
Looking forward to when it's the timeline that "animate" while the playhead is "standing still" :)

src/app/rundown/components/full-segment-timeline/full-segment-timeline.component.html
```diff
@@ -0,0 +1,22 @@
+<sofie-timeline-markers
+    class="c-full-segment-timeline__markers"
+    [time]="0"
+    [pixelsPerSecond]="pixelsPerSecond"
+></sofie-timeline-markers>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: I can't. Custom HTML elements are required to have a closing tag.

src/app/rundown/components/full-segment-timeline/full-segment-timeline.component.html
```diff
@@ -0,0 +1,22 @@
...
+<div class="c-full-segment-timeline__parts">
+    <sofie-part
+            *ngFor="let part of segment.parts"
+            [part]="part"
+            [pieceLayers]="pieceLayers"
+            [pixelsPerSecond]="pixelsPerSecond"
+            [isOnAir]="part.isOnAir"
+            [time]="part.isOnAir ? time : 0"
+            (partSelectedAsNextEvent)="emitSetNextEvent($event)"
+    ></sofie-part>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: I can't. Custom HTML elements are required to have a closing tag.

src/app/rundown/components/full-segment-timeline/full-segment-timeline.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  templateUrl: './full-segment-timeline.component.html',
+  styleUrls: ['./full-segment-timeline.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class FullSegmentTimelineComponent implements AfterViewInit, OnChanges {
+  @Input()
+  public segment: Segment
+
+  @Input()
+  public pieceLayers: PieceLayer[]
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Currently it changes from segment to segment. Later it will change also change per segment on ingest changes.

src/app/rundown/components/full-segment-timeline/full-segment-timeline.component.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  @Input()
+  public time: number
+
+  @Output()
+  public setNextEvent: EventEmitter<{segmentId: string, partId: string}> = new EventEmitter()
```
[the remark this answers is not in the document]
KvelaGorrrrnio: I have changed all occurences of `{ segmentId: string, partId: string }` with `RundownCursor`.

src/app/rundown/components/full-segment-timeline/full-segment-timeline.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  public ngAfterViewInit() {
+    this.pixelsPerSecond = this.getPixelsPerSecond()
+  }
+
+  public ngOnChanges(changes: SimpleChanges): void {
+    if (!this.segment.isOnAir) {
+      return
+    }
+
+    if (Date.now() - this.internalTimeReference < 200) {
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Correct. I have removed the scaling logic, so this is no longer present.

src/app/rundown/components/timeline-markers/timeline-markers.component.ts
KvelaGorrrrnio: I don't get rid of it, it just gets moved. I have dragged the logic for drawing the subsection out into `drawSubsectionFromIndex`.

src/app/shared/pipes/timestamp.pipe.ts
```diff
@@ -0,0 +1,29 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'timestamp'
+})
+export class TimestampPipe implements PipeTransform {
```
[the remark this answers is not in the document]
KvelaGorrrrnio: The closest I have found is `DatePipe`. It might be sufficient for our needs, but doesn't match what the this implementation does (allowing the hour count to go beyond 23).

src/app/rundown/components/segment/segment.component.ts
KvelaGorrrrnio: No. If there are no changes to the `isOnAir` property, then the animation should keep on playing.

src/app/rundown/components/segment/segment.component.ts
KvelaGorrrrnio: If we remove the boilerplate, we just need to pass down the `rundownId` instead to the PartComponent. Is that desirable?

src/app/rundown/components/segment/segment.component.html
KvelaGorrrrnio: Not really. We are just relying on the blocking nature of the div.
But the span can be made blocking as well. Do you want us to change that?

src/app/rundown/components/segment/segment.component.html
```diff
@@ -1,14 +1,29 @@
...
-  </span>
-  <span *ngIf="segment.isNext" class="next">
-    NEXT
-  </span>
-  <div class="parts">
-    <sofie-part *ngFor="let part of segment.parts" [part]="part" [isRundownActive]="isRundownActive" (partSelectedAsNextEvent)="emitSetNextEvent($event)"></sofie-part>
+<div class="c-segment" [class.c-segment--on-air]="segment.isOnAir" [class.c-segment--set-as-next]="segment.isNext">
+  <div class="c-segment__header">
+    <div class="c-segment__header-toolbar">
+      <span>123</span>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Just a placeholder for future timing values.

src/app/rundown/components/piece/piece.component.ts
KvelaGorrrrnio: Before the task is set to done, we will have addressed this. But it requires that piece.type is populated in the backend.

src/app/rundown/components/piece/piece.component.ts
KvelaGorrrrnio: Yes, but it is not yet populated in the backend. And we focused on the sprint-goal :wink:

src/app/rundown/components/piece/piece.component.html
```diff
@@ -1,3 +1,5 @@
-<div class="piece">
-  {{piece.name}}
+<div [ngClass]="'piece ' + getPieceTypeModifierClass()" [style.width]="getPieceWidthInPixels()" [title]="piece.name" [style.left]="getOffsetInPixels()">
+  <div class="piece__text" [ngClass]="{ 'piece__text--extend': partDuration > expectedPartDuration }">
```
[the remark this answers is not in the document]
KvelaGorrrrnio: No. It is now a span.

src/app/rundown/components/part/part.component.ts
```diff
@@ -1,42 +1,77 @@
...
-      transition('offAir => onAir', animate(500)),
-      transition('offAir => isNext', animate(500)),
-      transition('onAir => isNext', animate(500))
-    ]),
-  ]
+  styleUrls: ['./part.component.scss']
 })
-export class PartComponent {
+export class PartComponent implements OnChanges {
+  // TODO: Figure out how to "subscribe" to changes for this on the part property.
```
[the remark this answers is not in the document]
KvelaGorrrrnio: That is a solution, but I think it will be hard to track what happens where in the application.
I think the proper way to handle this is to do immutable "updates" to the entity state variables.

src/app/rundown/components/part/part.component.html
```diff
@@ -1,17 +1,12 @@
...
-  </span>
-  <span *ngIf="part.isNext" class="is-next">
-    NEXT
-  </span>
-  {{part.id}}
-  <button *ngIf="isRundownActive && !part.isNext && !part.isOnAir" (click)="setPartAsNext()">Set Part as Next</button>
-  <div class="piece-container">
-    <sofie-piece *ngFor="let piece of part.pieces" [piece]="piece"></sofie-piece>
-    <sofie-ad-lib-piece *ngFor="let adLibPiece of part.adLibPieces" [adLibPiece]="adLibPiece"></sofie-ad-lib-piece>
+  <div class="c-part__piece-layers">
```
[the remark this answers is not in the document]
KvelaGorrrrnio: It is not strictly necessary, but by having it, we separate the responsibilities of styling.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.html
```diff
@@ -1,15 +1,15 @@
...
 </div>
 
-<div class="c-rundown-overview__no-rundowns" *ngIf="!isLoading && basicRundowns.length === 0">
+<div *ngIf="!isLoading && basicRundowns.length === 0" class="c-rundown-overview__no-rundowns">
   <h2>There are currently no rundowns available.</h2>
   <p>Check your connection and if you have configured the Ingest Gateway.</p>
 </div>
 
-<table mat-table [dataSource]="basicRundowns" *ngIf="!isLoading && basicRundowns.length > 0">
+<table *ngIf="!isLoading && basicRundowns.length > 0" mat-table [dataSource]="basicRundowns">
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Why?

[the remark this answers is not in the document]
KvelaGorrrrnio: I am not that familiar with templates in angular. Unless they offer some performance-wise optimizations, I feel like this just brings more nesting and decreases the readability.

src/app/rundown/components/rundown/rundown.component.html
KvelaGorrrrnio: Updated.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.html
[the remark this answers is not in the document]
LindvedKrvang: It's a way, among others, to reduce the amount of `*ngIf` in your mark-up so you dont have:
```
<div *ngIf="someCondition" />
<div *ngIf="!someCondition" />
```
It's also a way to create reuseable "pieces" of mark-up within your component.

Note, by default Angular will not render a template without "being instructed" to render it.
https://angular.io/api/core/ng-template

> I feel like this just brings more nesting and decreases the readability.

I must admit I disagree on that one. Templates allow you to extract a piece of your mark-up just as you would extract a piece of your code into a new method.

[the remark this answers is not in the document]
LindvedKrvang: It's an old "design pattern" that was an easy way to display data before the advent of modern standards-based browsers and is now considered a bad practice in a lot of cases since they are more difficult to optimize for the requirements of a modern webpage (responsiveness, accessibility, SEO etc - not that we care for most of those things...)

src/app/rundown/components/full-segment-timeline/full-segment-timeline.component.html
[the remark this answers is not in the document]
LindvedKrvang: That... is a surprise to me. I'm almost positive I have done so many times before. But my compiler complains if I do it now, so you are right.

src/app/rundown/components/part/part.component.ts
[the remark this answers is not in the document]
LindvedKrvang: That also sounds like a viable solution :)

src/app/shared/services/piece-layer.service.ts
KvelaGorrrrnio: Now returns `Set<PieceLayer>`.

src/app/rundown/components/segment/segment.component.ts
LindvedKrvang: `Segment` aleady have the `rundownId` on it, so you would only need the input for rundownId on the Part componenet, so it's not that bad.

LindvedKrvang (approved):

andr9528 (approved):

KvelaGorrrrnio (approved):

PR #18, opened 2023-09-22
Part.getDuration takes autoNext.overlap into account.

KvelaGorrrrnio:
_**Note:** Depends on https://github.com/tv2/sofie-server/pull/11._

### DIFF
--- a/src/app/core/models/part.spec.ts
+++ b/src/app/core/models/part.spec.ts
@@ -86,7 +86,7 @@ describe(Part.name, () => {
                 describe('part has no expected duration', () => {
                     describe('default part duration is longer than the time since part was taken', () => {
                         it('returns the default part duration', () => {
-                            const playedDuration: number = 3000
+                            const playedDuration: number = DEFAULT_PART_DURATION_IN_MS / 2
                             const testee = createPart({
                                 isOnAir: true,
                                 executedAt: Date.now()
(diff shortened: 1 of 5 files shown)


src/app/core/models/part.ts
```diff
@@ -24,6 +26,7 @@ export class Part {
   public expectedDuration?: number
   public executedAt: number
   public playedDuration: number
+  public autoNext: AutoNext | undefined
```
LindvedKrvang, line 29: Any reason for not writing it as `public autoNext?: AutoNext`?

src/app/core/models/part.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+                            const result: number = testee.getDuration()
+
+                            expect(result).toBe(DEFAULT_PART_DURATION_IN_MS)
+                        })
+                    })
+
+                    describe('default part duration is shorter than the time since part was taken', () => {
+                        it('returns the time since part was taken', () => {
+                            const playedDuration: number = 5000
```
LindvedKrvang, line 104: Maybe do: `const playedDuration: number = DEFAULT_PART_DURATION_IN_MS + 5000`?
Then if DEFAULT_PART_DURATION_IN_MS has its value updated in the future to something higher than 5000, then this test wont break.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/app/core/models/part.ts
KvelaGorrrrnio: No. Is changed now.

src/app/core/models/part.spec.ts
KvelaGorrrrnio: Good observation. Is changed now.

PR #20, opened 2023-09-25
Sof 1529/last part in segment indicator

KvelaGorrrrnio:
The PR introduces the end of segment indicator as shown below.
Skærmbillede 2023-09-25 kl  13 56 58

Skærmbillede 2023-09-25 kl  13 58 16

Skærmbillede 2023-09-25 kl  13 57 15


src/app/rundown/components/segment-end-indicator/segment-end-indicator.component.html
```diff
@@ -0,0 +1,11 @@
+<sofie-timeline-flag
+    [label]="lastPartInSegment?.autoNext ? 'AUTO' : (lastPartInSegment?.isOnAir ? 'NEXT' : undefined)"
```
andr9528, line 2: Don't know if it can be done any better in Angulars Html templates, but nesting ternaries are something we don't like, as far as i know.

src/app/rundown/components/part/part.component.html
```diff
@@ -8,5 +8,5 @@
       <sofie-piece *ngFor="let piece of pieces" [expectedPartDuration]="part.expectedDuration ? part.expectedDuration : DEFAULT_PART_DURATION_IN_MS" [partDuration]="duration" [piece]="piece" [pixelsPerSecond]="pixelsPerSecond"></sofie-piece>
     </div>
   </div>
-  <sofie-timeline-next-indicator [part]="part"></sofie-timeline-next-indicator>
+  <sofie-timeline-flag [label]="part.isNext ? 'NEXT' : undefined" [turnedOn]="part.isNext"></sofie-timeline-flag>
```
andr9528, line 11: There is propably a good reason for it, by why the renaming of the component?

src/app/rundown/components/segment-end-indicator/segment-end-indicator.component.ts
```diff
@@ -0,0 +1,18 @@
...
+  host: {
+    class: 'c-segment-end-indicator',
+  },
+})
+export class SegmentEndIndicatorComponent {
+  @Input()
+  public lastPartInSegment?: Part
+
+  @HostBinding('class.c-segment-end-indicator--active')
+  public get isLastPart() { return this.lastPartInSegment?.isOnAir ?? false }
```
andr9528, line 17: From the name of the getter, with the knowledge that i have supplied a part to an input called `lastPartInSegment`, i'd guess that this would always return true.
Even if the last part is not on air, it is still the last part.

Also adding return type to the getter, if possible, would be nice.

src/app/rundown/components/part/part.component.html
RasmusAlbrektsen: I think we're open for suggestions here, however, the renaming was due to it serving multiple purposes as either border flags, the actual Next-cursor and the changed look once reaching the last part in a segment.

LindvedKrvang: Because the component now does something different, so the name should be updated to reflect what it does now.

src/app/rundown/components/segment-end-indicator/segment-end-indicator.component.ts
LindvedKrvang: It will only return true when the last Part is on Air.
A better name for the getter would be `isLastPartOnAir`

LindvedKrvang (approved):

src/app/rundown/components/segment-end-indicator/segment-end-indicator.component.ts
RasmusAlbrektsen: Adding return types gave some issues with some versions of these getters, but after trying it again just now it didn't complain. I don't remember if this was an issue here @KvelaGorrrrnio?

RasmusAlbrektsen (approved):

src/app/rundown/components/part/part.component.html
KvelaGorrrrnio: Previously, it was only used for the next-flag indicator on a part. Now is a bit more generic, such that it can also be used for the end of segment indicator.

src/app/rundown/components/segment-end-indicator/segment-end-indicator.component.ts
KvelaGorrrrnio: That was bad naming. I have changed it to `isLastPartOnAir` and added return type.
We had funny issues where it wouldn't compile if we provided a return type – but it works here.

PR #21, opened 2023-09-25
SOF-1529/ rundown view loading and background color

RasmusAlbrektsen:
This PR introduces a basic "loading"-functionality inside the Rundown View. An example can be seen below:

image

### DIFF
--- a/src/app/core/services/rundown-state.service.ts
+++ b/src/app/core/services/rundown-state.service.ts
@@ -17,14 +17,18 @@ import { ConnectionStatusObserver } from './connection-status-observer.service'
 
 @Injectable()
 export class RundownStateService implements OnDestroy {
+  private readonly isLoadingSubject: BehaviorSubject<boolean>
   private readonly rundownSubjects: Map<string, BehaviorSubject<Rundown>> = new Map()
   private eventSubscriptions: EventSubscription[]
+
+  private isLoading = true
 
   constructor(
       private readonly rundownService: RundownService,
       private readonly rundownEventObserver: RundownEventObserver,
       private readonly connectionStatusObserver: ConnectionStatusObserver
   ) {
+    this.isLoadingSubject = new BehaviorSubject<boolean>(this.isLoading)
     this.subscribeToEvents()
   }
 
@@ -54,6 +58,11 @@ export class RundownStateService implements OnDestroy {
     this.fetchRundown(rundownId)
         .then(rundown => rundownSubject.next(rundown))
         .catch(error => console.error('[error]', `Encountered error while fetching rundown with id '${rundownId}':`, error))
+  }
+
+  private setIsLoading(isLoading: boolean): void {
+    this.isLoading = isLoading
+    this.isLoadingSubject.next(this.isLoading)
   }
 
   private subscribeToRundownEvents(): EventSubscription[] {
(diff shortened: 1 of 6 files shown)


src/app/rundown/components/rundown/rundown.component.html
```diff
@@ -1,21 +1,28 @@
...
-        [segment]="segment"
-        [isOnAir]="segment.isOnAir"
-      ></sofie-segment>
+      <div>
+        <button *ngIf="!rundown?.isActive; else activeRundownButtons" (click)="activateRundown()">Activate Rundown</button>
+        <ng-template #activeRundownButtons>
+          <button (click)="takeNext()">Take Next</button>
+          <button (click)="resetRundown()">Reset Rundown</button>
+          <button (click)="deactivateRundown()">Deactivate Rundown</button>
+        </ng-template>
```
LindvedKrvang, line 16: If you want, this ng-template can be placed on the "top level":
```
// Your loading div

// Your #rundown template

// Your #activeRundownButtons template
```

I think it comes down to personal preference, but one of the benefits of templates is you can "extract" it from the surrounding markup just like you can extract a method.

src/app/rundown/components/rundown/rundown.component.ts
```diff
@@ -16,7 +16,10 @@ export class RundownComponent implements OnInit, OnDestroy {
 
   public rundown?: Rundown
   public adLibPieceIdentifiers: Identifier[] = []
+  public isLoading: boolean = true
   private rundownSubscription?: SubscriptionLike
+  private isLoadingSubscription?: SubscriptionLike
+  private subscriptions: SubscriptionLike[] = []
```
LindvedKrvang, line 22: This isn't used and should be deleted.

src/app/rundown/components/rundown/rundown.component.ts
```diff
@@ -16,7 +16,10 @@ export class RundownComponent implements OnInit, OnDestroy {
 
   public rundown?: Rundown
   public adLibPieceIdentifiers: Identifier[] = []
+  public isLoading: boolean = true
```
LindvedKrvang, line 19: Your current implementation does not consider if we are done loading and we didn't get a Rundown (which is fine).
But since you only have two states "I don't have a Rundown" and "I have a Rundown", then you don't need this `isLoading` variable because `rundown` is optional.

As long as `rundown` is undefined, you show your loading message. When it's no longer undefined, you show the Rundown.

LindvedKrvang: If you implemenet it like this, then you don't need to subscribe to the loading subject and then you don't need to implement the loading subject in the state service at all (atleast for now).

src/app/rundown/components/rundown/rundown.component.ts
LindvedKrvang: By "this" I mean the `subscriptions` variable.

src/app/rundown/components/rundown/rundown.component.ts
LindvedKrvang: But it does also mean we at this point in time wouldn't be able to show a message saying "This Rundown has been deleted" etc...

src/app/rundown/components/rundown/rundown.component.ts
RasmusAlbrektsen: Good catch, a left-over from a different implementation i tried. I'll remove it!

src/app/rundown/components/rundown/rundown.component.ts
RasmusAlbrektsen: I have changed this to the much simpler suggestion!

src/app/rundown/components/rundown/rundown.component.html
RasmusAlbrektsen: I extracted the template. I liked the idea of extracting it, and since it is only temporary anyway, i thought it best.

src/app/core/services/rundown-state.service.ts
```diff
@@ -56,6 +60,11 @@ export class RundownStateService implements OnDestroy {
         .catch(error => console.error('[error]', `Encountered error while fetching rundown with id '${rundownId}':`, error))
   }
 
+  private setIsLoading(isLoading: boolean): void {
```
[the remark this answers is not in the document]
RasmusAlbrektsen: No longer relevant after removing it.

andr9528 (approved):

src/app/rundown/components/rundown/rundown.component.html
```diff
@@ -1,21 +1,28 @@
 <div class="title">{{rundown?.name}}</div>
-<div class="container">
-  <div>
+
+<div *ngIf="isLoading; else loadedRundown" class='c-loading-rundown'>
+  <mat-card-title>Loading the rundown...</mat-card-title>
+</div>
+
+<ng-template #loadedRundown>
```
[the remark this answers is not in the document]
LindvedKrvang: Ahh, fair point :) 

Another way of looking at it is having your "main scenario" of the component to _not_ be in a template, but have other "alternative scenarios" in templates.
In other words: You are in a component called `RundownComponent`. The main thing this component wants to do is render a `Rundown` and displaying that we are loading a Rundown is something we _sometimes_ has to do, so it's not the main thing of this component.
So your markup would look like this instead:
```
<div *ngIf="rundown; else #loading">
  // Render Rundown markup
</div>

<ng-templante #loading>
  // Render loading markup
</ng-template>
```

LindvedKrvang (approved):

KvelaGorrrrnio (approved):

PR #22, opened 2023-09-26
feat: Adds ability to set first segment part as next by clicking header

RasmusAlbrektsen:
This PR includes the ability of clicking on the segment header and thereby setting the first part of the segment as next part. By request of @LindvedKrvang :-)

### DIFF
--- a/src/app/rundown/components/segment/segment.component.ts
+++ b/src/app/rundown/components/segment/segment.component.ts
@@ -11,7 +11,6 @@ import { Segment } from '../../../core/models/segment'
 import { Part } from '../../../core/models/part'
 import {PieceLayerService} from "../../../shared/services/piece-layer.service";
 import { PieceLayer } from '../../../shared/enums/piece-layer'
-import { RundownCursor } from '../../../core/models/rundown-cursor'
 import {RundownService} from "../../../core/abstractions/rundown.service";
 
 @Component({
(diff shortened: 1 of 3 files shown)


LindvedKrvang (approved):

src/app/rundown/components/segment/segment.component.ts
```diff
@@ -91,6 +94,11 @@ export class SegmentComponent implements OnChanges, OnDestroy {
     this.timeReference = timeSpendUntilActivePart + timeSpendInActivePart
   }
 
+  public setFirstPartAsNext(): void {
+    const firstPart: Part = this.segment.parts[0]
```
andr9528, line 98: Technically a Segment can have 0 parts, thus this _could_ return undefined. 
Don't know if this is something that is handled elsewhere, or even if it should be handled on the frontend.

andr9528 (approved):
Only had one comment, which isn't blocking a merge in my mind, just something that might be worth noting.

src/app/rundown/components/segment/segment.component.ts
RasmusAlbrektsen: For now i've added handling for the first part not being set as next if it contains no pieces. Whether the handling should extend further into scenarios where we e.g. have a first part as invalid and then should skip to the one after, is not considered. But i don't know if that is even relevant at this point in time.

LindvedKrvang (approved):

src/app/rundown/components/segment/segment.component.ts
KvelaGorrrrnio: Perhaps it should search for the first valid part, if any, instead.

KvelaGorrrrnio (approved):

src/app/rundown/components/segment/segment.component.ts
RasmusAlbrektsen: I have changed it to find a valid part instead!

PR #24, opened 2023-09-26
SOF-1529: Immutable data updates for reference values

KvelaGorrrrnio:
Converting the entity structure from class based to using entity services and entity objects, to get reactivity for reference values.


src/app/core/services/rundown-state.service.ts
```diff
@@ -125,15 +132,17 @@ export class RundownStateService implements OnDestroy {
       return
     }
 
-    part.insertAdLibPiece(event.adLibPiece)
+    // TODO: Out-commented due to current changes in AdLib API. Wait until that is somewhat stable.
+    //part.insertAdLibPiece(event.adLibPiece)
```
andr9528, line 136: I think you can safely remove these. Same goes for the similarly outcommented adlib related line further down.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.ts
```diff
@@ -36,11 +36,13 @@ export class RundownOverviewComponent implements OnInit, OnDestroy {
   }
 
   public navigateToRundown(basicRundown: BasicRundown): void {
-    this.router.navigate([Paths.RUNDOWNS, basicRundown.id])
+    const routeSegments: string[] = [Paths.RUNDOWNS, basicRundown.id]
+    this.router.navigate(routeSegments)
+        .catch(() => console.warn('[info] Failed navigating with route segments:', routeSegments))
```
andr9528, line 41: soooo which is it `warn` or `info`?

andr9528 (approved):

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.ts
KvelaGorrrrnio: Good point 😂 I'll change both to warn.

src/app/core/services/rundown-state.service.ts
KvelaGorrrrnio: It has been removed.

src/app/rundown/components/full-segment-timeline/full-segment-timeline.component.ts
```diff
@@ -27,4 +24,8 @@ export class FullSegmentTimelineComponent {
   public setNextEvent: EventEmitter<RundownCursor> = new EventEmitter()
 
   public pixelsPerSecond: number = 30
+
+  public trackPart(_: number, part: Part): string {
```
LindvedKrvang, line 28: Is the `_: number` so you don't have to write `() => trackPart(part)` where you call the method?

If yes, then you have written your method for _how you intent to use it_ instead of _what does it need to do_, which is a bad habit/practice ;)

src/app/rundown/components/rundown/rundown.component.ts
```diff
@@ -74,4 +75,8 @@ export class RundownComponent implements OnInit, OnDestroy {
     }
     this.rundownService.takeNext(this.rundown.id).subscribe()
   }
+
+  public trackSegment(_: number, segment: Segment): string {
```
LindvedKrvang, line 79: Same comment as for `trackPart`

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/app/rundown/components/full-segment-timeline/full-segment-timeline.component.ts
KvelaGorrrrnio: I have tried doing that, but the Angular syntax for pipes do not allow for lambda functions in expressions 😞

PR #26, opened 2023-09-27
SOF-1622 Create editorial line component. Display budget duration on …

temk-tv2:
- Editorial line component
- Display budget duration on segment


src/app/rundown/components/editorial-line/editorial-line.component.scss
```diff
@@ -0,0 +1,7 @@
+:host {
+  top: var(--toolbar-height);
+  bottom: 0;
+  border-left: 2px dashed grey;
+  margin-left: -1px;
+  position: absolute;
```
KvelaGorrrrnio, line 6: Minor thing. It would be nice to the structural attributes as the first entries – so in this case position at the top. The (incomplete) order I personally use  is:
- Structural (position, display, ...)
- Placement and sizing
- Colors
- Text formatting

This way, the higher up the list you are, the higher impact it has on the layout.

src/app/rundown/components/editorial-line/editorial-line.component.html
KvelaGorrrrnio, line 1: Since the component has no internal HTML, perhaps this file can be deleted and the template can be set in the component declaration file (.ts).

src/app/rundown/components/editorial-line/editorial-line.component.ts
```diff
@@ -0,0 +1,19 @@
...
+})
+export class EditorialLineComponent {
+    @Input()
+    public budgetDuration: number
+
+    @Input()
+    public pixelsPerSecond: number
+
+    @HostBinding('style.left.px')
+    get left() {
```
KvelaGorrrrnio, line 16: Explicit access modifier and return type would be nice here, such that it becomes: `public get left(): number`

KvelaGorrrrnio (approved):
Looks good :+1: A few minor comments to be addressed.

src/app/rundown/components/editorial-line/editorial-line.component.ts
LindvedKrvang: Aren't we enforcing explicit access level modifiers?

src/app/rundown/components/segment/segment.component.html
```diff
@@ -2,7 +2,7 @@
   <div class="c-segment__header" (click)="setFirstValidPartAsNext()">
     <div class="c-segment__header-toolbar">
       <span>123</span>
-      <span style="margin-left: 50px;">00:30</span>
+      <span style="margin-left: 50px;" *ngIf="segment.budgetDuration !== undefined">{{segment.budgetDuration / 1000 | timestamp}}</span>
```
LindvedKrvang, line 5: You can write `!!segment.budgetDuration` instead of `segment.budgetDuration !== undefined`.
I think that `*ngIf=segment.budgetDuration` would also be enough :)

LindvedKrvang, line 5: Why 1000? Would be nice to extract this into a variable with a descriptive name :)

LindvedKrvang (approved):

src/app/rundown/components/editorial-line/editorial-line.component.ts
andr9528: Sounds like a missing linting rule.

KvelaGorrrrnio: We haven't setup any linting yet for the frontend. I'll setup it up on monday.

PR #27, opened 2023-09-28
doc: Added Developer convention document.

KvelaGorrrrnio:
Draft for our developer conventions in the frontend. Logging strategy applies to all of our projects, so that is also defined on [Confluence](https://tv2cms.atlassian.net/wiki/spaces/SOF/pages/4348477470/Logging+Strategy).

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -31,7 +31,7 @@ _**TODO:** Find out how to handle global settings such as color palettes, sizing
 
 **Attribute order**
 
-When writing scss the suggested attribute order is:
+In order to have a certain degree of consistency, the following style attribute order is suggested:
 
 * **Structural:** Position type, display type, box-sizing, etc.
 * **Placement and sizing:** Absolute position, width, height, padding, margin, etc.
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -4,10 +4,18 @@
 
 ### Naming conventions
 
-_**TODO:** Should we use units formats for e.g. number values (`timeSinceYesterdayInMs` | `activatedRundownAtEpochTime`)._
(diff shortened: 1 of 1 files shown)


RasmusAlbrektsen (approved):

andr9528 (commented):
There are some todo's that i guess we need to decide on before this gets merged?
When/If we create a similar document for Sofie-Server, how much duplication, if any, will there be? 

The Logging section for sure, but if it is only that single one, i think it is fine. If however there ends up being multiple sections duplicated, then moving/making this document somewhere shared by both environment, might be wiser.

Using GitHub wiki as a Git Submodule is one way i personally have keept a large number of .md files grouped, and accesible from the main repository. Doing something similar should allow the document to be shared between diffrent repositories, if we end up needing to do that.

DEVELOPER.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Naming conventions
+
+**Variables**
+
+_**TODO:** Find out how to handle global settings such as color palettes, sizings, etc..._
+
+**Attribute order**
+
+When writing scss the suggested attribute order is:
```
LindvedKrvang, line 34: Should we emphasize that the goal here is to have consistency?

LindvedKrvang (approved):

DEVELOPER.md
KvelaGorrrrnio: Yes. I have updated it to reflect that.

KvelaGorrrrnio:
@andr9528 It is a valid concern. Logging strategy is something we want across all of our projects, so it makes sense that we have one source of truth for it. Currently the source of truth is on Confluence.

Regarding the todos: I have updated the one with units for number variables. But this should be a living document that we can alter and add to when the need arise.

PR #30, opened 2023-10-02
Setup and apply linting

KvelaGorrrrnio:
- The project is setup with eslint and prettier for typescript linting, using the same rules as sofie-server.
- Stricter type check for production build is added (same as sofie-server).
- `yarn lint` now does a lint check.
- `yarn lint-fix` has been added.
- Yarn is updated to version 3.6.3.
- `yarn validate` now uses `yarn npm audit` instead of `yarn audit`.
- Linting rules and type checking for production build has been applied/fixed.
- Updated a few components to not use BEMIT.


andr9528 (approved):
Half-blind approve. Way to many changes to get a good overview of what changes has been made, that isn't just Linting...

Would be cool if the difference didn't show indenting and single space changes, as a difference. That would have cut away most of the 'differences', making it easier to spot the changes that actually matters.
Searching google, this seems to be a possibility, wonder if it is something that can be turned on for GitHub repositories. (`git diff --ignore-space-change`)

andr9528:
To add to my prior comment. By adding `?w=1` to end of the URL when viewing chaged files in a PR, it will ignore whitespace changes. This is NOT persistet between different reviews, unless you add a Chrome/Firefox extension. You can read more about it here: https://stackoverflow.com/questions/37007300/how-to-ignore-whitespace-in-github-when-comparing

KvelaGorrrrnio:
@andr9528 You can also use hide whitspace from the GUI :)
Skærmbillede 2023-10-05 kl  07 33 03

The most relevant parts to review are the files outside of the src folder. The only changes that should be able to change something are the change of not using BEMIT for `SegmentEndIndicatorComponent` and `TimelineFlagComponent`.

RasmusAlbrektsen (approved):
Looks good!

PR #32, opened 2023-10-04
API base url configuration is based on execution environment

KvelaGorrrrnio:
Development points to `http://localhost:3005/api` while production points to the website origin suffixed with `/api`.

### DIFF
--- a/src/environments/environment-interface.ts
+++ b/src/environments/environment-interface.ts
@@ -1,0 +1,4 @@
+export interface Environment {
+  production: boolean
+  apiBaseUrl: string
+}
--- a/src/environments/environment.dev.ts
+++ b/src/environments/environment.dev.ts
@@ -2,5 +2,5 @@ import { Environment } from './environment-interface'
 
 export const environment: Environment = {
   production: true,
-  apiBaseUrl: `${window.location.origin}/api`,
+  apiBaseUrl: 'http://localhost:3005/api',
 }
--- a/src/environments/environment.dev.ts
+++ b/src/environments/environment.dev.ts
@@ -1,6 +1,6 @@
 import { Environment } from './environment-interface'
 
 export const environment: Environment = {
-  production: true,
+  production: false,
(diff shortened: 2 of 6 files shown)


andr9528 (approved):

src/environments/environment-interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface Environment {
+  production: boolean
```
LindvedKrvang, line 2: Boolean is quite limitting. 
Already now we have three different environments: `Production`, `Staging` and `Development`

src/environments/environment.dev.ts
```diff
@@ -0,0 +1,6 @@
+import { Environment } from './environment-interface'
+
+export const environment: Environment = {
+  production: true,
```
LindvedKrvang, line 4: This is the `environment.dev.ts`, shouldn't this be false?

src/environments/environment-interface.ts
KvelaGorrrrnio: `production` was already defined. And do we intend to make builds specific to a staging environment where the configuration is different from production?

src/environments/environment.dev.ts
KvelaGorrrrnio: It should and it is now.

RasmusAlbrektsen (approved):

src/environments/environment-interface.ts
LindvedKrvang: I don't think so, but using a boolean for this is still vague.
What environment are we in when `production = false`? It could be anything.

LindvedKrvang: But I think that an argument could be made that we shouldn't include our environment files in VCS at all. We can't store keys (and production URLS) in the environment when we publish them to Git.

KvelaGorrrrnio: I just checked it out and `production` is a variable that Angular itself uses.

PR #36, opened 2023-10-06
Sof 1645/add logger library

KvelaGorrrrnio:
- Adds @tv2media/logger library and removes console statements with log statements that uses a logger.
- Adds support to change execution env with `http://<sofie-url>/?ENV=production|prod|development|dev|staging|stage|local`.
- Adds support to change log level with `http://<sofie-url>/?LOG_LEVEL=error|warn|info|debug|trace`.


src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -7,6 +7,7 @@ import { BasicRundown } from '../models/basic-rundown'
 import { BasicRundownService } from '../abstractions/basic-rundown.service'
 import { RundownEvent } from '../models/rundown-event'
 import { BasicRundownEntityService } from './models/basic-rundown-entity.service'
+import { Logger } from '../../../../../../mediatech-logger'
```
LindvedKrvang, line 10: Is this right?
`Logger` is declared in `core/abstractions/logger.service`.

You have quite a few of these imports. (And some imports that uses `core/abstractions/logger.service`, so something isn't right)

src/app/core/services/tv2-logger.service.ts
```diff
@@ -0,0 +1,39 @@
+import { DefaultLogger } from '@tv2media/logger/web'
+import { environment } from '../../../environments/environment'
+
+export class Tv2LoggerService extends DefaultLogger {
+  constructor() {
+    ensureWindowEnv({
+      ENV: getExecutionEnvFromUrl() ?? (environment.production ? 'production' : 'development'),
```
LindvedKrvang, line 7: Are we going to let staging count as 'development'?

src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -17,12 +18,16 @@ export class BasicRundownStateService implements OnDestroy {
...
+
   constructor(
     private readonly basicRundownService: BasicRundownService,
     private readonly rundownEventObserver: RundownEventObserver,
     private readonly connectionStatusObserver: ConnectionStatusObserver,
-    private readonly basicRundownEntityService: BasicRundownEntityService
+    private readonly basicRundownEntityService: BasicRundownEntityService,
+    logger: Logger
   ) {
+    this.logger = logger.tag('BasicRundownStateService')
```
LindvedKrvang, line 30: Are we interested in the syntax: `logger.tag(BasicRundownStateService.name)`?

src/app/shared/components/header/header.component.ts
```diff
@@ -9,16 +10,22 @@ import { Router } from '@angular/router'
...
-  constructor(private readonly router: Router) {}
+  constructor(
+    private readonly router: Router,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('HeaderComponent')
+  }
 
   public navigateHome(): void {
     const routeSegments: string[] = [Paths.HOME]
```
LindvedKrvang, line 23: Minor thing not really related to this PR:
`routeSegments` makes it sound like that it's actually Segments that are used for routes (which it is not), whereas `segmentRoutes` makes it sound like that it is a Route used for Segments (which it is)

src/main.ts
```diff
@@ -10,4 +10,6 @@ if (environment.production) {
 
 platformBrowserDynamic()
   .bootstrapModule(AppModule)
-  .catch(err => console.error(err))
+  // console.error is used here since dependency injection is set up at this moment.
```
LindvedKrvang, line 13: Shouldn't this be: `is not setup`?

LindvedKrvang: Minor thing: Consider placing the comment on the same line as what you are commenting (as long as it is short enough).
So it would be something like:
`.catch(error => ...) // Comment here`

This makes it less likely that the comment will be separated from the code that it is commenting on later in the lifespan of the project.

src/app/core/services/tv2-logger.service.ts
KvelaGorrrrnio: If nothing else is specified, I think it is a valid strategy. The question should probably be if we want to be able to persist modifications to the execution environment or log level in local storage.

src/app/core/services/basic-rundown-state.service.ts
KvelaGorrrrnio: If we want to be able to have these readable in production, then no.

src/app/core/services/basic-rundown-state.service.ts
KvelaGorrrrnio: That is incorrect and is fixed.

src/main.ts
KvelaGorrrrnio: The reason it was on two lines, was that I had trouble getting the comment and the rule on the same line. I found a way to do that, so it is on the same line now.

src/app/shared/components/header/header.component.ts
KvelaGorrrrnio: Segment is used here as in URL path segments. Should I change this to `pathSegments` or `urlSegments`?

src/app/core/services/tv2-logger.service.ts
LindvedKrvang: Modifications like, what loglevel is being used?

LindvedKrvang (approved):

src/app/shared/components/header/header.component.ts
LindvedKrvang: Neither ;)

`pathSegments` and `urlSegments` still indicates that it is a subtype of `Segment`, which they are _not_.
`segmentPaths` and `segmentUrls` indicates that they are paths or URLs, which they are.

I have no preference between `segmentRoutes`, `segmentPaths` and `segmentUrls` as long as `segment` isn't the last word.

src/app/rundown/components/rundown/rundown.component.ts
```diff
@@ -23,18 +24,22 @@ export class RundownComponent implements OnInit, OnDestroy {
...
   constructor(
     private readonly route: ActivatedRoute,
     private readonly rundownService: RundownService,
     private readonly rundownStateService: RundownStateService,
-    private readonly dialogService: DialogService
-  ) {}
+    private readonly dialogService: DialogService,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('RundownComponent')
```
andr9528, line 36: I know we can't use the `[class].name` in production, from a similar comment made by Lindved. But is there any other way we can ensure the string here always matches the class it is in. Maybe a Linting rule? It is just very easy thing to miss chainging, should a class/component be renamed.

andr9528 (approved):

RasmusAlbrektsen (approved):

src/app/core/services/tv2-logger.service.ts
KvelaGorrrrnio: Yes.

src/app/shared/components/header/header.component.ts
LindvedKrvang: We might be misunderstanding each other. The `segment` part of the name does not refer to our entity `Segment`? Since we have an entity/concept of a `Segment` in this codebase I thought that was what you where referring to...
If you are not referring to the entity I think you should remove `segment` from the name alltogether to avoid further confusion.

If it does refer to our entity, then agree to disagree ;) 

Sidenote: If there is only one path then offcourse you would call it `SegmentPath` instead of `SegmentPaths`. `routeSegments` (which is the original variable name) also indicates that there are multiple routes/paths ;)

src/app/rundown/components/rundown/rundown.component.ts
LindvedKrvang: This really is a small thing that is not worth spending that much time/resources on.

src/app/shared/components/header/header.component.ts
KvelaGorrrrnio: It is not the entity. I have changed it to `segmentedPath`, though I still think that `pathSegments` better describes the content.

PR #37, opened 2023-10-10
Fix/Active Status Doesn't Update in Rundown overview

RasmusAlbrektsen:
This PR fixes how we did not use the right method to update the active status on basic rundowns.

### DIFF
--- a/src/app/core/services/basic-rundown-state.service.ts
+++ b/src/app/core/services/basic-rundown-state.service.ts
@@ -68,7 +68,7 @@ export class BasicRundownStateService implements OnDestroy {
       if (basicRundown.id !== event.rundownId) {
         return basicRundown
       }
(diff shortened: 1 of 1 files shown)


src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -68,7 +68,7 @@ export class BasicRundownStateService implements OnDestroy {
       if (basicRundown.id !== event.rundownId) {
         return basicRundown
       }
-      return this.basicRundownEntityService.deactivate(basicRundown)
```
LindvedKrvang, line 71: Sounds like a good usecase to write a test for ;)

LindvedKrvang (approved):

andr9528 (approved):

PR #38, opened 2023-10-10
Sof 1627/create shelf component

RasmusAlbrektsen:
This PR creates the Shelf component, and has made it possible to drag using a handle. 
This includes a WIP version of the Keyboard which is subject to change in SOF-1631, and is mostly for the purpose of demonstration.


src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
```diff
@@ -0,0 +1,17 @@
...
+        class="keybinding"
+        style="padding: 12px 20px; margin: 10px;"
+        [class.keybinding--matched]="isKeyBindingMatched(keyBinding)"
+    >
+        {{ keyBinding.label }} ({{ displayKeyBinding(keyBinding) }})
+    </div>
+    <pre style="padding: 20px; color: white;">
+        Pressed:keys
+        {{ pressedKeys }}
+    </pre>
```
LindvedKrvang, line 16: Is this for development purposes?
If yes, is it still needed?

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
```diff
@@ -0,0 +1,17 @@
+<div [style.flex-basis.px]="dragOffsetInPixels" [style.height.px]="dragOffsetInPixels">
+    <div #dragHandle class="drag-handle">
+        <sofie-icon-button [iconButton]="IconButton.BARS" [iconButtonSize]="IconButtonSize.XL"></sofie-icon-button>
+    </div>
+    <div
+        *ngFor="let keyBinding of keyBindings" (click)="keyBinding.action()"
+        class="keybinding"
+        style="padding: 12px 20px; margin: 10px;"
```
LindvedKrvang, line 8: This is inline styling. We don't want inline styling.

LindvedKrvang: What you could do is create some "helper" CSS classes that provides various forms of paddings and margins.
Something like `sp`, `mp`, `lp`, `spr`, `mpt`, `lm`.
The acronyms stands for `small padding`, `medium padding`, `large padding`, `small padding right`, `medium padding top` and `large margin`.

The acronums aren't pretty, but they are a handy way to quickly add paddings and margins to your markup.
You could discuss with Anders if this is something we want.

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  protected readonly IconButtonSize = IconButtonSize
+
+  @ViewChild('dragHandle')
+  private readonly dragHandleElement: ElementRef<HTMLDivElement>
+
+  constructor(
+    private readonly keyboardBindingService: KeyboardBindingService,
+    private readonly keyboardBindingMatcher: KeyboardBindingMatcher
+  ) {
+    // TODO: handle closing subscriptions on teardown
```
LindvedKrvang, line 28: Isn't this already handled by the `unsubsubscribe` in the `ngOnDestroy`?

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
```diff
@@ -0,0 +1,67 @@
...
+import { KeyBinding } from '../../models/key-binding'
+import { KeyboardBindingMatcher } from '../../services/keyboard-binding.matcher'
+import { IconButton, IconButtonSize } from '../../../shared/enums/icon-button'
+
+@Component({
+  selector: 'sofie-producer-shelf',
+  templateUrl: './producer-shelf.component.html',
+  styleUrls: ['./producer-shelf.component.scss'],
+})
+export class ProducerShelfComponent implements OnDestroy {
```
LindvedKrvang, line 12: What's the reason for calling it "producer" shelf?

LindvedKrvang, line 12: It looks like this component is responsible for both handling the dragging of the shelf and the content within the shelf.
It would be nice if these two things would be separated.

This can be achieved with `content injection` https://angular.io/guide/content-projection

We could then have one component incharge of the drag functionality, `DragableShelfComponent`, that we can inject content into.
In this case we would inject a `RundownViewShelfComponent` (or whatever name make sense) that knows how to display the keyboard etc.

src/app/producer-shelf/factories/camera-key-binding.factory.ts
```diff
@@ -0,0 +1,21 @@
+import { KeyBinding } from '../models/key-binding'
+import { Injectable } from '@angular/core'
+
+@Injectable()
+export class CameraKeyBindingFactory {
```
LindvedKrvang, line 5: Does this need to this specific e.i. does it need to be a `CameraKeyBindingFactory` and not an `ActionKeyBindingFactory`?

src/app/producer-shelf/models/key-binding.ts
```diff
@@ -0,0 +1,7 @@
+export interface KeyBinding {
+  readonly key: string
+  readonly modifiers: string[]
+  readonly label: string
+  readonly action: () => void
+  readonly onKeyPress: boolean
```
LindvedKrvang, line 6: `onKeyPress` is a common term for the function that should be executed on the press of the key.
Here it is a boolean which is a bit confusing. What does it mean when it is true, and what does it mean when it is false? That is not clear to me.

Personally, I would prefer `onKeyPress` to be the name for the attribute containing the function to be executed instead of `action`.
`action` is a bit vague and since we have a concept of `Action` in the codebase it is easy to confuse this with being an attribute that holds an actual Action, which it does not.

src/app/producer-shelf/producer-shelf.module.ts
```diff
@@ -0,0 +1,16 @@
...
+import { CameraKeyBindingFactory } from './factories/camera-key-binding.factory'
+import { SharedModule } from '../shared/shared.module'
+
+@NgModule({
+  declarations: [ProducerShelfComponent],
+  imports: [CommonModule, SharedModule],
+  providers: [{ provide: KeyboardBindingService, useClass: HardcodedProducerKeyboardBindingService }, KeyboardBindingMatcher, CameraKeyBindingFactory],
+  exports: [ProducerShelfComponent],
+})
+export class ProducerShelfModule {}
```
LindvedKrvang, line 16: What's the difference between a `ProducerShelf` and a `Shelf`?

src/app/producer-shelf/services/hardcoded-producer-keyboard-binding.service.ts
```diff
@@ -0,0 +1,145 @@
+import { KeyboardBindingService } from '../abstractions/keyboard-binding.service'
+import { BehaviorSubject, fromEvent, Subject, Subscription } from 'rxjs'
+import { KeyBinding } from '../models/key-binding'
+import { KeyboardBindingMatcher } from './keyboard-binding.matcher'
+import { HostListener, Injectable } from '@angular/core'
+import { CameraKeyBindingFactory } from '../factories/camera-key-binding.factory'
+import { RundownService } from '../../core/abstractions/rundown.service'
+
+@Injectable()
+export class HardcodedProducerKeyboardBindingService implements KeyboardBindingService {
```
LindvedKrvang, line 10: I assume this class is for the demo purposes of the keyboard?

If yes, I'm not going to spend much time reveiwing it

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
RasmusAlbrektsen: I decided to delete the content in the `ProducerShelf`, since it is all being implemented in an entirely new fashion for @KvelaGorrrrnio's PR.

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
RasmusAlbrektsen: It was indeed. I have emptied the component for now, however.

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
RasmusAlbrektsen: I extracted the drag/resize functionality into a wrapper component named `DraggableShelf`.

RasmusAlbrektsen: This was used to create the `DraggableShelf` component. The `ProducerShelf` will now be wrapped by the drag functionality.

src/app/producer-shelf/factories/camera-key-binding.factory.ts
RasmusAlbrektsen: Are you doing this in your PR @KvelaGorrrrnio?

src/app/producer-shelf/models/key-binding.ts
RasmusAlbrektsen: Yes, i can delete it here and scrub out most of the things not currently used. But it will happen in the PR right after. 
Should i delete all these thing and let @KvelaGorrrrnio introduce it all instead?

KvelaGorrrrnio: yes

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.scss
```diff
@@ -0,0 +1,6 @@
+:host {
+  position: relative;
+  display: block;
+  border-top: 2px solid #555;
+  pointer-events: none;
```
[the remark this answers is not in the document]
RasmusAlbrektsen: Because it is impolite to point Anders...
I have removed it, there was no reason :-)

src/app/producer-shelf/models/key-binding.ts
RasmusAlbrektsen: Thy will be done. It has been removed!

src/app/draggable-shelf/draggable-shelf.component.html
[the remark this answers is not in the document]
RasmusAlbrektsen: I have moved it to the shared module!

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.scss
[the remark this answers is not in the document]
KvelaGorrrrnio: Sorry for my bad manners :wink:

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
LindvedKrvang: I sense a merge conflict in the making :D

src/app/shared/components/draggable-shelf/draggable-shelf.component.html
```diff
@@ -0,0 +1,6 @@
+<div [style.flex-basis.px]="dragOffsetInPixels" [style.height.px]="dragOffsetInPixels">
```
KvelaGorrrrnio, line 1: You can use host bindings to remove this div.

KvelaGorrrrnio (approved):
🚀

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
LindvedKrvang: > This is for development purposes,

We should try and not commit stuff for development purposes (we should definately not merge them). It's extremely easy to forget to remove them. Escpecially when it has been pushed and merged, because now there is no longer a git diff they will appear in, so you have to remember it yourself.

In my current draft PR I have committed some stuff for development purposes, but I intent to remove them before marking the PR ready for review.

src/app/producer-shelf/producer-shelf.module.ts
LindvedKrvang: Answered in another comment

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
```diff
@@ -1,17 +0,0 @@
-<div [style.flex-basis.px]="dragOffsetInPixels" [style.height.px]="dragOffsetInPixels">
```
LindvedKrvang: This file is now empty. Is that on purpose?

LindvedKrvang (approved):

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
RasmusAlbrektsen: It is, i just wanted it to exist so i could show the selector structure still. But as we discussed, the content inside it was useless, so for now i scrubbed it.

src/app/shared/components/draggable-shelf/draggable-shelf.component.html
RasmusAlbrektsen: I have done so now! :-)

PR #39, opened 2023-10-10
Sof 1627/refactor rundown, header and shelf into rundown view component

RasmusAlbrektsen:
This PR creates a Rundown View component and restructures the HTML so that the the following components are moved inside of it, rather than being inside the Rundown Component:
* The Rundown Component
* The Rundown Header Component
* The Producer Shelf


src/app/rundown-view/components/rundown-view/rundown-view.component.html
```diff
@@ -0,0 +1,11 @@
+<sofie-rundown-header *ngIf="rundown" class="c-rundown-header" [rundown]="rundown"></sofie-rundown-header>
+
+<div *ngIf="!rundown; else loadedRundown" class='c-loading-rundown'>
+    <mat-card-title>Loading the rundown...</mat-card-title>
+</div>
+
+<ng-template #loadedRundown>
+    <sofie-rundown [rundown]="rundown"></sofie-rundown>
+</ng-template>
```
LindvedKrvang, line 9: If you flip the "default state" of this component to be that the Rundown is loaded, you should be able to write something like:
```
<ng-container *ngIf=rundown; else loadingRundown">
  <sofie-rundown-header ...></sofie-rundown-header>
  <sofie-rundown ...></sofie-rundown>
</ng-container>

<ng-template #loadingRundown>
  <mat-card-title>Loading the rundonw...</mat-card-title>
</ng-template>
```

Sidenote: `ng-container` does not get added to the DOM, so it's perfect for elements who has no visiual responsibilities. So you don't need to use a `div` that adds padding and stuff.

LindvedKrvang (changes requested):
The request changes is only for the comment in `rundown.component.ts`

andr9528 (approved):

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -21,30 +20,29 @@ export class RundownComponent implements OnInit, OnDestroy {
     }
   }
 
+  @Input()
   public rundown?: Rundown
```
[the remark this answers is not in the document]
RasmusAlbrektsen: Good catch! That's my bad, i somehow didn't remove this when i scrubbed the files. I wanted it to be the Rundown View that has the state service, and then then rundown is propagated as an input from there. I have changed it to do so now!

src/app/rundown-view/components/rundown-view/rundown-view.component.html
RasmusAlbrektsen: I really liked that suggestion, it has been implemented.

src/app/rundown-view/components/rundown-view/rundown-view.component.ts
```diff
@@ -0,0 +1,42 @@
...
+      this.logger.error("[error]: No rundownId found. Can't fetch Rundown")
+      return
+    }
+    this.rundownStateService
+      .subscribeToRundown(rundownId, rundown => {
+        this.rundown = rundown
+      })
+      .then(unsubscribeFromRundown => {
+        this.rundownSubscription = unsubscribeFromRundown
+      })
```
KvelaGorrrrnio, line 35: Since the parameter of the lambda is a Subscription(Like), you could call the parameter for `rundownSubscription` like the class attribute.

src/app/rundown-view/components/rundown-view/rundown-view.component.ts
```diff
@@ -0,0 +1,42 @@
...
+      .catch(error => this.logger.error(`[error] Failed subscribing to rundown with id '${rundownId}'.`, error))
```
KvelaGorrrrnio, line 36: The logger is used incorrectly here. You can use the `.data(data: unknown)` method prior to calling `.error`, such that this becomes:
```suggestion
      .catch(error => this.logger.data(error).error(`[error] Failed subscribing to rundown with id '${rundownId}'.`))
```

The second parameter of `.error` should be a metadata object that will be spread out into the root of the log entry. The same structure that is the outcome of the suggestion, can be achieved by wrapping the error in an object like so:
```suggestion
      .catch(error => this.logger.error(`[error] Failed subscribing to rundown with id '${rundownId}'.`, { error }))
```

LindvedKrvang (approved):
You have three folders pointing to an empty html file
image

src/app/rundown-view/components/rundown-view/rundown-view.component.ts
RasmusAlbrektsen: Changed to fit the name!

src/app/rundown-view/components/rundown-view/rundown-view.component.ts
RasmusAlbrektsen: I have used the suggestion!

PR #41, opened 2023-10-13
Sof 1545/able to set next from gui

RasmusAlbrektsen:
This PR introduces a generic context menu that is wrapped in the Rundown View by the Rundown Header, Segments and Parts. 
Furthermore, the arrow keys can now be used to navigate the next flag using `RundownNavigationService`.


LindvedKrvang (approved):
I'm not a fan of `rundown.isActive` being parsed so much around to the different components.
In most cases you just use them to check whether a component should be rendered _after_ you have called the component.
If you make that check _before_ you call your component you can get rid of a lot of these inputs.

src/app/rundown-view/components/part-context-menu/part-context-menu.component.ts
```diff
@@ -0,0 +1,32 @@
...
+  public readonly offAirContextMenuOptions: ContextMenuOption[] = [
+    {
+      label: 'Set part as next',
+      contextAction: (): void => this.setPartAsNext(),
+    },
+  ]
+
+  constructor(private readonly rundownService: RundownService) {}
+
+  public setPartAsNext(): void {
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Changed to private.

src/app/rundown-view/factories/key-binding.factory.ts
```diff
@@ -6,14 +6,23 @@ import { PartActionType } from '../../shared/models/action-type'
 import { RundownService } from '../../core/abstractions/rundown.service'
 import { Rundown } from '../../core/models/rundown'
 import { DialogService } from '../../shared/services/dialog.service'
+import { RundownNavigationService } from '../../shared/services/rundown-navigation-service'
+import { RundownCursor } from '../../core/models/rundown-cursor'
+import { Logger } from '../../core/abstractions/logger.service'
 
 @Injectable()
 export class KeyBindingFactory {
```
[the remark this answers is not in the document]
KvelaGorrrrnio: You can review this. The changes from SOF-1631 is merged into this, so these are the key bindings that should be used until we make them configurable.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.html
```diff
@@ -1,15 +1,9 @@
...
+<ng-container *ngIf="!rundown?.isActive; then deactivatedContextMenu; else activatedContextMenu"></ng-container>
+<ng-template #deactivatedContextMenu>
+    <sofie-context-menu [contextMenuOptions]="deactivatedContextMenuOptions" [contextMenuHeader]="shortenedRundownName"></sofie-context-menu>
+</ng-template>
 
-    <button class="c-context-menu-item" cdkMenuItem (click)="resetRundown()">Reset Rundown</button>
-    <button class="c-context-menu-item"></button>
+<ng-template #activatedContextMenu>
+    <sofie-context-menu [contextMenuOptions]="activatedContextMenuOptions" [contextMenuHeader]="shortenedRundownName"></sofie-context-menu>
+</ng-template>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: I have made a new computed property called `contextMenuOptions` that makes this check, such that the template is clean.
Adding onChange seems like overkill.

src/app/rundown-view/components/part-context-menu/part-context-menu.component.html
```diff
@@ -0,0 +1,3 @@
+<sofie-context-menu *ngIf="isRundownActive" [contextMenuOptions]="part.isOnAir ? [] : offAirContextMenuOptions"></sofie-context-menu>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: The part context menu is now only shown conditionally from the parent component.

src/app/rundown-view/components/segment-context-menu/segment-context-menu.component.html
```diff
@@ -0,0 +1,2 @@
+<sofie-context-menu *ngIf="isRundownActive" [contextMenuOptions]="contextMenuOptions"></sofie-context-menu>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: isRundownActive has been removed and the conditional check is made in the parent component.

src/app/rundown-view/components/segment-context-menu/segment-context-menu.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  @Input()
+  public isRundownActive: boolean
+
+  public readonly contextMenuOptions: ContextMenuOption[] = [
+    {
+      label: 'Set segment as Next',
+      contextAction: (): void => this.setFirstValidPartAsNext(),
+    },
+  ]
```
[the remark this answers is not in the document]
KvelaGorrrrnio: I am playing with the thought of moving it out into a singleton context menu component, and add a ContextMenuService that is responsible for actually setting the options and displaying it.

src/app/shared/components/context-menu/context-menu.component.html
```diff
@@ -0,0 +1,12 @@
+<div class="context-menu" cdkMenu>
+    <span *ngIf="contextMenuHeader" class="context-menu-header" cdkMenuGroup> {{contextMenuHeader}} </span>
+    <ng-container *ngFor="let option of contextMenuOptions">
+        <button
```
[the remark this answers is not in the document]
KvelaGorrrrnio: The loop has been moving to the button element and the container is removed.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -2,54 +2,82 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core'
...
+      label: 'Activate (On Air)',
+      contextAction: (): void => this.openActivationDialog(),
+    },
+    {
+      label: 'Reset Rundown',
+      contextAction: (): void => this.resetRundown(),
+    },
+  ]
+
+  public readonly contextMenuOptionsForActiveRundown: ContextMenuOption[] = [
```
LindvedKrvang, line 30: Do these need to be public?

LindvedKrvang (approved):

src/app/rundown-view/components/part-context-menu/part-context-menu.component.html
```diff
@@ -0,0 +1,3 @@
+<sofie-context-menu [contextMenuOptions]="part.isOnAir ? [] : offAirContextMenuOptions"></sofie-context-menu>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: This wouldn't show anything. I have removed this condition, since the parent component only shows it when on air.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
KvelaGorrrrnio: No. They are now private.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -2,54 +2,82 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core'
...
+
+  private readonly contextMenuOptionsForInactiveRundown: ContextMenuOption[] = [
+    {
+      label: 'Activate (On Air)',
+      contextAction: (): void => this.openActivationDialog(),
+    },
+    {
+      label: 'Reset Rundown',
+      contextAction: (): void => this.resetRundown(),
+    },
```
andr9528, line 27: This one is duplicated for the `contextMenuOptionsForActiveRundown`.   
Perhabs make a `commonContextMenuOptions`, which contains any overlap, and with spreed operator is put into the others. 

Although making that `commonContextMenuOptions`, might make it harder to order them... Adding a priority atribute could fix that, ordering the options from e.g lowest to highest.

Just a suggestion. Not that critical when there is only one that overlap, but as soon as there is 2-3 or more, then this might make sense. This might already be the case as soon as the 'Reload INews data', or what ever it is going to be named, is added.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -2,54 +2,82 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core'
...
+    if (this.rundown.isActive) {
       return
     }
     this.dialogService.createConfirmDialog(this.rundown.name, `Are you sure you want to activate the Rundown?`, 'Activate', () => this.activate())
   }
 
-  public activate(): void {
-    if (!this.rundown?.id) {
+  public openDeactivationDialog(): void {
+    if (!this.rundown.isActive) {
```
andr9528, line 62: At what point is this check performed, while the rundown is inactive?

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -2,54 +2,82 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core'
...
+  }
 
   constructor(
     private readonly rundownService: RundownService,
     private readonly dialogService: DialogService
   ) {}
 
   public openActivationDialog(): void {
-    if (!this.rundown || this.rundown.isActive) {
+    if (this.rundown.isActive) {
```
andr9528, line 55: At what point is this check performed, while the rundown is active?

andr9528: Now that i think a bit about it, it might be needed, incase the Rundown is activated via another frontend.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
andr9528: Now that i think a bit about it, it might be needed, incase the Rundown is deactivated via another frontend.

KvelaGorrrrnio: Another frontend should trigger update of the rundown entity used here, so this check should be unnecessary. I have removed the check and tested that it still works.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
KvelaGorrrrnio: Another frontend should trigger update of the rundown entity used here, so this check should be unnecessary. I have removed the check and tested that it still works.

andr9528: Testing on develop, and found that the context menu updates if the rundown active state changes while it is open, thus the if check should not be needed, ifi have my logic right.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
andr9528: Testing on develop, and found that the context menu updates if the rundown active state changes while it is open, thus the if check should not be needed, ifi have my logic right.

src/app/rundown-view/components/part-context-menu/part-context-menu.component.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      declarations: [PartContextMenuComponent],
+      providers: [{ provide: RundownService, useValue: instance(mock<RundownService>()) }],
+    }).compileComponents()
+
+    fixture = TestBed.createComponent(PartContextMenuComponent)
+    component = fixture.componentInstance
+    const testEntityFactory: TestEntityFactory = new TestEntityFactory()
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Not really. I am planning on doing it, but in a separate PR. The Virtual Keyboard some stuff for this class, so it will most likely trigger a cascade of merge conflicts to do it in this one :grin:

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
KvelaGorrrrnio: Regarding the context menu. I would very much like to remove these components and have a single context menu component, and a service which is responsible for setting the correct options. That will also fix the issue I showed you with multiple context menus popping up on activating a rundown. But I think it is out of scope for this PR.

PR #42, opened 2023-10-14
Listen for infinites and unplanned execution

KvelaGorrrrnio:
- Listens for infinite piece insertion.
- Listens for part inserted as on air and as next.


LindvedKrvang (approved):
Nice job :)

andr9528 (approved):

PR #43, opened 2023-10-16
SOF-1631: Key aliasing

KvelaGorrrrnio:
Introduces key aliasing, such that `Shift` is triggered by `ShiftLeft` and `ShiftRight`, `AnyEnter` is triggered by `Enter` and `NumpadEnter`, and so forth.

### DIFF
--- a/src/app/keyboard/services/tv2-key-alias.service.ts
+++ b/src/app/keyboard/services/tv2-key-alias.service.ts
@@ -1,0 +1,25 @@
+import { KeyAliasService } from '../abstractions/key-alias-service'
+
+export class Tv2KeyAliasService implements KeyAliasService {
+  public getAliasesForKey(key: string): string[] {
+    switch (key) {
+      case 'AnyEnter':
+        return ['Enter', 'NumpadEnter']
+      case 'Shift':
+        return ['ShiftLeft', 'ShiftRight']
+      case 'Control':
+        return ['ControlLeft', 'ControlRight']
+      case 'Alt':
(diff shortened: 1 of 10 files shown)


src/app/keyboard/services/tv2-key-alias.service.ts
```diff
@@ -0,0 +1,25 @@
+import { KeyAliasService } from '../abstractions/key-alias-service'
+
+export class Tv2KeyAliasService implements KeyAliasService {
+  public getAliasesForKey(key: string): string[] {
+    switch (key) {
+      case 'AnyEnter':
```
LindvedKrvang, line 6: Should we have enums for this or is it overkill?

LindvedKrvang (approved):

src/app/keyboard/services/tv2-key-alias.service.ts
KvelaGorrrrnio: I thought about it, and I think it is a good idea, but currently we have the rule about enum values to be uppercased snakecase and that enum keys should be the same as the values, which makes it impossible for this.

LindvedKrvang: We could make an exception for that rule?

KvelaGorrrrnio: I'll do it, but I'll base it on the virtual keyboard branch.

andr9528 (approved):

PR #47, opened 2023-10-19
Sof 1633: Virtual keyboard

KvelaGorrrrnio:
The primary focus of this the PR is the Virtual keyboard that can be seen in the picture below.
It shows an ISO 102 keyboard (without numpad). It uses displays key bindings with their respective color. Currently, only cameras specify a custom color. Modifier keys alters what is displayed on the keyboard. Clicking on a registered key binding will trigger all actions associated with the key.

Skærmbillede 2023-10-19 kl  20 35 12

Other things that have been done in the PR:
- The panel for static buttons has been setup in the producer shelf.
- Scrollbars in the rundown view have same color profile as the component they belong to.
- Logo in the regular header is correctly sized again.
- Services that are used for generating test content is moved to the new module `test`.
- Added missing Action types from backend.


src/app/test/factories/test-logger.factory.ts
```diff
@@ -0,0 +1,12 @@
+import { Logger } from '../../core/abstractions/logger.service'
+import { anyString, anything, instance, mock, when } from '@typestrong/ts-mockito'
+
+export class TestLoggerFactory {
+  public createLogger(): Logger {
```
andr9528, line 5: This could/should propable just be static, as it only meant to be used during tests.

RasmusAlbrektsen (approved):

src/app/test/factories/test-logger.factory.ts
LindvedKrvang: No, we just utilize that we use dependency injection.

src/app/keyboard/components/virtual-keyboard-key/virtual-keyboard-key.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  @Input()
+  @HostBinding('style.--key-weight')
+  public weight: number = 1
+
+  @Input()
+  public background?: string
+
+  @HostBinding('style.background')
+  public get displayBackground(): string {
+    return this.background ?? '#333333'
```
LindvedKrvang, line 37: Magic string

src/app/keyboard/components/virtual-keyboard-key/virtual-keyboard-key.component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  @Input()
+  public label?: string
+
+  @Input()
+  @HostBinding('class.matched')
+  public isMatched: boolean = false
+
+  @Input()
+  public keyboardLayout: KeyboardLayoutMap
```
LindvedKrvang, line 26: Do a 'key' really need to know the entire layout of the keyboad?
You are only using it to get the 'mappedKey'. Maybe the 'mappedKey' should just be an input?

It makes more sense that a layout knows about its keys. Just like it's our Segments that knows about its Parts, not the Part that knows about its Segment.

src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
```diff
@@ -1,31 +1,89 @@
...
 @Component({
   selector: 'sofie-virtual-keyboard',
   templateUrl: './virtual-keyboard.component.html',
   styleUrls: ['./virtual-keyboard.component.scss'],
 })
-export class VirtualKeyboardComponent {
+export class VirtualKeyboardComponent implements OnChanges {
   @Input()
-  public keyBindings: KeyBinding[]
+  public keyBindings: StyledKeyBinding[]
```
LindvedKrvang, line 16: The only difference between `KeyBinding` and `StyleKeyBinding` is that `StyledKyeBinding` has an optional `background` attribute.
Shouldn't this be `backgroundColor` since you store the color of the background in it.

src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
```diff
@@ -1,31 +1,89 @@
...
+    this.updatePhysicalKeyboardLayout()
+    navigator.keyboard
+      ?.getLayoutMap()
+      .then(keyboardLayoutMap => (this.keyboardLayoutMap = new Map(keyboardLayoutMap)))
+      .catch(error => this.logger.data(error).warn('Failed getting keyboard layout.'))
+      .finally(() => this.updatePhysicalKeyboardLayout())
+  }
+
+  private updatePhysicalKeyboardLayout(): void {
+    const physicalKeyboardLayoutService: PhysicalKeyboardLayoutService = new PhysicalKeyboardLayoutService(this.keyboardLayoutMap)
```
LindvedKrvang, line 43: Why is this not dependency injected?

LindvedKrvang: I assume it's because you parse in a constructor paramter.
The only thing you call in the constructor is `registerCustomKeyboardLayoutMapEntries()`.
You could just parse the parameter to `createIso102KeyboardLayout` and call the method as the first thing in there.

Sidenote: You have been making abstractions for everything else. Why not for the KeyboardLayoutService as well?
Then you can just call the method `createKeyboardLayout()` and have an `iso102KeyboardLayoutService` implementation. Then this component wouldn't even know that it's using a iso102 keyboard.

src/app/keyboard/services/physical-keyboard-layout.service.ts
```diff
@@ -0,0 +1,129 @@
+import { KeyboardLayoutKey, KeyboardLayout } from '../value-objects/keyboard-layout'
+
+export class PhysicalKeyboardLayoutService {
```
LindvedKrvang, line 3: Lots of magic strings in here.
But in this context I think it make sense to keep them as is ;) 

Sidenote: I like the use of a matrix to represent the keyboard :)

src/app/test/factories/test-logger.factory.ts
LindvedKrvang: And "it's used by tests" is not a good enough argument.

LindvedKrvang (changes requested):
It looks good.

The request for changes is for where you decided to bypass our dependency injection ;)

src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
KvelaGorrrrnio: The background can besides a solid color, be an image or a gradient.

src/app/keyboard/components/virtual-keyboard-key/virtual-keyboard-key.component.ts
KvelaGorrrrnio: Is removed.

src/app/keyboard/components/virtual-keyboard-key/virtual-keyboard-key.component.ts
KvelaGorrrrnio: The variable name is not very descriptive. It is not the layout per se, but a map of key codes to key values.
I have moved the mapping logic out into the virtual keyboard, causing this variable to be removed.

src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
KvelaGorrrrnio: PhysicalKeyboardLayoutService abstraction is added with the methods `registerKeyboardKeyLabels` and `getPhysicalKeyboardLayout` and DI is used.

src/app/keyboard/components/virtual-keyboard-key/virtual-keyboard-key.component.ts
```diff
@@ -8,11 +8,11 @@ import { ChangeDetectionStrategy, Component, HostBinding, Input } from '@angular
 })
 export class VirtualKeyboardKeyComponent {
   @Input()
-  public key: string
+  public keyLabel: string
 
   @HostBinding('class')
   public get keyClass(): string {
-    return this.key ? this.key.toLowerCase() : 'none'
+    return this.keyLabel === '<none>' ? 'none' : this.keyLabel.toLowerCase()
```
LindvedKrvang, line 15: Magic strings :)

LindvedKrvang (approved):

src/app/keyboard/components/virtual-keyboard-key/virtual-keyboard-key.component.ts
KvelaGorrrrnio: Magic string is now a constant :D

PR #49, opened 2023-10-23
SOF 1671: Mapped keys on key bindings

KvelaGorrrrnio:
- `StyledKeyBinding.mappedKeys` is added and simulates AHK key remapping.
- `StyledKeyBinding.mappedKeys` and `StyledKeyBindings.keys` are shown on the virtual keyboard.
- Default labels for keyboard is added, such that virtual keyboard keys says `A` instead of `KeyA`.
- All services use the environment configuration for http calls.
- Local development uses hostname from URL for determining API urls. This allows to test local frontend (and backend) from other computer on same network.


src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
```diff
@@ -53,34 +61,55 @@ export class VirtualKeyboardComponent implements OnChanges {
...
   }
 
+  private doesModifiersMatchExclusively(keyBinding: StyledKeyBinding): boolean {
+    const modifierKeysForKeyBinding: string[] = keyBinding.keys.filter(key => this.keyAliasService.isModifierKeyOrAliasedModifierKey(key))
+
+    if (modifierKeysForKeyBinding.length !== this.currentModifierKeystrokes.length) {
+      return false
+    }
+
+    // TODO: Respect key order and exclusivity
```
LindvedKrvang, line 78: Is this planned to be part of this PR?

src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
```diff
@@ -39,6 +43,10 @@ export class VirtualKeyboardComponent implements OnChanges {
       .finally(() => this.updatePhysicalKeyboardLayout())
   }
 
+  private setFallbackKeyboardLayoutMap(): void {
+    this.keyboardLayoutMap = this.defaultKeyboardLayoutMapService.createDefaultKeyboardLayoutMap()
+  }
```
LindvedKrvang, line 48: If you used dependency injection for the `PhysicalKeyboardLayoutService` then you wouldn't need this. The `PhysicalKeyboardLayoutService` would just itself have an `KeyboardLayoutMapService` injected to it.
This way `VirtualKeyboardComponent` does not need to know about `DefaultKeyboardLayoutMapService`, it just need to call `KeyboardLayoutService.createKeyboardLayout()`.

src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
KvelaGorrrrnio: I was not planning on doing it in this PR, since tit is not something that we use, but something that other parties might be interested in.

src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
KvelaGorrrrnio: This part is removed and not using DI is fixed in #47.

LindvedKrvang (approved):

andr9528 (approved):

PR #53, opened 2023-11-01
SOF-1665: Reload Ingest Data Option in Header Context Menu

RasmusAlbrektsen:
This PR adds the option to "Reingest data" from the Header Context Menu using the endpoints implemented in the following PR's:
https://github.com/tv2/inews-ftp-gateway/pull/88
https://github.com/tv2/sofie-server/pull/38


LindvedKrvang (approved):

src/app/core/services/http-ingest.service.ts
```diff
@@ -0,0 +1,21 @@
...
+const INGEST_URL: string = `${environment.apiBaseUrl}/ingest-operations`
+
+@Injectable()
+export class HttpIngestService implements IngestService {
+  constructor(
+    private readonly http: HttpClient,
+    private readonly httpErrorService: HttpErrorService
+  ) {}
+
+  public reingestRundownData(rundownId: string): Observable<void> {
```
KvelaGorrrrnio, line 17: I don't think the concept of ingest is necessary for this feature as mentioned in Sofie server. This is still an action on the rundown, and can be moved to RundownService (HttpRundownService).

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -41,6 +42,10 @@ export class RundownHeaderContextMenuComponent {
       label: 'Reset Rundown',
       contextAction: (): void => this.openResetRundownDialog(),
     },
+    {
+      label: 'Reingest data',
+      contextAction: (): void => this.reingestData(),
```
KvelaGorrrrnio, line 47: Instead of calling it reingestData, won't `reingestRundown` be more accurate?
Same comment for the label above.

RasmusAlbrektsen: This was the label that Jacob decided on since it should also signify that you load in configuration changes. Is at least what is written on the task. https://tv2cms.atlassian.net/jira/software/c/projects/SOF/boards/784?selectedIssue=SOF-1665

src/app/core/services/http-ingest.service.ts
RasmusAlbrektsen: The concept has been removed and the operation is on the RundownService instead.

LindvedKrvang (approved):

KvelaGorrrrnio (approved):

PR #54, opened 2023-11-02
Sof 1628/remote actions

KvelaGorrrrnio:
- Remote actions are displayed in the ALL ACTIONS container.
- Camera key bindings are moved from digit keys to function keys.
- Remote actions are bound to digit keys.

### DIFF
--- a/src/app/rundown-view/factories/key-binding.factory.ts
+++ b/src/app/rundown-view/factories/key-binding.factory.ts
@@ -15,6 +15,9 @@ import { Logger } from '../../core/abstractions/logger.service'
 const CAMERA_COLOR: string = '#005919'
 const REMOTE_COLOR: string = '#ac29a5'
 
+const FUNCTION_KEY_PREFIX: string = 'F'
+const DIGIT_KEY_PREFIX: string = 'Digit'
+
 @Injectable()
 export class KeyBindingFactory {
   private readonly logger: Logger
@@ -46,7 +49,7 @@ export class KeyBindingFactory {
       throw new Error('Expected camera number to be an integer.')
     }
     return {
-      keys: [`F${cameraNumber}`],
+      keys: [`${FUNCTION_KEY_PREFIX}${cameraNumber}`],
       label: cameraAction.name,
       onMatched: () => this.actionService.executeAction(cameraAction.id, rundownId).subscribe(),
       shouldMatchOnKeyRelease: true,
@@ -64,7 +67,7 @@ export class KeyBindingFactory {
       throw new Error('Expected camera number to be an integer.')
     }
     return {
-      keys: ['Alt', `F${cameraNumber}`],
+      keys: ['Alt', `${FUNCTION_KEY_PREFIX}${cameraNumber}`],
       label: cameraAction.name,
       onMatched: () => this.actionService.executeAction(cameraAction.id, rundownId).subscribe(),
       shouldMatchOnKeyRelease: true,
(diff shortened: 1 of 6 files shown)


src/app/rundown-view/factories/key-binding.factory.ts
```diff
@@ -12,6 +12,9 @@ import { RundownNavigationService } from '../../shared/services/rundown-navigati
 import { RundownCursor } from '../../core/models/rundown-cursor'
 import { Logger } from '../../core/abstractions/logger.service'
 
+const CAMERA_COLOR: string = '#005919'
+const REMOTE_COLOR: string = '#ac29a5'
```
LindvedKrvang, line 16: Here a comment about what colour it is would be nice.
So it would be `const CAMERA_COLOR: string = '#005919' // Green`

src/app/rundown-view/factories/key-binding.factory.ts
```diff
@@ -38,32 +41,61 @@ export class KeyBindingFactory {
...
 
   private createInsertCameraOnAirKeyBinding(cameraAction: Tv2CameraAction, rundownId: string): StyledKeyBinding {
-    const cameraNumber: number = cameraAction.metadata.cameraNumber
+    const cameraNumber: number = parseInt(cameraAction.metadata.cameraNumber)
+    if (Number.isNaN(cameraNumber)) {
+      throw new Error('Expected camera number to be an integer.')
+    }
     return {
-      keys: [`Digit${cameraNumber}`],
+      keys: [`F${cameraNumber}`],
```
LindvedKrvang, line 49: I assume `F` is the function key prefix?
Could be nice to have that extracted to a const.

src/app/rundown-view/services/tv2-action-group.service.ts
```diff
@@ -14,6 +15,8 @@ export class Tv2ActionGroupService {
   private groupActionsByContentTypeReducer(grouping: Tv2ActionContentTypeGrouping, action: Tv2Action): Tv2ActionContentTypeGrouping {
     if (this.isCameraAction(action)) {
       grouping.camera.push(action)
+    } else if (this.isRemoteAction(action)) {
```
LindvedKrvang, line 18: Couldn't this just be a `switch`?
All these `isX` methods just looks at the `contentType` which all are a `Tv2ActionContentType`.

LindvedKrvang (approved):

src/app/rundown-view/services/tv2-action-group.service.ts
RasmusAlbrektsen: Agreed, i'd prefer a switch over this.

RasmusAlbrektsen (approved):

src/app/rundown-view/factories/key-binding.factory.ts
KvelaGorrrrnio: Comments on color variables are destined to become outdated. When piece rendering is done, these values will be changed to css variables, such that this becomes something like:
```typescript
const CAMERA_COLOR: string = 'var(--camera-color)'
```

src/app/rundown-view/factories/key-binding.factory.ts
KvelaGorrrrnio: Correct. The `F` is extracted to the constant `FUNCTION_KEY_PREFIX`.

src/app/rundown-view/services/tv2-action-group.service.ts
KvelaGorrrrnio: Changed to switch case with type assertions. Using type guards give greater flexibility in when we want to acknowledge an action as the specified type – e.g. if a certain metadata attribute is set correctly.

PR #57, opened 2023-11-06
New ActionType.INSERT_PIECE_AS_NEXT_AND_TAKE

LindvedKrvang:
In https://github.com/tv2/sofie-server/pull/41 a new PieceType was introduced.
This PR is to add it to the frontend


src/app/shared/models/action-type.ts
```diff
@@ -8,5 +8,6 @@ export enum PartActionType {
 export enum PieceActionType {
   INSERT_PIECE_AS_ON_AIR = 'INSERT_PIECE_AS_ON_AIR',
   INSERT_PIECE_AS_NEXT = 'INSERT_PIECE_AS_NEXT',
+  INSERT_PIECE_AS_NEXT_AND_TAKE = 'INSERT_PIECE_AS_NEXT_AND_TAKE',
```
KvelaGorrrrnio, line 11: The change is fine, but a question regarding the general structure and meaning of the piece enumerations.
We write `INSERT_PIECE_AS_X` (e.g. `INSERT_PIECE_AS_ON_AIR`, but pieces are not standalone and can technically not bet inserted as X.
Wouldn't it be more accurate if the piece enumerations have the form `INSERT_PIECE_IN_X_PART` (e.g. `INSERT_PIECE_IN_ON_AIR_PART` and `INSERT_PIECE_IN_NEXT_PART_AND_TAKE_NEXT`).

KvelaGorrrrnio (approved):

src/app/shared/models/action-type.ts
LindvedKrvang: I guess this comes down to the question: How explicit do we want to be?
If the majority is in favour of also adding `PART` to it, then I don't have any arguments against it.

andr9528: Isn't it implied that when inserting a piece as on air, it is inserted on a part?
Same thing for `TAKE`. When is a `TAKE` not a `TAKE_NEXT`.

I personally think the shorter version that Lindved used, is better. 
An small added benefit is that it is also easier to pronounce, if one need to discus it.

andr9528 (approved):

PR #59, opened 2023-11-06
SOF-1623: Piece types and output layers on TV2 piece metadata

KvelaGorrrrnio:
Note: Depends on https://github.com/tv2/sofie-server/pull/35.
- `Tv2Piece.metadata.type: Tv2PieceType` added.
- `Tv2Piece.metadata.outputLayer: Tv2OutputLayer` added.
- `Piece.type` removed.
- The computation for expected duration of parts with auto next is improved – but not entirely correct yet.


LindvedKrvang (approved):

src/app/shared/services/tv2-output-layer.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Part } from '../../core/models/part'
+import { Tv2OutputLayer } from '../../core/models/tv2-output-layer'
+import { Tv2Piece } from '../../core/models/tv2-piece'
+
+export class Tv2OutputLayerService {
```
andr9528, line 5: Do we technically need the `Tv2...` part of the name? 
All of the frontend i technically ours.
Same goes for other types which has had `Tv2` prefixed to it.

KvelaGorrrrnio: For now we want to keep it easy to fork and start working on for other Sofie collaborators. There are some things that need clearer separation for this to ideal, but having the Tv2 prefix for our specific code helps with that. And it keeps it the types matching between the frontend and blueprints.

andr9528 (approved):

PR #64, opened 2023-11-10
SOF-1682: End words

KvelaGorrrrnio:
- End words added to rundown view.
- Placeholders for part countdown, segment countdown and segment timer added to rundown view.

Skærmbillede 2023-11-10 kl  22 54 59

### DIFF
--- a/src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.html
+++ b/src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.html
@@ -1,0 +1,23 @@
+<sofie-spacer></sofie-spacer>
+
+<div class="subheader-panel">
+    <span class="subheader-panel__label">Part countdown</span>
+    <span class="subheader-panel__value">00:04</span>
+</div>
+
+<div class="subheader-panel">
+    <span class="subheader-panel__label">Segment countdown</span>
+    <span class="subheader-panel__value">00:12</span>
+</div>
+
+<div class="subheader-panel">
+    <span class="subheader-panel__label">Segment timer</span>
+    <span class="subheader-panel__value">00:08</span>
+</div>
+
+<div class="subheader-panel end-words">
+    <span class="subheader-panel__label">End words</span>
+    <span class="subheader-panel__value">{{ endWords }}</span>
+</div>
+
+<sofie-spacer></sofie-spacer>
--- a/src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.ts
+++ b/src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.ts
@@ -1,0 +1,50 @@
+import { ChangeDetectionStrategy, Component, Input, OnChanges, SimpleChange, SimpleChanges } from '@angular/core'
+import { Rundown } from '../../../core/models/rundown'
+import { Piece } from '../../../core/models/piece'
+import { Tv2PieceMetadata } from '../../../core/models/tv2-piece'
+import { Tv2PieceType } from '../../../core/enums/tv2-piece-type'
+import { Part } from '../../../core/models/part'
+
+@Component({
+  selector: 'sofie-rundown-subheader',
+  templateUrl: './rundown-subheader.component.html',
+  styleUrls: ['./rundown-subheader.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class RundownSubheaderComponent implements OnChanges {
+  @Input()
+  public rundown: Rundown
+
+  public endWords?: string
+
+  public ngOnChanges(changes: SimpleChanges): void {
+    const rundownChange: SimpleChange = changes['rundown']
+    if (rundownChange) {
+      this.setEndWords()
+    }
+  }
+
+  private setEndWords(): void {
+    this.endWords = this.rundown.segments
+      .find(segment => segment.isOnAir)
+      ?.parts
+      .find(part => part.isOnAir && this.doesPartContainVideoClipPiece(part))
+      ?.pieces
+      .find(piece => this.isManusPiece(piece))
+      ?.name
+  }
+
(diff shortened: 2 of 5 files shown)


LindvedKrvang:
I'm not mistaken if there is yet to be a PR for `SOF-1683/rundown-timings`?
It's a bit weird to review the "follow-up" PR, before the "main" PR.

KvelaGorrrrnio:
> I'm not mistaken if there is yet to be a PR for `SOF-1683/rundown-timings`? It's a bit weird to review the "follow-up" PR, before the "main" PR.

There is going to be a PR for SOF-1683/rundown-timings. But the functionality is non-overlapping.

RasmusAlbrektsen:
> > I'm not mistaken if there is yet to be a PR for `SOF-1683/rundown-timings`?
> 
> There is going to be a PR for SOF-1683/rundown-timings.

Yeah, i'll throw it up as a draft so it is easier to recognize at the moment.

src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  @Input()
+  public rundown: Rundown
+
+  public endWords?: string
+
+  public ngOnChanges(changes: SimpleChanges): void {
+    const rundownChange: SimpleChange = changes['rundown']
+    if (rundownChange) {
+      this.setEndWords()
+    }
```
andr9528, line 24: Updating type for `rundownChange` to include the possibility of it being undefined would help explain why there is an undefiend check.
At the moment you are checking that a value is not undefined, which doesn't appear from the type to even have the possibility.

src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  private setEndWords(): void {
+    this.endWords = this.rundown.segments
+      .find(segment => segment.isOnAir)
+      ?.parts.find(part => part.isOnAir && this.doesPartContainVideoClipPiece(part))
+      ?.pieces.find(piece => this.isManusPiece(piece))?.name
+  }
```
andr9528, line 32: Do i understand this right, that you can't have a part with a manus piece that doesn't also contain a video clip piece?

KvelaGorrrrnio: You can have a part with a manus piece but not a video clip piece – but it should not be shown as end words.

src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.ts
KvelaGorrrrnio: Nice catch – I have updated it.

src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.ts
andr9528: Okay that explains the logic in place here. Thanks.

andr9528 (approved):

src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.html
```diff
@@ -0,0 +1,23 @@
...
+    <span class="subheader-panel__label">Segment timer</span>
+    <span class="subheader-panel__value">00:08</span>
+</div>
+
+<div class="subheader-panel end-words">
+    <span class="subheader-panel__label">End words</span>
+    <span class="subheader-panel__value">{{ endWords }}</span>
+</div>
+
+<sofie-spacer></sofie-spacer>
```
LindvedKrvang, line 23: There is not really anything in here that is header specific, so there is no reason to call this `subheader`. 
This indicates that you have named the component from where you going to use it, instead of what it does.

What it does is display varius timings and the end words. 
If you can't find a fitting name for that then maybe that shows that this shouldn't a component on its own? Or at the very least, this is the wrong place to separate the component?

A further indications that this seems to be developed with where it is going to be used instead of what it does is the usages of `SofieSpacer`.
Why do the component need to pre- and postpend space to itself? My guess is to make it look nice where it is used. But the component is not allowed to know anything about where it is used...

src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Tv2PieceType } from '../../../core/enums/tv2-piece-type'
+import { Part } from '../../../core/models/part'
+
+@Component({
+  selector: 'sofie-rundown-subheader',
+  templateUrl: './rundown-subheader.component.html',
+  styleUrls: ['./rundown-subheader.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class RundownSubheaderComponent implements OnChanges {
```
LindvedKrvang, line 14: Again, nothing in here that is header specific, so `RundownSubheaderComponent` seems to be the wrong name.

src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.html
KvelaGorrrrnio: It is not header specific, but subheader specific ;) And on a serious note: I agree that subheader is a bad. I have changed it to `OnAirDetailsPanel`.

A note regarding the spacer. It is not padding the component itself. The spacer is placed inside the component and is used to restrict the space available to the sub panels.

src/app/rundown-view/components/rundown-subheader/rundown-subheader.component.ts
KvelaGorrrrnio: Updated to `OnAirDetailsPanelComponent`.

LindvedKrvang (approved):

PR #65, opened 2023-11-13
Sof 1683/Rundown Timing

RasmusAlbrektsen:
This PR creates the basic framework for rundown timing and implements the `Planned start`, `Diff` and `Planned end` in the Rundown Header.


andr9528 (approved):

src/app/core/services/rundown-timing.service.spec.ts
```diff
@@ -0,0 +1,428 @@
...
+              const expectedResult: number = 0
+
+              jasmine.clock().tick(playedDurationInMsForOnAirPart)
+
+              const result: number = testee.getRundownScheduleOffsetInMs(rundown)
+
+              expect(result).toBe(expectedResult)
+            })
+          })
+          describe('when next segment is after the on air segment', () => {
```
LindvedKrvang, line 374: A new line after a test would be greatly appreciated. There's quite a few places in this file where there is no new line after the test.

src/app/core/services/rundown-timing.service.ts
```diff
@@ -0,0 +1,125 @@
+import { BackwardRundownTiming, RundownTiming } from '../models/rundown-timing'
+import { RundownTimingType } from '../enums/rundown-timing-type'
+import { Rundown } from '../models/rundown'
+import { Segment } from '../models/segment'
+import { PartEntityService } from './models/part-entity.service'
+import { Part } from '../models/part'
+
+export class RundownTimingService {
+  private readonly partEntityService: PartEntityService = new PartEntityService()
```
LindvedKrvang, line 9: You are bypassing our dependency injection here. That's a big no no ;) 
Also, `RundownTimingService` is also not exported to be dependency injected.

src/app/core/services/rundown-timing.service.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (backwardRundownTiming.expectedDurationInMs) {
+      return backwardRundownTiming.expectedEndEpochTime - backwardRundownTiming.expectedDurationInMs
+    }
+
+    return undefined
+  }
+
+  public getExpectedDurationInMs(rundownTiming: RundownTiming): number | undefined {
+    return rundownTiming.expectedDurationInMs
+  }
```
LindvedKrvang, line 36: Is this method really needed?
Why can't the caller just do `rundownTiming.expectedDurationInMs` themselves?

src/app/core/services/rundown-timing.service.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  private getAggregatedDurationInMs(rundown: Rundown): number {
+    return rundown.timing.expectedDurationInMs ?? rundown.segments.reduce((accumulatedDurationInMs, segment) => accumulatedDurationInMs + this.getSegmentDurationInMs(segment), 0)
+  }
+
+  private getSegmentDurationInMs(segment: Segment): number {
+    return segment.budgetDuration !== undefined
+      ? segment.budgetDuration
+      : segment.parts.reduce((accumulatedPartDuration: number, part: Part) => accumulatedPartDuration + this.getPartDuration(part), 0)
```
LindvedKrvang, line 66: Would this give the same result?
```
return segment.budgetDuration ?? segment.parts.reduce((accumulatedPartDuration: number, part: Part) => accumulatedPartDuration + this.getPartDuration(part), 0)
```

src/app/core/services/rundown-timing.service.ts
```diff
@@ -0,0 +1,125 @@
...
+      .reduce((accumulatedPartDuration: number, part: Part) => accumulatedPartDuration + this.partEntityService.getDuration(part), 0)
+    const onAirPartPlayedDuration: number = this.partEntityService.getPlayedDuration(onAirSegment.parts[onAirPartIndex])
+    return playedPartDuration + onAirPartPlayedDuration
+  }
+
+  private getPartDuration(part: Part): number {
+    if (part.isUntimed) {
+      return 0
+    }
+    return part.playedDuration > 0 ? part.playedDuration : part.expectedDuration ?? 0
```
LindvedKrvang, line 123: Be carefull when mixing more operations into ternaries. They quickly become obscure to read.
I think the use here is fine, but it's a slippery slope, hence the comment :)

src/app/rundown-view/components/rundown-header/rundown-header.component.ts
```diff
@@ -6,8 +6,9 @@ import { ShowStyleVariant } from '../../../core/models/show-style-variant'
 import { SubscriptionLike } from 'rxjs'
 import { IconButton, IconButtonSize } from '../../../shared/enums/icon-button'
 import { Logger } from '../../../core/abstractions/logger.service'
+import { RundownTimingService } from '../../../core/services/rundown-timing.service'
 
-const UPDATE_LOCAL_CLOCK_INTERVAL: number = 1000
+const TIME_RESOLUTION_INTERVAL: number = 1000
```
LindvedKrvang, line 11: Is once a second enough?

Also, there is no indication whether this is miliseconds or something else

src/app/rundown-view/components/segment/segment.component.html
```diff
@@ -6,7 +6,7 @@
   <div [cdkContextMenuTriggerFor]="contextMenu" class="c-segment__header">
     <div class="c-segment__header-toolbar">
       <span>123</span>
```
LindvedKrvang, line 8: Wasn't this removed?

src/app/rundown-view/pipes/timer/timer.pipe.ts
```diff
@@ -0,0 +1,27 @@
...
+})
+export class TimerPipe implements PipeTransform {
+  public transform(durationInMs: number): string {
+    const durationInSeconds: number = Math.floor(durationInMs / 1000)
+    const absoluteDurationInSeconds: number = Math.abs(durationInSeconds)
+    const seconds: number = absoluteDurationInSeconds % 60
+    const minutes: number = Math.floor((absoluteDurationInSeconds % 3600) / 60)
+    const hours: number = Math.floor(absoluteDurationInSeconds / 3600)
+    const sign: string = this.getSign(durationInSeconds)
+    return `${sign}${this.formatDigits(hours)}:${this.formatDigits(minutes)}:${this.formatDigits(seconds)}`
```
LindvedKrvang, line 17: We should really try to avoid doing manually time manipualtion since it's a field full of egdecases and many good libraries exits for it out there.

I will highly recommed us using the library `date-fns` for everything timing related.
I believe their method `formatDuration` will do what you are trying to achieve here (minus the positive/negative sign, but you can always prepend that).
https://date-fns.org/v2.30.0/docs/formatDuration

LindvedKrvang: Also, this Pipe is closely related to `TimestampPipe` in the form that is doing close to the same logic.
What are the differences between `TimerPipe` and `TimestampPipe`?
From their names alone I wouldn't know which one to use where.
This should be cleaned up.

LindvedKrvang (changes requested):
Good work.

You are bypassing our dependency injection, hence the request for changes.
Also, the new `TimerPipe` is very similar to `TimestampPipe` and both do timings manipulation, which I think we should avoid having to do manually.

src/app/core/services/rundown-timing.service.spec.ts
```diff
@@ -0,0 +1,428 @@
...
+  afterEach(() => jasmine.clock().uninstall())
+
+  describe(RundownTimingService.prototype.getRundownScheduleOffsetInMs.name, () => {
+    describe('when rundown is inactive', () => {
+      describe('when planned end is in the future', () => {
+        it('return the time difference between estimated end and planned end', () => {
+          const durationInMsToPlannedEndFromNow: number = 10_000
+          const rundownExpectedDurationInMs: number = 30_000
+          const numberOfSegments: number = 4
+          const segments: Segment[] = [...Array(numberOfSegments)].map((_, index) =>
```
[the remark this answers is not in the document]
RasmusAlbrektsen: I have changed it where possible! :-)

src/app/core/services/rundown-timing.service.spec.ts
RasmusAlbrektsen: Newlines inserted! My bad.

src/app/core/services/rundown-timing.service.ts
RasmusAlbrektsen: Dependecy injection is now used!

src/app/core/services/rundown-timing.service.ts
RasmusAlbrektsen: A good point :-)

src/app/core/services/rundown-timing.service.ts
RasmusAlbrektsen: I see how it might look weird, but it is to not mix up the way we handle `RundownTiming` when for instance it is used by `BasicRundowns`. 
When they need the planned end or start (located in the `RundownTiming`), we first have to check which type of time calculation we are dealing with: `FORWARD`, `BACKWARD` og `UNSCHEDULED`. Though this is not necessary for the `expectedDurationInMs`, i think it would be less than ideal if we handled some of this inside a service and some of it outside.
We might even change how this is implemented, and in that case it would be nice to have it wrapped in the service.

src/app/rundown-view/pipes/timer/timer.pipe.ts
RasmusAlbrektsen: This will be changed in SOF-1680, the idea is for `Timer` to replace `Timestamp` and have the specific logic only for our own needs. I checked `date-fns` it does not seem to easily cover our usecases, but i'm not against using a library if we find one that fits.

src/app/rundown-view/components/rundown-header/rundown-header.component.ts
RasmusAlbrektsen: This will be deleted in SOF-1680.

src/app/rundown-view/components/segment/segment.component.html
RasmusAlbrektsen: Might have been, i had no idea this was there. It is removed in SOF-1680!

src/app/core/services/rundown-timing.service.ts
RasmusAlbrektsen: This is changed in SOF-1680, but it might even be removed.

src/app/rundown-view/pipes/timer/timer.pipe.ts
LindvedKrvang: And what is the specific usecase?

src/app/test/factories/test-entity.factory.ts
```diff
@@ -2,6 +2,7 @@ import { Part } from '../../core/models/part'
 import { Segment } from '../../core/models/segment'
 import { Rundown } from '../../core/models/rundown'
 import { Piece } from '../../core/models/piece'
+import { RundownTimingType } from '../../core/enums/rundown-timing-type'
 
 export class TestEntityFactory {
   public createRundown(rundown: Partial<Rundown> = {}): Rundown {
```
LindvedKrvang, line 8: All the methods in here is only used for creating test entities in our tests.
This means that the public methods can be static. Then you don't have to instantiate a new `TesteEntityFactory` in every single test. (We do the same in the backend ;) )

LindvedKrvang (approved):

src/app/rundown-view/pipes/timer/timer.pipe.ts
LindvedKrvang: Just for the record: I'm very much not a fan of us having to do "time manipulation" manually.

PR #67, opened 2023-11-20
SOF-1680: On air timers

KvelaGorrrrnio:
_**Note**: Is depending on https://github.com/tv2/sofie-server/pull/56._
- Part countdown added to OnAirDetailsPanel.
- Segment countdown added to OnAirDetailsPanel.
- On air segment timer added to OnAirDetailsPanel.
- State services uses Observables instead of subscriptions as return type.
- State services checks subject listener count on new data, and cleans up if there are no listeners. As a result, ManagedSubscription has been removed.
- Calculating rundown and segment timings are centralized in `RundownTimingContextStateService`. In doing so, the RundownTimingService is using precomputed values for e.g. segment expected durations.
- `TimestampPipe` is removed in favor of `TimerPipe`, which allows for custom format strings.
- `Segment.budgetDuration` is renamed to `.expectedDurationInMs`.


src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -0,0 +1,187 @@
...
+import { RundownTimingService } from './rundown-timing.service'
+import { RundownTimingContext } from '../models/rundown-timing-context'
+import { Injectable } from '@angular/core'
+import { Segment } from '../models/segment'
+
+const HIGH_RESOLUTION_INTERVAL_DURATION_IN_MS: number = Math.floor(1000 / 30)
+const LOW_RESOLUTION_INTERVAL_DURATION_IN_MS: number = Math.floor(1000 / 4)
+
+@Injectable()
+export class RundownTimingContextStateService {
```
LindvedKrvang, line 13: FYI: The method order in this class is not a "linear read". 
The first usage of a method we encounter is `startLowResolutionTimer` from the constructor, but it's delcared as the fourth method. Similarly, `startHighResolutionTimer` is a private method that is declared before its first usage.

src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -0,0 +1,187 @@
...
+        currentEpochTime,
+        expectedStartEpochTimeForRundown,
+        expectedEndEpochTimeForRundown,
+        playedDurationInMsForOnAirPart,
+        playedDurationInMsForOnAirSegment,
+        durationInMsSpentInOnAirSegment,
+        remainingDurationInMsForRundown: remainingDurationInMs,
+      }
+      rundownTimingContextSubject.next(rundownTimingContext)
+    })
```
LindvedKrvang, line 72: Most of the values returned from the different methods are put directly into a `RundownTimingContext`. Why can't `RundownTimingService` just have a `getRundownTimingContext()` method? The only value in the `RundownTimingContext` that is not parsed to `RundownTimingService` is `rundownTimingContextSubject.value`.

So you could have something similar to:
```
const rundownTimingContext: RundownTimingContext = {
  ...rundownTimingContextSubject.value,
  ...rundownTimingService.getRundownTimingContext(rundown, currentEpochTime, previousRundownTimingContext.expectedDurationInMsForRundown)
}
```

src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    const hasActiveRundown: boolean = [...this.rundowns.values()].some(rundown => rundown.isActive)
+    if (hasActiveRundown && !this.isHighResolutionTimer) {
+      this.startHighResolutionTimer()
+    } else if (!hasActiveRundown && this.isHighResolutionTimer) {
+      this.startLowResolutionTimer()
+    }
```
LindvedKrvang, line 79: Do you really need to check for `isHighResolutionTimer`?

src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -0,0 +1,187 @@
...
+import { Injectable } from '@angular/core'
+import { Segment } from '../models/segment'
+
+const HIGH_RESOLUTION_INTERVAL_DURATION_IN_MS: number = Math.floor(1000 / 30)
+const LOW_RESOLUTION_INTERVAL_DURATION_IN_MS: number = Math.floor(1000 / 4)
+
+@Injectable()
+export class RundownTimingContextStateService {
+  private timeResolutionTimerId?: ReturnType<typeof setTimeout>
+  private isHighResolutionTimer: boolean = false
```
LindvedKrvang, line 15: This attribute feels unnecesary. Wouldn't it be enough to just check whether there is an active Rundown or not?

src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -0,0 +1,187 @@
...
+    this.stopTimeResolutionTimer()
+    this.isHighResolutionTimer = false
+    this.timeResolutionTimerId = setInterval(this.onTimeTick.bind(this), LOW_RESOLUTION_INTERVAL_DURATION_IN_MS)
+  }
+
+  private stopTimeResolutionTimer(): void {
+    if (this.timeResolutionTimerId === undefined) {
+      return
+    }
+    this.isHighResolutionTimer = false
```
LindvedKrvang, line 92: This seems unnecessary. Both your "startTimers" set this when you start them.
This is only necessary because of your logic in `onTimeTick()`.

src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -0,0 +1,187 @@
...
+      expectedDurationInMsForRundown: expectedDurationInMsForRundown,
+      expectedStartEpochTimeForRundown: expectedStartEpochTimeForRundown,
+      expectedEndEpochTimeForRundown: expectedEndEpochTimeForRundown,
+      remainingDurationInMsForRundown: remainingDurationInMsForRundown,
+      playedDurationInMsForOnAirSegment,
+      durationInMsSpentInOnAirSegment,
+      playedDurationInMsForOnAirPart,
+      expectedDurationsInMsForSegments,
+      startOffsetsInMsFromNextCursorForSegments: {},
+    }
```
LindvedKrvang, line 155: This is close to be a duplication of a good chunk of `onTimeTick()`. My comment there would also work for this.

src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -0,0 +1,187 @@
...
+    rundownTimingContextSubject.next(rundownTimingContext)
+    this.rundowns.set(rundown.id, rundown)
+  }
+
+  private getRundownTimingContextSubject(rundownId: string): BehaviorSubject<RundownTimingContext> | undefined {
+    const rundownTimingContextSubject: BehaviorSubject<RundownTimingContext> | undefined = this.rundownTimingContextSubjects.get(rundownId)
+    if (!rundownTimingContextSubject) {
+      return
+    }
+    const { wasRemoved } = this.removeSubjectIfHasNoObserversOrSubscriptions(rundownTimingContextSubject, rundownId)
```
LindvedKrvang, line 165: Same comment as for the other place you do this.

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.html
```diff
@@ -1,23 +1,23 @@
 <sofie-spacer></sofie-spacer>
 
-<div class="sub-panel">
-    <span class="sub-panel__label">Part countdown</span>
-    <span class="sub-panel__value">00:04</span>
+<div class="subpanel">
```
LindvedKrvang, line 3: The class is called `on-air-details-panel`. It doesn't know that it is being used as a `subpanel`. That's a detail of where it's being used, not a detail the component itself should know about.

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.html
```diff
@@ -1,23 +1,23 @@
...
+    <span class="subpanel__label">Part countdown</span>
+    <span class="subpanel__value"><span *ngIf="remainingDurationInMsForOnAirPart !== undefined" [class.overtime]="remainingDurationInMsForOnAirPart > 0" class="timer">{{ remainingDurationInMsForOnAirPart | timer: '#+HH?:mm:ss' }}</span></span>
```
LindvedKrvang, line 5: Not sure what to think about us having to supply `'#+HH?:mm:ss'`. It would be nice if that was just an implementation detail.

LindvedKrvang, line 5: Double spans?

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.ts
```diff
@@ -1,27 +1,97 @@
...
+  private updateOnAirPartTiming(rundownTimingContext: RundownTimingContext): void {
+    if (!this.onAirPart || !this.doesPartContainVideoClipOrVoiceOver(this.onAirPart)) {
+      this.remainingDurationInMsForOnAirPart = undefined
+    } else {
+      this.remainingDurationInMsForOnAirPart = rundownTimingContext.playedDurationInMsForOnAirPart - this.partEntityService.getExpectedDuration(this.onAirPart)
+    }
+  }
+
+  private doesPartContainVideoClipOrVoiceOver(part: Part): boolean {
+    const supportedPieceTypes: (string | undefined)[] = [Tv2PieceType.VIDEO_CLIP, Tv2PieceType.VOICE_OVER]
```
LindvedKrvang, line 62: Why is the type `(string | undefined)[]`?
You are initializing the array to enum values, when are they ever undefined?

It looks like you have changed your type because of how you want to "query" the array on line 63. Same principle apply here as with naming: Call it what it is, not how you want to use it. In this case it's an array of `TvPieceType`.

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.ts
```diff
@@ -1,27 +1,97 @@
...
+      const expectedDurationInMsForOnAirSegment: number = rundownTimingContext.expectedDurationsInMsForSegments[this.onAirSegment.id] ?? 0
+      this.remainingDurationInMsForOnAirSegment = rundownTimingContext.playedDurationInMsForOnAirSegment - expectedDurationInMsForOnAirSegment
+    }
+
+    if (!this.onAirSegment?.executedAtEpochTime) {
+      this.durationInMsSpentInOnAirSegment = undefined
+    } else {
+      this.durationInMsSpentInOnAirSegment = rundownTimingContext.durationInMsSpentInOnAirSegment
+    }
+  }
```
LindvedKrvang, line 79: Could be a tenary

src/app/rundown-view/components/rundown-header/rundown-header.component.ts
```diff
@@ -22,9 +22,9 @@ export class RundownHeaderComponent implements OnInit, OnDestroy, OnChanges {
...
 
   public showStyleVariant?: ShowStyleVariant
-  public currentLocalDate: Date = new Date()
-  private timeResolutionIntervalId: ReturnType<typeof setInterval>
-  private showStyleVariantSubscription?: SubscriptionLike
+  public currentLocalDate: number
+  private showStyleVariantSubscription?: Subscription
+  private rundownTimingContextSubscription?: Subscription
   public readonly IconButton = IconButton
   public readonly IconButtonSize = IconButtonSize
```
LindvedKrvang, line 29: Would be nice if we didn't mix public and private variables but instead kept the public variables together and the private variables together.

src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -0,0 +1,187 @@
...
+import { RundownTimingContext } from '../models/rundown-timing-context'
+import { Injectable } from '@angular/core'
+import { Segment } from '../models/segment'
+
+const HIGH_RESOLUTION_INTERVAL_DURATION_IN_MS: number = Math.floor(1000 / 30)
+const LOW_RESOLUTION_INTERVAL_DURATION_IN_MS: number = Math.floor(1000 / 4)
+
+@Injectable()
+export class RundownTimingContextStateService {
+  private timeResolutionTimerId?: ReturnType<typeof setTimeout>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: I was a bit torn between these two. Personally I like Node.Timeout since it is the type. But the type has nothing to do with Node, since this is a browser based project.

src/app/core/services/models/part-entity.service.ts
```diff
@@ -73,16 +73,29 @@ export class PartEntityService {
...
     return 0
   }
 
+  public getExpectedDuration(part: Part): number {
+    if (part.isUntimed) {
+      return 0
+    }
+    if (!part.expectedDuration) {
+      return 0
+    }
```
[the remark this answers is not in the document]
KvelaGorrrrnio: It was for the separation, but I have updated it to a single condition.

src/app/core/services/rundown-timing-context-state.service.ts
KvelaGorrrrnio: Updated to have a linear flow.

src/app/core/services/rundown-timing-context-state.service.ts
KvelaGorrrrnio: I have updated it to use object indexing instead of destructuring and have added the type `boolean` to the variables.

src/app/core/services/rundown-state.service.ts
```diff
@@ -75,16 +74,34 @@ export class RundownStateService implements OnDestroy {
...
     const activeRundown: Rundown = this.rundownEntityService.activate(rundownSubject.value)
     rundownSubject.next(activeRundown)
   }
 
+  private getRundownSubject(rundownId: string): BehaviorSubject<Rundown> | undefined {
+    const rundownSubject = this.rundownSubjects.get(rundownId)
+    if (!rundownSubject) {
+      return
+    }
+    const { wasRemoved } = this.removeSubjectIfHasNoObservers(rundownSubject)
```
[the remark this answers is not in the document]
KvelaGorrrrnio: To the first comment:
I have updated it to use object indexing instead of destructuring and have added the type `boolean` to the variables.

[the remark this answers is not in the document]
KvelaGorrrrnio: To the second comment:
Is there a way to observe unsubscribes on observables (without making a custom wrapper)?

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.html
KvelaGorrrrnio: The subpanel class is used for subpanels inside the `on-air-details-panel`.

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.html
KvelaGorrrrnio: We want to be able to do custom formatting. How would you achieve that without having either a lot of custom pipes or a format input?
This is the same princip as the build-in date pipe.

KvelaGorrrrnio: Yes.

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.ts
KvelaGorrrrnio: I have changed the variable type to `Tv2PieceType` and modified the associated lambda function make use of it.

src/app/rundown-view/components/rundown-header/rundown-header.component.ts
KvelaGorrrrnio: They are now ordered.

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.ts
KvelaGorrrrnio: Changed to a ternary.

src/app/core/services/rundown-state.service.ts
```diff
@@ -75,16 +74,34 @@ export class RundownStateService implements OnDestroy {
...
+  private getRundownSubject(rundownId: string): BehaviorSubject<Rundown> | undefined {
+    const rundownSubject = this.rundownSubjects.get(rundownId)
+    if (!rundownSubject) {
+      return
+    }
+    const wasRemoved: boolean = this.removeSubjectIfHasNoObservers(rundownSubject).wasRemoved
+    return wasRemoved ? undefined : rundownSubject
+  }
+
+  private removeSubjectIfHasNoObservers(rundownSubject: BehaviorSubject<Rundown>): { wasRemoved: boolean } {
```
andr9528, line 94: The name of this method sounds weird in my mind. Propably because the sentence it almost writes are missing a single word. Sounds like it is missing an 'It' between 'If' and 'Has'. I understand why you didn't add that, as the name is already quite long.

Would the name `removeObserverlessSubject` maybe be better?

src/app/core/services/rundown-timing-context-state.service.ts
KvelaGorrrrnio: The `onRundownChanged` method handles the "static" values, that aren't affected by time alone, where `onTimeTick` updates the values that are affected by the passing of time.

To make it easier to get an overview of it, I have split up the functionality on `onTimeTick` into a method that handles updating the time related properties of a rundown timing context and a method for adjusting the timer interval.

src/app/core/services/rundown-timing-context-state.service.ts
KvelaGorrrrnio: See https://github.com/tv2/ng-sofie/pull/67#discussion_r1401946310.

src/app/core/services/rundown-state.service.ts
andr9528: `RundownTimingContextStateService.removeSubjectIfHasNoObserversOrSubscriptions` is similarly named, thus might benefit form similar renaming.

KvelaGorrrrnio: `removeObserverlessSubject` sounds like it expects a subject that has no observers and will always remove it. The `If` is there to indicate the conditional effect. I have updated it with an `It` inserted.

src/app/core/services/rundown-timing-context-state.service.ts
KvelaGorrrrnio: Is removed.

src/app/core/services/rundown-timing-context-state.service.ts
KvelaGorrrrnio: Since this is using an interval as the timer, yes. But I have updated the time tick scheduling to use timeouts instead. It becomes a bit more imprecise, but removes the need for the state variable `isHighResolutionTimer`.

src/app/core/services/rundown-timing-context-state.service.ts
KvelaGorrrrnio: See https://github.com/tv2/ng-sofie/pull/67#discussion_r1401972906.

andr9528 (approved):

src/app/core/services/rundown-timing-context-state.service.ts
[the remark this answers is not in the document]
LindvedKrvang: But isn't it technically still the type?

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.html
LindvedKrvang: I don't have any suggestions, so I guess we will keep it as is :)

LindvedKrvang: Double okay ;)

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #68, opened 2023-11-21
SOF-1681: Remote segment countdown

KvelaGorrrrnio:
In the segment header in the rundown view, a countdown is shown for future segments containing a remote piece. Clicking the countdown will toggle it to show a timestamp for when the segment is expected to start.

Skærmbillede 2023-11-21 kl  08 59 14


src/app/core/services/rundown-timing.service.ts
```diff
@@ -5,6 +5,11 @@ import { PartEntityService } from './models/part-entity.service'
 import { Part } from '../models/part'
 import { Injectable } from '@angular/core'
 
+interface AccumulatedResult<Result, Accumulator> {
+  result: Result
+  accumulator: Accumulator
+}
```
LindvedKrvang, line 11: This is a quite confusing interface. I have been looking at the places you use it and I can't tell what it is the values for `result` and `accumulator` is supposed to mean.

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -41,20 +40,36 @@ export class RundownComponent implements OnInit, OnDestroy {
...
     } else {
-      this.remainingDurationInMsForOnAirPart = rundownTimingContext.playedDurationInMsForOnAirPart - this.partEntityService.getExpectedDuration(onAirPart)
+      this.remainingDurationInMsForOnAirPart = this.partEntityService.getExpectedDuration(onAirPart) - rundownTimingContext.playedDurationInMsForOnAirPart
+    }
+
+    if (!onAirPart) {
+      this.startOffsetsInMsFromPlayheadForSegments = {}
+    } else {
+      this.startOffsetsInMsFromPlayheadForSegments = this.getStartOffsetsInMsFromPlayheadForSegments(rundownTimingContext)
     }
```
LindvedKrvang, line 57: Could be a ternary

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -41,20 +40,36 @@ export class RundownComponent implements OnInit, OnDestroy {
...
 
     const onAirPart: Part | undefined = this.rundown.segments.find(segment => segment.isOnAir)?.parts.find(part => part.isOnAir)
 
-    if (!onAirPart || !this.doesPartContainVideoClipOrVoiceOver(onAirPart)) {
+    if (!onAirPart) {
       this.remainingDurationInMsForOnAirPart = undefined
     } else {
-      this.remainingDurationInMsForOnAirPart = rundownTimingContext.playedDurationInMsForOnAirPart - this.partEntityService.getExpectedDuration(onAirPart)
+      this.remainingDurationInMsForOnAirPart = this.partEntityService.getExpectedDuration(onAirPart) - rundownTimingContext.playedDurationInMsForOnAirPart
+    }
```
LindvedKrvang, line 51: Could be a ternary.

Also, you make the exact same check twice right after each other.
Since you are not using ternarys you should combine the two if-statements.

src/app/core/services/rundown-timing.service.ts
andr9528: Only because i have seen the tests for `getStartOffsetsInMsFromNextCursorForSegments` do i have an Okay understanding of what this interface does.

That said, without the tests showing an example of how it is used in pratice, I too would have been very confused.

Maybe renaming `accumulator` to something along the lines of `accumulatedMsOffset` could help. What ever renaming is applied should also be appled similarly in `getStartOffsetsInMsFromNextCursorForSegments`.

KvelaGorrrrnio: I have specialized the interface to the following snippet and have tried to simplify the method that uses it.
```typescript
interface AccumulatedStartOffsetInMsForSegmentsData {
  startOffsetsInMsFromNextCursorForSegments: Record<string, number>
  accumulatedStartOffsetInMsFromNextCursor: number
}
```

src/app/rundown-view/components/rundown/rundown.component.ts
KvelaGorrrrnio: The condition is removed.

src/app/rundown-view/components/rundown/rundown.component.ts
KvelaGorrrrnio: Is changed to a ternary.

src/app/core/services/rundown-timing.service.ts
andr9528: Hmm, in my mind the longer names / extraction to a seperate method didn't help make it more clear what happens.

I have an idea for a possibility refactor, that i'd like to spend a bit of time on, on monday, if it is alright with you. I will be branching from your branch, and then creaing a small seperate PR, so you can review my idea.

Or maybe you just want to talk me throught the idea with you. Let me know what you prefer.

LindvedKrvang (approved):

src/app/core/services/rundown-timing.service.ts
LindvedKrvang: The more specific names helped :)

RasmusAlbrektsen (approved):

PR #70, opened 2023-11-23
feat/ Display VO Pieces

RasmusAlbrektsen:
This PR lets the frontend depict pieces differently when they have voice over. For now this is introduced for `REPLAY` and `VIDEO_CLIP` piece types. 
This is dependent on the blueprint changes in: https://github.com/tv2/sofie-blueprints-inews/pull/227


src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -84,6 +85,11 @@ export class OffsetablePieceComponent {
 
   @HostBinding('class')
   public get getPieceTypeModifierClass(): string {
-    return (this.piece as Tv2Piece).metadata.type.toLowerCase().replace(/_/g, '-')
+    const piece: Tv2Piece = this.piece as Tv2Piece
+    const pieceTypeModifierClass: string = piece.metadata.type.toLowerCase().replace(/_/g, '-')
+    if (piece.metadata.audioMode && piece.metadata.audioMode === Tv2AudioMode.VOICE_OVER) {
+      return pieceTypeModifierClass + '-voice-over'
```
LindvedKrvang, line 91: Magic string. Would be nice to atleast have it as a local variable.

LindvedKrvang (approved):

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.scss
```diff
@@ -54,6 +54,19 @@
...
+    background-color: #8d1010;
+  }
+
+  &.video-clip-voice-over {
+    background: linear-gradient(to bottom,  #1769ff 0%,#1769ff 50%, #005919 50%,#005919 100%);
+  }
+
+  &.replay-voice-over {
+    background: linear-gradient(to bottom,  #8d1010 0%,#8d1010 50%, #005919 50%,#005919 100%);
+  }
```
KvelaGorrrrnio, line 68: Instead of using two classes here, we could use two classes instead. That would also make it possible to make a default voice over styling, e.g.:
```scss
  &.voice-over {
    /* default voice over styling */
  }

  &.replay {
    /* default replay styling */

    &.voice-over {
      /* replay voice over styling */
    }
  }
  
    &.server {
    /* default server styling */

    &.voice-over {
      /* replay voice over styling */
    }
  }
```

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -84,6 +85,11 @@ export class OffsetablePieceComponent {
...
+    }
+    return pieceTypeModifierClass
```
KvelaGorrrrnio, line 93: If we use two classes instead of one, we can also simplify this a bit to something like:
```typescript
return [
  piece.metadata.type.toLowerCase().replaceAll('_', '-'),
  (piece.metadata.audioMode ?? Tv2AudioMode.FULL).toLowerCase().replace('_', '-'),
].join(' ')

andr9528: Just rember to use `.replaceAll` in all instances where replacing is done, when using strings as patterns, or continue to use a regex as has already been used.

Personally i'd want to use `.replaceAll` in these very simple replace cases.

If more places than this are converting a Screaming Snake case string to a css class name, then maybe a helper method could be good?

RasmusAlbrektsen: Until we change to es2021 then replaceAll is not an option. It will not be used for now.

RasmusAlbrektsen: Implemented!

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
RasmusAlbrektsen: This no longer exists!

andr9528 (approved):

KvelaGorrrrnio (approved):

PR #79, opened 2023-12-08
SOF-1677 Standardize Yarn Scripts

andr9528:
Copied `clean` and `reset` script from Sofie Server repository.

~~Looking at the raw copied scripts, I removed the `*.log` from the reset, as the reset didn't appear to do anything if it that was there, and it failed to find any matching files. Has the `reset` been tested in Sofie Server?~~
Adding `--glob` to the rimraf command, fixes the above statement.


KvelaGorrrrnio (changes requested):
`*.log` removes files like `yarn-error.log`, which are not there by default, but if the developer makes a mistake or the build system encounters an error. So it will still be fine to keep :)

Currently the only scripts that adhere to the agreed standards are `build`, `lint`, `lint-fix`, `clean` and `reset`. The rest does not :)

andr9528:
The `yarn reset` is split into two `rimraf` commands, as the `*.log` segment, makes the rimraf command only target the specific file type in the targeted folders. From my understanding, we want to nuke the folders AND all files with the `*.log` extension.

package.json
KvelaGorrrrnio, line 1: It would be nice to have the `test-watch` and `build-watch` scripts available as well:
```json
{
  "scripts": {
    "build-watch": "ng build --watch --configuration production",
    "test-watch": "ng test",
  }
}
```

KvelaGorrrrnio (commented):
Looks good :+1:

KvelaGorrrrnio (approved):

LindvedKrvang (approved):

PR #82, opened 2023-12-11
Sof 1700/bind action triggers to keyboard

LindvedKrvang:
Keybindings are now generated from ActionTriggers.
This PR includes:
- Listening for ActionTrigger events.
- Generating Keyboard shortcuts bindings when either an Action-, ActionTrigger- or RundownEvent is being emitted.


src/app/shared/services/http-action-trigger.service.ts
```diff
@@ -0,0 +1,23 @@
...
+  constructor(
+    private readonly http: HttpClient,
+    private readonly httpErrorService: HttpErrorService
+  ) {
+    super()
+  }
+
+  public getActionTriggers(): Observable<ActionTrigger[]> {
+    return this.http.get<ActionTrigger[]>(ACTION_TRIGGER_URL).pipe(catchError(error => this.httpErrorService.catchError(error)))
+  }
```
andr9528, line 22: We say that what we get is an array of ActionTrigger, but the result is not validated. to ensure that the result is actually of this type, like we do for other entities/actions.

RasmusAlbrektsen (approved):
Looks good!

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -0,0 +1,199 @@
...
+import { RundownStateService } from '../../core/services/rundown-state.service'
+import { Rundown } from '../../core/models/rundown'
+import { Tv2ActionParser } from '../../shared/abstractions/tv2-action-parser.service'
+import { KeyBindingFactory } from '../factories/key-binding.factory'
+import { Tv2Action, Tv2ActionContentType } from '../../shared/models/tv2-action'
+import { Logger } from '../../core/abstractions/logger.service'
+import { StyledKeyBinding } from '../../keyboard/value-objects/styled-key-binding'
+import { ActionService } from '../../shared/abstractions/action.service'
+
+// TODO: We need ONE central place to define colours
```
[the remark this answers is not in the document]
LindvedKrvang: That enum is currently used for "system" colours like the pulsating dot. Not sure if we should mix that with the colours for our Parts and Pieces.
But we definately should make one place to define the colours.
I will write a story for it.

src/app/shared/services/http-action-trigger.service.ts
LindvedKrvang: The data is now being validated.

andr9528 (approved):

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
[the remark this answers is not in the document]
RasmusAlbrektsen: Ah yes, my mistake. But i agree, it would be nice with a single source for the colors!

src/app/shared/services/http-action-trigger.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    private readonly httpErrorService: HttpErrorService,
+    private readonly actionTriggerParser: ActionTriggerParser
+  ) {
+    super()
+  }
+
+  public getActionTriggers(): Observable<ActionTrigger[]> {
+    return this.http.get<ActionTrigger[]>(ACTION_TRIGGER_URL).pipe(
+      catchError(error => this.httpErrorService.catchError(error)),
+      map(actionTriggers => this.actionTriggerParser.parseActionTriggers(actionTriggers))
```
andr9528, line 25: With the now merged SOF-1666, this needs to be updated to support the changes in the backend. The below code snippet should do it. The containing file should also be moved to the new `./htpp/` folder

```ts
public getActionTriggers(): Observable<ActionTrigger[]> {
    return this.http.get<HttpResponse<ActionTrigger[]>>(ACTION_TRIGGER_URL).pipe(
      catchError(error => this.httpErrorService.catchError(error)),
      map(response => this.actionTriggerParser.parseActionTriggers(response .data))
    )
  }
}
```

andr9528 (changes requested):
With the now merged SOF-1666 a few changes are needed to support the backend changes. Have made a comment at the method that need changes.

src/app/shared/services/http-action-trigger.service.ts
LindvedKrvang: There is a task created for this already https://tv2cms.atlassian.net/browse/SOF-1756

LindvedKrvang:
> With the now merged SOF-1666 a few changes are needed to support the backend changes. Have made a comment at the method that need changes.

There is a task created for this https://tv2cms.atlassian.net/browse/SOF-1756 already
Since it is both a frontend and backend change it will be in a separate PR

andr9528 (approved):

PR #83, opened 2023-12-18
SOF-1699 add empty router for setting pages with 2 new routes

vivanov90:
I create a draft pool request so that the team can review the structure of the settings router and decide whether it is good. 
I use 2 different ways of loading the components, because I expect one to be quite small (only one component), and the other to be significantly larger with its own service, etc.
```
{ path: 'action-triggers', loadChildren: () => import('./action-triggers/action-triggers.module').then(module => module.ActionTriggersModule) },
      { path: 'clear-cache', component: ClearCacheComponent },
```

### DIFF
--- a/src/app/settings/components/clear-cache/clear-cache.component.html
+++ b/src/app/settings/components/clear-cache/clear-cache.component.html
@@ -1,0 +1,1 @@
+Clear cache
--- a/src/app/settings/components/settings/settings-menu/settings-menu.component.html
+++ b/src/app/settings/components/settings/settings-menu/settings-menu.component.html
@@ -1,0 +1,8 @@
+<mat-toolbar class="c-side-bar">
+  <mat-nav-list class="c-side-bar__nav-item">
+    <a routerLinkActive="active-tab" [routerLink]="['/settings/action-triggers']" i18n>settings.action-triggers.label</a>
+  </mat-nav-list>
+  <mat-nav-list class="c-side-bar__nav-item">
+    <a routerLinkActive="active-tab" [routerLink]="['/settings/clear-cache']" i18n>settings.clear-cache.label</a>
+  </mat-nav-list>
+</mat-toolbar>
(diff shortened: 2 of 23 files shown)


src/app/settings/components/clear-cache/clear-cache.component.html
```diff
@@ -0,0 +1 @@
+Clear cache
```
LindvedKrvang, line 1: Is this component in anticipation of SOF-1754?

src/app/settings/components/settings/settings-menu/settings-menu.component.html
```diff
@@ -0,0 +1,8 @@
+<mat-toolbar class="c-side-bar">
```
LindvedKrvang, line 1: We have been talking about (but not done much about it) that we would like to "abstract away" the dependency of component libraries, so that if we decide to change them, we wouldn't have to change our entire codebase.

We are currently not 100% set in stone that we want to use Angular Material.
So it would be nice if we could make our own `ToolbarComponent` etc. These components would for now just use Angular Material. So the `SofieToolbarComponent` would just use `<mat-toolbar>...</mat-toolbar>` in its markup.

The benenfit would be that our entire frontend codebase wouldn't be dependent on a specific component library, but just our own components. So if we decide to change component library, we would only need to update those components and not our entire frontend.

I'm curious to hear your input on this suggestion :)

src/app/settings/components/settings/settings.component.html
```diff
@@ -0,0 +1,6 @@
+<sofie-header></sofie-header>
+
+<mat-drawer-container class="c-settings">
```
LindvedKrvang, line 3: Same comment as for the `mat-toolbar` :)

LindvedKrvang (commented):
It looks good. I'm a fan of the router approach.

I didn't approve since it is a draft PR and draft PRs useally indicates that there will be more changes in the future :)

src/app/settings/components/clear-cache/clear-cache.component.html
vivanov90: Yes

src/app/settings/components/settings/settings.component.html
vivanov90: Will try to remove it for now.

PR #87, opened 2024-01-03
WIP - SOF-1755/parsers to validators

k-primeholding-dim:
Branch is based on ongoing work in  SOF-1699/create-action-triggers-page, thus result will be directed back there.
update: since SOF-1699 will span even further in diverging, to finish this task my new target branch has to be develop.


LindvedKrvang (approved):

KvelaGorrrrnio (commented):
A validator, as I understand it, is a construct that checks if a given set of data conforms to a specific structure or upholds some predefined rules. I don't think that validator is a fitting description for these classes. The data that goes in to the our current validators are not guaranteed to have the same structure when returned.

k-primeholding-dim:
Will need to wait til [SOF-1699](https://github.com/tv2/ng-sofie/pull/86) is merged, then perhaps recreate the code changes.

PR #88, opened 2024-01-11
SOF-1720/display minishelf

k-primeholding-dim:
In SOF-1720 a new component is added called MiniShelf. This component should render in rundown view, based on logic to be visible or not, and that logic uses data on the segment it represent.

Expects a Segment and Tv2VideoClipAction as inputs.

A couple of new services are introduced too:
ConfigurationService -> HttpConfigurationService
MediaStateService -> MediaService -> HttpMediaService
MiniShelf reacts on changes in segment and media data.


LindvedKrvang (commented):
Just a quick comment:
Mini Shelf are two words, so we should uppercase the "s" of Shelf when concatenating the two words i.e. `miniShelf`.

This means that:
- Filenames would be `mini-shelf.component.ts`
- CSS classes would be `c-mini-shelf`
- Component name would be `MiniShelfComponent`
- etc

src/app/rundown-view/components/minishelf/mini-shelf.component.html
```diff
@@ -0,0 +1,27 @@
+<div class="c-mini-shelf">
+  <div
+    class="c-mini-shelf__part"
+    *ngFor="let part of segment?.parts;"
```
KvelaGorrrrnio, line 4: From the class definition, the segment input variable is required, so the optional chaining her should be removed.

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,26 @@
...
+  constructor(logger: Logger) {
+    this.logger = logger.tag('MinishelfComponent')
+  }
+
+  public ngOnChanges(changes: SimpleChanges): void {
+    this.logger.info('ngOnChanges for MinishelfComponent', changes)
+  }
+  public ngOnDestroy(): void {
+    this.logger.debug('ngOnDestroy for MinishelfComponent')
+  }
```
KvelaGorrrrnio, line 25: If these are not planned to have functionality in them, it would be nice if the class didn't implement these interfaces.

src/app/shared/services/media-data.ts
```diff
@@ -0,0 +1,4 @@
+export interface MediaData {
```
LindvedKrvang, line 1: Just call it `Media`, no need to include `Data`.

src/app/shared/services/studio-configuration.ts
```diff
@@ -0,0 +1,21 @@
...
+  constructor(url: string) {
+    this.settings = new mediaPreviewUrl(url)
+  }
+}
+export class StudioConfiguration implements Configuration {
+  public data: Settings
+  constructor(url: string) {
+    this.data = new Settings(url)
+  }
+}
```
LindvedKrvang, line 21: These don't utilize any functionality of being a class, so they might as well just be interfaces.

LindvedKrvang (commented):
I know this is still a work in progress. If I have commented on something that isn't done, then just disregard that comment :)

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  public ngOnDestroy(): void {
+    this.mediaDataSubscription?.unsubscribe()
+    this.configurationServiceSubscription?.unsubscribe()
+  }
+
+  protected get mediaPreviewUrl(): string {
+    const url: string = `${this.configurationMediaPreviewUrl}/media/thumbnail/${this.segment.metadata?.miniShelfVideoClipFile}`
+    return this.configurationMediaPreviewUrl ? url : 'assets/sofie-logo.svg'
```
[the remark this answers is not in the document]
k-primeholding-dim: In such case then I'll need an asset image to use instead. Also in case thumbnail image is not reachable (fails to load or is non-existent ) and in:
`  protected handleMissingImage(event: Event): void {
    ;(event.target as HTMLImageElement).src = 'assets/sofie-logo.svg'
  }`

src/app/rundown-view/components/minishelf/mini-shelf.component.html
```diff
@@ -0,0 +1,19 @@
...
+  class="c-mini-shelf"
+  title="{{ segment.name }}"
+  (click)="emitActionEvent()"
+>
+    <img
+        class="c-mini-shelf__thumbnail"
+        src="{{ mediaPreviewUrl }}"
+        alt="{{ segment.name? segment.name : 'Title is not available.' }}"
+        (error)=handleMissingImage($event)
+    />
```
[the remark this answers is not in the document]
k-primeholding-dim: TODO reminder added, or should I create a sub-task?

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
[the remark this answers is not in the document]
LindvedKrvang: Yes, we should ask Jacob what he wants in that case

src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -155,4 +163,38 @@ export class ZodEntityParser implements EntityParser {
...
+    return <Tv2Action>this.tv2ActionParser.parse(tv2Action)
+  }
+
+  private readonly mediaDataParser = zod.object({
+    id: zod.string(),
+    duration: zod.number(),
+  })
+
+  public parseMedia(media: unknown): Media {
+    return <Media>this.mediaDataParser.parse(media)
```
LindvedKrvang, line 198: Should just be `mediaParser` and not `mediaDataParser`

src/app/rundown-view/components/minishelf/mini-shelf.component.html
```diff
@@ -0,0 +1,20 @@
...
+    <div class="c-mini-shelf__note">
+        <span class="c-mini-shelf__title"
+        >{{ getSanitizedTitle() ? getSanitizedTitle() : 'Title is not available.'  | slice:0:30 | uppercase }}{{ segment.name.length > 30 ? '...' : '' }}</span>
+        <span class="c-mini-shelf__duration">{{ mediaDuration === 0 || isNaN(mediaDuration) ? '-- : --' : mediaDuration | timer : 'HH?:mm:ss' }}</span>
```
LindvedKrvang, line 16: The logic here seems like something that could belong inside the `timer` pipe. So if the pipe receives an invalid input, it just returns `--:--`.

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,97 @@
...
+  selector: 'sofie-mini-shelf',
+  styleUrls: ['./mini-shelf.component.scss'],
+  templateUrl: './mini-shelf.component.html',
+})
+export class MiniShelfComponent implements OnInit, OnDestroy, OnChanges {
+  @Input() public segment: Segment
+  @Input() public videoClipAction: Tv2VideoClipAction | undefined
+
+  private readonly defaultAssetForThumbnail: string = 'assets/sofie-logo.svg'
+  protected mediaDuration: number = 0
```
LindvedKrvang, line 21: You are fetching the entire Media anyway so you might as well just store the entire Media instead of just the duration.
Then we utilize that we are sending objects back and forth

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,97 @@
...
+  private configurationMediaPreviewUrl: string
+  private configurationServiceSubscription: Subscription
+  private executeActionSubscription: Subscription
+  private serverPostrollDuration: number = 0
```
LindvedKrvang, line 25: The `configurationMediaPreviewUrl` and `serverPostrollDuration` are both attributes on the same interface `StudioConfiguration`.
Instead of saving all the attributes individually you should just store the `StudioConfiguration` which you are fetching anyway.

Utilize the objects :D

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,97 @@
...
+    private readonly actionService: ActionService,
+    private readonly configurationService: ConfigurationService,
+    private readonly mediaStateService: MediaStateService
+  ) {}
+
+  public ngOnInit(): void {
+    this.configurationServiceSubscription = this.configurationService.getStudioConfiguration().subscribe((configuration: StudioConfiguration) => {
+      this.configurationMediaPreviewUrl = configuration.data.settings.mediaPreviewUrl
+      this.serverPostrollDuration = configuration.data.blueprintConfiguration.ServerPostrollDuration
+    })
```
LindvedKrvang, line 37: Piggybacking on the comment above about storing the `StuioConfiguration`, then this will just become:
```
this.configurationServiceSubscription = this.configurationService.getStudioConfiguration().subscribe((configuration: StudioConfiguration) =>{
  this.configuration = configuration
})
```

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,97 @@
...
+      if (!media || !media.duration) {
+        return
+      }
+      if (media.duration < this.serverPostrollDuration) {
+        this.mediaDuration = this.NaN
+      } else {
+        this.mediaDuration = media.duration - this.serverPostrollDuration
+      }
+    }
+  }
```
LindvedKrvang, line 53: If you invert the first if-statement you can reduce the nesting to one:
```
private async updateMediaDataDuration(): Promise<void> {
  if (!this.segment.metada?.miniShelfVideoClipFile) {
    return
  }
  ...
}
```

But a lot of the logic in this method seems redundant. If you just store the entire Media object instead of just the duration this would just become:
```
private async fetchMediaForMiniShelf(): Promise<void> {
  if (!...) {  // check if there isn't a miniShelVideoClipFile
    return
  }
  this.media = await this.mediateStateService.getMedia(...)
}
```
And then in your html file your can subtract the ServerPostRollDuration when you are displaying it.

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,97 @@
...
+      .replace(/&quot;/g, '"') // remove &quot;
+      .replace(/&apos;/g, "'") // remove &apos;
+      .trim() // and finally trim;
+  }
+
+  protected executeAction(): void {
+    if (!this.videoClipAction) {
+      return
+    }
+    this.executeActionSubscription = this.actionService.executeAction(this.videoClipAction.id, this.segment.rundownId).subscribe()
```
LindvedKrvang, line 88: You don't need to store the subscribtion for this.
The angular framework unsubscribes automatically from this :) 

So this can be written as:
```
public executeAction(): void {
  this.actionService.executeAction(this.videoClipAction.id, this.segment.rundownId).subscribe()
}
```

Also, why is it protected? You seems to be using `procteted` and `public` arbitrarily on the methods in here?

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  protected handleMissingImage(event: Event): void {
+    ;(event.target as HTMLImageElement).src = this.defaultAssetForThumbnail
+  }
+
+  protected readonly NaN = NaN
+  protected readonly isNaN = isNaN
```
LindvedKrvang, line 96: This seems to be instance variables. All instance variables should be a the top of the class so we have a single place for all instance variables and then we don't have to "hunt" for them in files we are not familliar with.

Also, the `NaN = NaN` seems redundant and I would prefer it got the deleted.
The same for `isNaN = isNaN`. You do use this one in the html, but if you move the check to the `timer` then you don't need this here.

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -47,6 +62,20 @@ export class RundownComponent implements OnInit, OnDestroy {
     this.startOffsetsInMsFromPlayheadForSegments = this.getStartOffsetsInMsFromPlayheadForSegments(rundownTimingContext)
   }
 
+  private onActionsChanged(actions: Action[]): void {
+    this.videoClipActions = actions.filter((action): action is Tv2VideoClipAction => {
+      return this.entityParser.parseTv2Action(action) && (<Tv2Action>action).metadata?.contentType === Tv2ActionContentType.VIDEO_CLIP
```
LindvedKrvang, line 67: The entityParser are supposed to be used when we receive the data from the backend. They are not supposed to be used in the components. The data should already have been through the parser before it ever sees a component.

What do you need it for here?

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -64,9 +93,31 @@ export class RundownComponent implements OnInit, OnDestroy {
...
+    const videoClipFile: string | undefined = segment.metadata?.miniShelfVideoClipFile
+    if (videoClipFile !== undefined) {
+      const action: Tv2VideoClipAction | undefined = this.videoClipActions.find(action => {
+        return action.metadata?.fileName === videoClipFile
+      })
+      if (action !== undefined) {
+        return { ...actionMap, [segment.id]: action }
+      }
+    }
+    return actionMap
```
LindvedKrvang, line 121: By using "defensive" programming i.e. reversing the if-statement and returning early, you can reduce the layers of nesting.
We should aim for only have one layer of nesting pr method.
This would then become:
```
private miniShelfSegmentsReducer(...): ... {
  const videoClipFile: string | undefined = ...
  if (!videoClipFile) {
    return actionMap
  }
  const action: Tv2VideoClipAction | undefined = ...
  return action ? { ...actionMap, [segment.id]: action } : actionMap
}
```

src/app/shared/services/http/http-configuration-service.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(
+    private readonly http: HttpClient,
+    private readonly httpErrorService: HttpErrorService,
+    private readonly entityParser: EntityParser
+  ) {}
+
+  public getStudioConfiguration(): Observable<StudioConfiguration> {
+    return this.http.get<HttpResponse<StudioConfiguration>>(`${environment.apiBaseUrl}/configurations/studio`).pipe(
+      catchError(error => this.httpErrorService.catchError(error)),
+      map(response => this.entityParser.parseStudioConfiguration(response))
```
LindvedKrvang, line 22: You are parsing the entire `Response` object to the parser. You should only parse the `data` part of the Response, so it becomes `map(response => this.entityParser.parseStudioConfiguration(response.data))`.

Just like you did for Media :)

src/app/shared/services/media-state.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public async getMedia(id: string): Promise<Media | undefined> {
+    let media: Media | undefined = undefined
+    try {
+      media = await this.fetchMedia(id)
+    } catch (error) {
+      this.logger.error(`Failed to fetch media with id: ${id}, reason: ${error}`)
+    }
+    return media
+  }
```
LindvedKrvang, line 26: The HttpMediaService is already catching the error if we fail to find one, you don't need to do it again here
image

But I wouldn't spend much time on the `MediaStateService` for now. SOF-1652 is going to set it up with listening for Media events etc and I believe @RasmusAlbrektsen is going to pick up that task

src/app/shared/services/media-state.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public fetchMedia(id: string): Promise<Media> {
```
LindvedKrvang, line 28: Why is this public? It isn't part of the interface

src/app/shared/services/studio-configuration.ts
```diff
@@ -0,0 +1,14 @@
...
+interface Settings {
+  mediaPreviewUrl: string
+}
+
+export interface StudioConfiguration {
+  data: {
+    settings: Settings
+    blueprintConfiguration: BlueprintConfiguration
+  }
+}
```
LindvedKrvang, line 14: The `data` is not part of the `StudioConfiguration`. It's part of the `Response` object that we receive from the backend.
This interface should just be:
```
export interface StudioConfiguration {
  settings: Settings
  blueprintConfiguration: BlueprintConfiguration
}
```

src/app/shared/services/studio-configuration.ts
```diff
@@ -0,0 +1,14 @@
+interface BlueprintConfiguration {
+  ServerPostrollDuration: number
+}
+
+interface Settings {
```
LindvedKrvang, line 5: Would be nice if these where called `StudioSettings` so we knew that they were the settings for Studio. Same thing applies for `BlueprintConfiguration` :)

src/app/rundown-view/components/minishelf/mini-shelf.component.scss
```diff
@@ -0,0 +1,42 @@
...
+    height: 18px;
+    background: #1768ff;
+    padding: 2px;
+    text-shadow: black 0 0 2px;
+    font-weight: 450;
+    font-size: 12px;
+  }
+
+  &__title {
+    color: white;
```
LindvedKrvang, line 35: Viktor has introduced variables for colors that we should be using. Not sure if they have reached develop yet.

LindvedKrvang (commented):
Looks good. A few things here and there :)

src/app/rundown-view/components/minishelf/mini-shelf.component.scss
k-primeholding-dim: Not yet. Should we bump a sub-task as to re-align this once Victor's task is merged?

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
k-primeholding-dim: I prefer to have the calculus in a method, else html will do calculus and pipe transformations based on value.

src/app/rundown-view/components/minishelf/mini-shelf.component.html
k-primeholding-dim: Indeed, but there we'll find tests that contradict with this, with leads me in direction that the timer pipe contract is stated already on this subject, i.e. if duration is negative pipe will display '-', but if it is negative 500ms will return '01' and so on.

src/app/rundown-view/components/minishelf/mini-shelf.component.scss
LindvedKrvang: We have SOF-1758. Maybe write a comment on that one that this is a place that needs to be updated?

src/app/core/parsers/zod-entity-parser.service.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+          ServerPostrollDuration: 123,
+        },
+      }
+      const result = (): StudioConfiguration => testee.parseStudioConfiguration(expectedStudioConfiguration)
+      expect(result).toThrow()
+    })
+  })
+
+  describe(ZodEntityParser.prototype.parseTv2Action.name, () => {
+    it('parses a Tv2Adtion ', () => {
```
KvelaGorrrrnio, line 66: Tv2Action is misspelled.

And same note as the previous comment, that this is organisation specific logic.

src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -155,4 +163,36 @@ export class ZodEntityParser implements EntityParser {
   public parseShowStyleVariant(showStyleVariant: unknown): ShowStyleVariant {
     return this.showStyleVariantParser.parse(showStyleVariant)
   }
+
+  private readonly studioConfigurationParser = zod.object({
+    settings: zod.object({
+      mediaPreviewUrl: zod.string().startsWith('http://', 'Media preview url must start with http://').and(zod.string().min(9, 'Media preview url must more than 9 characters long')),
+    }),
+    blueprintConfiguration: zod.object({
+      ServerPostrollDuration: zod.number().min(0, 'Server postroll duration must be 0 or more'),
```
KvelaGorrrrnio, line 172: It would be nice if the sentences in messages we provide are ended with a period.

src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -155,4 +163,36 @@ export class ZodEntityParser implements EntityParser {
...
+    return <Tv2Action>this.tv2ActionParser.parse(tv2Action)
+  }
+
+  private readonly mediaDataParser = zod.object({
+    id: zod.string(),
+    duration: zod.number(),
+  })
+
+  public parseMedia(media: unknown): Media {
+    return <Media>this.mediaDataParser.parse(media)
```
KvelaGorrrrnio, line 196: There should not be a type assertion here.

src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -155,4 +163,36 @@ export class ZodEntityParser implements EntityParser {
...
+  }
+
+  private readonly tv2ActionParser = zod.object({
+    metadata: zod.object({
+      contentType: zod.nativeEnum(Tv2ActionContentType),
+    }),
+  })
+
+  public parseTv2Action(tv2Action: unknown): Tv2Action {
+    return <Tv2Action>this.tv2ActionParser.parse(tv2Action)
```
KvelaGorrrrnio, line 187: There should not be a type assertion here.

src/app/rundown-view/components/minishelf/mini-shelf.component.html
KvelaGorrrrnio: As I see it, the issue lies somewhere else than here. First of all, why do we receive a `NaN`? And why won't we put 0 milliseconds into the timer pipe?

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,95 @@
...
+    private readonly actionService: ActionService,
+    private readonly configurationService: ConfigurationService,
+    private readonly mediaStateService: MediaStateService
+  ) {}
+
+  public ngOnInit(): void {
+    this.configurationServiceSubscription = this.configurationService.getStudioConfiguration().subscribe((studioConfiguration: StudioConfiguration) => {
+      this.studioConfiguration = studioConfiguration
+    })
+    void this.updateMediaDataDuration()
```
KvelaGorrrrnio, line 35: `void` is in most cases a bad practice that hides an issue. In this case, if the promise returned by `updateMediaDataDuration` is rejected, then we won't catch it (besides from the console log). If this is a node process, then the unhandled rejected promise can crash the application.

I suggest to add a `.catch` that at least logs the error.

src/app/core/models/segmentMetadata.ts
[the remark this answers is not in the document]
thedgm: ah indeed - shish-kebab it is :) thanks!

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
k-primeholding-dim: Idea was to represent facts - i.e. if actual duration can not be determined - display --:-- which differs than 00:00 since latter is not true.

src/app/core/parsers/zod-entity-parser.service.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      }
+      const parseTv2Action: Tv2Action = testee.parseTv2Action(expectedTv2Action)
+      expect(parseTv2Action).toEqual(expectedTv2Action)
+    })
+  })
+
+  describe(ZodEntityParser.prototype.parseMedia.name, () => {
+    it('parses a Media data ', () => {
+      const testee: ZodEntityParser = new ZodEntityParser()
+      const expectedMedia: Media = <Media>{
```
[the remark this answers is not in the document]
LindvedKrvang: It should be noted however, that we should try to limit the use of `as`

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -18,41 +19,51 @@ export class MiniShelfComponent implements OnInit, OnDestroy, OnChanges {
...
+
+    const media: Media = await this.mediaStateService.getMedia(this.segment.metadata?.miniShelfVideoClipFile)
+    if (!media) return
+    this.media = media
+  }
+
+  private calculateMediaDuration(): void {
+    if (!this.segment.metadata?.miniShelfVideoClipFile) return
+    if (!this.studioConfiguration) return
+    if (this.media.duration < this.studioConfiguration.blueprintConfiguration.ServerPostrollDuration) return
```
LindvedKrvang, line 57: This is in the end going to come down to personal preference, but you could put these conditionals in one if-statement:
```
if (!this.segment.metadata?.miniShelfVideoClipFile
    || !this.studioConfiguration
    || this.media.duration < this.studioConfiguration.blueprintConfiguration.ServerPostrollDuration) {
  return
}
```
I do agree however, that the above conditional might become "too much" to read. In cases like those an option could be to extract it to its own method:

```
private calcuateMediaDuration(): void {
  if (!this.canMediaBeCalculated()) {
    return
  }
  // Calculate operation here
}

private canMediaBeCalculated(): booelan {
  return this.segment.metadata?.miniShelfVideoClipFile
    && this.studioConfiguration
    && this.media.duration >= this.studioConfiguration.blueprintConfiguration.ServerPostrollDuration
}
```
Again, the above comes down to personal preference :) 

Another note: You could get rid of the check if the media.duration is smaller that the ServerPostrollDuration by using `Math.max()`:
```
this.mediaCaluclateDuration = Math.max(this.media.duration - this.studioConfiguration.blueprintConfiguration.ServerPostrollDuration, 0)
```
The above also comes down to personal preference so take it or leave it :)

src/app/shared/services/studio-configuration.ts
```diff
@@ -0,0 +1,12 @@
...
+}
+
+interface StudioSettings {
+  mediaPreviewUrl: string
+}
+
+export interface StudioConfiguration {
+  settings: StudioSettings
+  blueprintConfiguration: BlueprintConfiguration
+}
```
LindvedKrvang, line 12: I didn't catch this the first time around, but this file is in the wrong place.
You have placed in the `services` folder which is a folder reserved for Services, like the `ConfigurationService`.
The definitions of our interfaces and classes should be in our `models` folder.

LindvedKrvang (approved):
Looks good :)
The approve is with the assumption that the `studio-configuration.ts` is going to be moved to the correct place (see the comment).

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -56,7 +56,7 @@ export class MiniShelfComponent implements OnInit, OnDestroy, OnChanges {
     if (!this.studioConfiguration) return
     if (this.media.duration < this.studioConfiguration.blueprintConfiguration.ServerPostrollDuration) return
 
-    this.mediaCalculatedDuration = this.media.duration - this.studioConfiguration.blueprintConfiguration.ServerPostrollDuration
+    this.mediaCalculatedDuration = Math.max(this.media.duration - this.studioConfiguration.blueprintConfiguration.ServerPostrollDuration, 0)
```
LindvedKrvang, line 59: Since you are now using `Math.max` the conditional on line 57 is no longer needed and should ideally be removed

LindvedKrvang (approved):

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -47,6 +60,20 @@ export class RundownComponent implements OnInit, OnDestroy {
     this.startOffsetsInMsFromPlayheadForSegments = this.getStartOffsetsInMsFromPlayheadForSegments(rundownTimingContext)
   }
 
+  private onActionsChanged(actions: Action[]): void {
+    this.videoClipActions = actions.filter((action): action is Tv2VideoClipAction => {
+      return action && (<Tv2Action>action).metadata?.contentType === Tv2ActionContentType.VIDEO_CLIP
+    })
```
KvelaGorrrrnio, line 66: When you have a single expression to return from a lambda function, the curly brackets and the return keyword are redundant. The check for whether the action is a truthy value should not be necessary. And the type assertion should be changed to `action as Tv2Action`. So in all this can be simplified to:
```typescript
    this.videoClipActions = actions.filter((action): action is Tv2VideoClipAction => (action as Tv2Action).metadata?.contentType === Tv2ActionContentType.VIDEO_CLIP)
```

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -64,9 +91,29 @@ export class RundownComponent implements OnInit, OnDestroy {
...
+    this.miniShelfSegmentActionMappings = this.miniShelfSegments.reduce(this.miniShelfSegmentsReducer.bind(this), {})
+  }
+
+  private miniShelfSegmentsReducer(actionMap: Record<string, Tv2VideoClipAction>, segment: Segment): Record<string, Tv2VideoClipAction> {
+    const videoClipFile: string | undefined = segment.metadata?.miniShelfVideoClipFile
+    if (!videoClipFile) return actionMap
+
+    const action: Tv2VideoClipAction | undefined = this.videoClipActions.find(action => {
+      return action.metadata?.fileName === videoClipFile
+    })
```
KvelaGorrrrnio, line 115: The curly brackets and the return keyword are redundant and can be removed.

src/app/shared/services/media-state.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  ) {
+    this.logger = logger.tag('MediaStateService')
+  }
+
+  public async getMedia(id: string): Promise<Media> {
+    try {
+      return await lastValueFrom(this.fetchMedia(id))
+    } catch (error) {
+      this.logger.data(error).error('Failed to fetch media.')
+      return await Promise.reject(error)
```
KvelaGorrrrnio, line 23: The await keyword is redundant and can be removed – note that this will lead to the async keyword in the method definition can also be removed.

src/app/rundown-view/components/minishelf/mini-shelf.component.html
```diff
@@ -0,0 +1,20 @@
...
+<!--  TODO: add ThumbnailComponent and use it here-->
+    <img
+        class="c-mini-shelf__thumbnail"
+        src="{{ mediaPreviewUrl }}"
+        alt="{{ segment.name? segment.name : 'Title is not available.' }}"
+        (error)=handleMissingImage($event)
+    />
+    <div class="c-mini-shelf__note">
+        <span class="c-mini-shelf__title"
+        >{{ getSanitizedTitle() ? getSanitizedTitle() : 'Title is not available.'  | slice:0:30 | uppercase }}{{ segment.name.length > 30 ? '...' : '' }}</span>
```
KvelaGorrrrnio, line 15: The newlines for the span tag seem a bit odd here. And the the clipping of the segment name, can be done with a `max-width: <some-number>%;` and `text-overflow: ellipsis;` on the `.c-mini-shelf__title` class. That way the browser determines when to clip the segment name based on available space.

src/app/rundown-view/components/minishelf/mini-shelf.component.html
```diff
@@ -0,0 +1,20 @@
+<div
+  class="c-mini-shelf"
+  title="{{ segment.name }}"
+  (click)="executeAction()"
```
KvelaGorrrrnio, line 4: Since the associated action is executed, when clicking on the mini shelf, it would be nice with some visual feedback that it is possible. It would be nice if the cursor changed to show that one could interact with the mini shelf and perhaps make a slight style variation on hover.

src/app/rundown-view/components/minishelf/mini-shelf.component.scss
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  &__duration {
+    color: yellow;
```
KvelaGorrrrnio, line 39: Personal opinion incoming: I don't see why this should be yellow, instead of white as the rest of the text.

src/app/core/parsers/zod-entity-parser.service.spec.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: And type assertions in general 😄

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
KvelaGorrrrnio: > Idea was to represent facts - i.e. if actual duration can not be determined - display --:-- which differs than 00:00 since latter is not true.

I think this comment was to another thread?
But to answer it: I think the idea is nice. The easy solution would be to not display a duration if the duration is set to 0. But I don't mind it just saying 00:00.

src/app/shared/services/media-state.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  private readonly logger: Logger
+
+  constructor(
+    private readonly mediaService: MediaService,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('MediaStateService')
+  }
+
+  public async getMedia(id: string): Promise<Media> {
```
KvelaGorrrrnio, line 18: The purpose of state services are to hold the (cached) state of entity/subject that are created for and notify users of the state, when it changes through observables. This class does not contain any state handling, so either it should be removed, and the `mediaService` should be used directly in the mini shelf, or it should be changed to a state service acting as the other state services.

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    return this.segment.name
+      .replace(/<[^>]*>/g, '') // remove html tags
+      .replace(/(\r\n|\n|\r)/gm, ' ') // remove newlines
+      .replace(/_/g, ' ') // replace underscores with spaces;
+      .replace(/&nbsp;/g, ' ') // remove &nbsp;
+      .replace(/\s+/g, ' ') // remove extra spaces
+      .replace(/&amp;/g, '&') // remove &amp;
+      .replace(/&quot;/g, '"') // remove &quot;
+      .replace(/&apos;/g, "'") // remove &apos;
+      .trim() // and finally trim;
```
KvelaGorrrrnio, line 89: Is this necessary? Have we encountered some of these special characters?

k-primeholding-dim: I see characters out of the set [letter, digit], and I assume is possible to encounter html breaker.

src/app/rundown-view/components/rundown/rundown.component.scss
```diff
@@ -6,7 +6,7 @@
   overflow: auto;
 
   .segments {
-    margin: 20px 0 300px 0;
+    margin: 0;
```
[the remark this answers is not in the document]
k-primeholding-dim: In order to comply with the requested positioning.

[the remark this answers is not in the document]
KvelaGorrrrnio: But this styling does not affect the mini shelf. It targets the spacing at the top and bottom of the rundown view.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #93, opened 2024-01-25
SOF-1721 - Cycle MiniShelves back and forth

k-primeholding-dim:
Uses MiniShelfComponent from https://github.com/tv2/ng-sofie/pull/88 and handles key events.

RundownComponent to listen for keyboard events like:
- Tab - find the next VideoClipAction in the next MiniShelf and insert it as Next. Continuously pressing “Tab” will find the next MiniShelf's VideoClipAction after that (but before a regular Segment) and insert it.
- Shift-Tab - advances in opposite direction of Tab


KvelaGorrrrnio:
Since this is based on #88, it would be nice if the target branch is changed to `SOF-1720/display-minishelf` and a note in the PR description that it is based on another feature branch/ PR.

k-primeholding-dim:
This feature branch should reach develop once complete.

LindvedKrvang:
> This feature branch should reach develop once complete. My other option was to block till #88 is merged.

No need to block this task untill #88 is merged and no need to not target #88 from PR. This PR will still reach `develop`.

PRs that have been made from other branches should still target the branch they have been made from. If they don't and target develop, like here, we have no idea which changes is the changes relevant for this PR (which are changes that should be reveiwed in this PR), and which changes are for the target PR (which are changes that should not be reveiwed in this PR).

PRs that are targeting other PRs will have their target branch updated automatically when their target PR is merged. By not targeting the PR/branch this is branched from, you are "blocking" us from reviewing this PR untill the original branch/PR has been merged.

The important thing is that _this_ PR should _not_ be merged untill the target PR has been merged.

src/app/shared/services/action-state.service.ts
```diff
@@ -118,4 +118,8 @@ export class ActionStateService {
     this.actionsSubjects.forEach(subject => subject.complete())
     this.eventSubscriptions.forEach(eventSubscription => eventSubscription.unsubscribe())
   }
+
+  public executeAction(actionId: string, rundownId: string): void {
+    this.actionService.executeAction(actionId, rundownId).subscribe()
+  }
```
LindvedKrvang, line 124: This is just a "parse-through" method. 
You should just inject `ActionService` directly into the component that needs to execute the Action

LindvedKrvang (changes requested):
The request for changes is primarily because you have introduced a lot of static methods which goes against a lot of the principles that we are following

k-primeholding-dim:
Closing this one, work continues in https://github.com/tv2/ng-sofie/pull/99

PR #99, opened 2024-02-01
SOF-1721 - MiniShelf cycling

LindvedKrvang:
Replaces: https://github.com/tv2/ng-sofie/pull/93

Based on (but heavily modified):
Proof of concept for dividing miniShelfs into groups and take the first shelf of the group, depending on placement in the Rundown
Later, evolved in deterministic cycling (as depicted here https://github.com/tv2/ng-sofie/pull/99#discussion_r1489039775 )

As result introduces:
- MiniShelfCycleService (executes actions as result of handling directional cycling)
- MiniShelfNavigationService (tracks and provides data for MiniShelfCycleService)

Listening for keys:
- Tab - find the next VideoClipAction in the next MiniShelf and insert it as Next. Continuously pressing “Tab” will find the next MiniShelf's VideoClipAction after that (but before a regular Segment) and insert it (from the first in the group, down and over).
- Shift-Tab - advances in opposite direction of Tab (from the last in the group, up and over)


src/app/rundown-view/components/minishelf/mini-shelf.component.html
```diff
@@ -1,5 +1,9 @@
 <div
-  class="c-mini-shelf"
+  class="c-mini-shelf
+{{ showOnAirBorder && !showNextBorder ? 'c-mini-shelf--on-air' : '' }}
+{{ !showOnAirBorder && showNextBorder ? 'c-mini-shelf--next' : '' }}
+{{ showOnAirBorder && showNextBorder ? 'c-mini-shelf--on-air-and-next' : '' }}
+{{ !showOnAirBorder && !showNextBorder ? 'c-mini-shelf--normal' : '' }}"
```
LindvedKrvang, line 6: You can simplify this a lot by writing:
image

Then, in your CSS you can write:
image

Notice how the `c-mini-shelf-normal` is gone?
You use it set the default border of the MiniShelf, so that can just be an attribute on `c-mini-self`:
image


CSS selectors works on a priority system. The more specific the selector, the higher the priority and we can utilize that to write less CSS. If the priority is equal, then it is just whatever has been read last. Hence `.next` is above `.on-air` in my example so `.on-air.next` only need to set the right border.


Note: The names I selected for my CSS classes is mainly meant for demonstrating purposes.

src/app/rundown-view/components/minishelf/mini-shelf.component.html
```diff
@@ -11,7 +15,7 @@
         (error)=handleMissingImage($event)
     />
     <div class="c-mini-shelf__note">
-        <span class="c-mini-shelf__title">{{ getSanitizedTitle() | uppercase }}</span>
+        <span class="c-mini-shelf__title">{{ getSanitizedTitle() | uppercase }} {{ showOnAirBorder ? 'ON-AIR' : ''}} {{ showNextBorder ? 'NEXT' : ''}}</span>
```
LindvedKrvang, line 18: We should not show with text that a minishelf is either `ON AIR` ot `NEXT`.
So this needs to be removed again.

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -16,29 +18,52 @@ import { Logger } from '../../../core/abstractions/logger.service'
...
+      const tv2Part: Tv2Part | undefined = onAirPart as Tv2Part | undefined
+      this.showOnAirBorder = !tv2Part || !this.videoClipAction ? false : tv2Part?.metadata?.actionId === this.videoClipAction?.id
+    })
+
+    this.rundownStateService.subscribeToNextPart(this.segment.rundownId).subscribe(nextPart => {
+      const tv2Part: Tv2Part | undefined = nextPart as Tv2Part | undefined
+      this.showNextBorder = !tv2Part || !this.videoClipAction ? false : tv2Part?.metadata?.actionId === this.videoClipAction?.id
+    })
+
+    await this.rundownStateService.subscribeToRundown(this.segment.rundownId).then(rundownObservable => {
```
LindvedKrvang, line 59: Just an FYI: The current implementation of `subscribeToRundown` is firing a lot of events. Just keep that in mind when deciding to use this one :) 

Also: This doesn't clear the borders when we Reset the Rundown.

src/app/rundown-view/services/mini-shelf-state.service.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+import { Tv2SegmentMetadata } from '../../core/models/tv2-segment-metadata'
+import { Tv2ActionContentType, Tv2VideoClipAction } from '../../shared/models/tv2-action'
+import { Observable } from 'rxjs'
+import { PartSetAsNextEvent, PartTakenEvent } from '../../core/models/rundown-event'
+
+describe(MiniShelfStateService.name, (): void => {
+  it('should create', (): void => {
+    const testee: MiniShelfStateService = createMinimalTestee()
+    expect(testee).toBeTruthy()
+  })
```
LindvedKrvang, line 16: This test is redundant. 
All the other test would fail if this failed.

src/app/rundown-view/services/mini-shelf-state.service.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  describe(MiniShelfStateService.prototype.setActions.name, (): void => {
+    it('should be defined', (): void => {
+      const testee: MiniShelfStateService = createMinimalTestee()
+      expect(testee.setActions).toBeDefined()
+    })
```
LindvedKrvang, line 22: This test is redundant (and all the other like it).
We are in Typescript. If the method doesn't exist we are going to compile errors all the places we use it.

src/app/rundown-view/services/mini-shelf-state.service.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+    it('should set the miniShelfGroups', (): void => {
+      const testee: MiniShelfStateService = createMinimalTesteeMiniShelfGroups()
+      expect(testee['miniShelfGroups'].size).toEqual(1)
+    })
+  })
+
+  describe(MiniShelfStateService.prototype.executeVideoClipActionForSegment.name, (): void => {
+    it('should be defined', (): void => {
+      const testee: MiniShelfStateService = createMinimalTestee()
```
LindvedKrvang, line 50: You are delegating all the important information about what this test is testing into a helper method.
That opens up for the possibillity that someone might break this test when changing that helper method when creating a different test.

src/app/rundown-view/services/mini-shelf-state.service.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  private isMiniShelf(segment: Segment): boolean {
+    return !!segment.metadata?.miniShelfVideoClipFile && segment.isHidden
+  }
+
+  public setActions(miniShelfSegmentActionMappings: Record<string, Tv2VideoClipAction>): void {
+    this.miniShelfSegmentActionMappings = miniShelfSegmentActionMappings
+  }
+
+  public executeVideoClipActionForSegment(segment: Segment): void {
```
LindvedKrvang, line 64: This should be a private method

src/app/rundown-view/services/mini-shelf-state.service.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (miniShelfGroup !== this.lastExecutedMiniShelfGroup) {
+      this.lastExecutedMiniShelfGroup = miniShelfGroup
+      this.currentMiniShelfIndex = 0
+    } else {
+      this.currentMiniShelfIndex = this.currentMiniShelfIndex >= miniShelfGroup.length - 1 ? 0 : this.currentMiniShelfIndex + 1
+    }
+    this.executeVideoClipActionForSegment(miniShelfGroup[this.currentMiniShelfIndex])
+  }
+
+  public findMiniShelfGroup(): Segment[] {
```
LindvedKrvang, line 97: This should be a private method

LindvedKrvang (commented):
We need to have a talk about testing

src/app/rundown-view/components/rundown/rundown.component.spec.ts
```diff
@@ -39,13 +101,58 @@ async function configureTestBed(
...
+  const mockedSubscriptionToRundownAutoNext: Subscription = mock<Subscription>()
+  when(mockedRundownEventObserver.subscribeToRundownAutoNext(anything())).thenReturn(instance(mockedSubscriptionToRundownAutoNext))
+  const mockedSubscriptionToRundownSetNext: Subscription = mock<Subscription>()
+  when(mockedRundownEventObserver.subscribeToRundownSetNext(anything())).thenReturn(instance(mockedSubscriptionToRundownSetNext))
+  const mockedSubscriptionToRundownReset: Subscription = mock<Subscription>()
+  when(mockedRundownEventObserver.subscribeToRundownReset(anything())).thenReturn(instance(mockedSubscriptionToRundownReset))
+  const mockedSubscriptionToRundownUpdates: Subscription = mock<Subscription>()
+  when(mockedRundownEventObserver.subscribeToRundownUpdates(anything())).thenReturn(instance(mockedSubscriptionToRundownUpdates))
+  return mockedRundownEventObserver
 }
```
LindvedKrvang, line 158: This goes for all your methods: Your methods would benefit greatly from having some newlines that would "group" the different things happening in your code:
```
function createMockOfRundownEventObserver(): RundownEventObserver {
  const mockedRundownEventObserver: RundownEventObserver = mock<RundownEventObserver>()
  const mockedSubscriptionToRundownActivation: Subscription = mock<Subscription>()
  when(mockedRundownEventObserver.subscribeToRundownActivation(anything())).thenReturn(instance(mockedSubscriptionToRundownActivation))

  const mockedSubscriptionToRundownDeactivation: Subscription = mock<Subscription>()
  when(mockedRundownEventObserver.subscribeToRundownDeactivation(anything())).thenReturn(instance(mockedSubscriptionToRundownDeactivation))

  const mockedSubscriptionToRundownAutoNext: Subscription = mock<Subscription>()
  when(mockedRundownEventObserver.subscribeToRundownAutoNext(anything())).thenReturn(instance(mockedSubscriptionToRundownAutoNext))

  const mockedSubscriptionToRundownSetNext: Subscription = mock<Subscription>()
  when(mockedRundownEventObserver.subscribeToRundownSetNext(anything())).thenReturn(instance(mockedSubscriptionToRundownSetNext))

  const mockedSubscriptionToRundownReset: Subscription = mock<Subscription>()
  when(mockedRundownEventObserver.subscribeToRundownReset(anything())).thenReturn(instance(mockedSubscriptionToRundownReset))

  const mockedSubscriptionToRundownUpdates: Subscription = mock<Subscription>()
  when(mockedRundownEventObserver.subscribeToRundownUpdates(anything())).thenReturn(instance(mockedSubscriptionToRundownUpdates))

  return mockedRundownEventObserver
}
```

src/app/rundown-view/components/rundown/rundown.component.spec.ts
```diff
@@ -10,11 +10,66 @@ import { Logger } from '../../../core/abstractions/logger.service'
...
+    verify(mockedRundownTimingContextStateService.subscribeToRundownTimingContext(mockedRundown.id)).once()
+    verify(mockedActionStateService.subscribeToRundownActions(mockedRundown.id)).once()
+
+    verify(mockedRundownEventObserver.subscribeToRundownDeactivation(anything())).once()
+    verify(mockedRundownEventObserver.subscribeToRundownAutoNext(anything())).once()
+    verify(mockedRundownEventObserver.subscribeToRundownSetNext(anything())).once()
+    verify(mockedRundownEventObserver.subscribeToRundownReset(anything())).once()
+    verify(mockedRundownEventObserver.subscribeToRundownUpdates(anything())).once()
+  })
+  it('should update MiniShelves and ActionMappings upon change in rundown', async () => {
```
LindvedKrvang, line 52: This test is also testing for two different responsibilities.

Sidenote: It would be nice if you had a newline between every test :)

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -86,6 +88,7 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
   }
 
   public ngOnChanges(changes: SimpleChanges): void {
+    this.logger.data(changes).debug('RundownComponent ngOnChanges')
```
LindvedKrvang, line 91: A wild debugger statement appears.
Would be nice to remove again :)

src/scss/_sofie-color-vars.scss
```diff
@@ -8,7 +8,9 @@ body {
   --black-color: #000000;
   --dark-color: #3a3a3a;
   --background-color: #343434;
+  --red-color: red;
   --green-color: green;
+  --bright-green-color: #00AC17;
```
LindvedKrvang, line 13: `--read-color` and `--bright-green-color` seems to be used for the OnAir- and NextBorder colors of the MiniShelves.
We already have the variables `--on-air-color` and `--next-color` on line 19 and 20 in this file. These variables should be used, so we don't have multiple variables for the same thing (thus defeating the point of the variables)

src/scss/_sofie-color-vars.scss
```diff
@@ -47,6 +49,8 @@ body {
   --gradient-black-rgba-with-06-opacity-color: rgba(0, 0, 0, 0.6);
   --gradient-black-rgba-with-08-opacity-color: rgba(0, 0, 0, 0.8);
   --gradient-green-rgba-with-04-opacity-color: rgba(0, 222, 99, 0.4);
+  --gradient-green-rgba-with-08-opacity-color: rgba(0, 222, 99, 0.8);
+  --gradient-red-rgba-with-08-opacity-color: rgba(255, 0, 0, 0.8);
```
LindvedKrvang, line 53: I can't find any usages of these variables. If that's true, they should be deleted again.

src/app/rundown-view/services/mini-shelf-state.service.ts
```diff
@@ -0,0 +1,114 @@
+import { Injectable } from '@angular/core'
+import { Rundown } from '../../core/models/rundown'
+import { Segment } from '../../core/models/segment'
+import { ActionService } from '../../shared/abstractions/action.service'
+import { RundownEventObserver } from '../../core/services/rundown-event-observer.service'
+import { PartTakenEvent, RundownActivatedEvent } from '../../core/models/rundown-event'
+import { Tv2VideoClipAction } from '../../shared/models/tv2-action'
+
+@Injectable()
+export class MiniShelfStateService {
```
KvelaGorrrrnio, line 10: In the project, we have a concept of state services which are classes that are responsible for holding the state of the subject they describe. The state services have internal mechanisms that react to events from the backend (or another source) in order to update and notify any subscribers about changes to the state.

Since this service seems to be told what the state is in order to compute and act on the given data, and does not expose the state for other classes to depend on, I don't see it as a state service in the same sense.

I suggest that the different responsibilities be separated out into their own classes, such that there is one responsibility of navigating/finding mini shelf groups; one that is handling the mini shelf state; and a mediator, which is allowed to use the two others to coordinate and act upon the outcome. By having the navigation and searching for mini shelf groups in a separate class, we have isolated the business logic of the feature and can easily test it.

 I have added a quickly sketched diagram below as an example. Note that this is just a proposal and should at best be used as inspiration :)

```mermaid
graph TD

RundownStateService["
  Runndown State Service
  [Service]
"]

ActionStateService["
  Action State Service
  [Service]
"]

MediaStateService["
  Media State Service
  [Service]
"]

MiniShelfStateService["
  Mini Shelf State Service
  [Service]

  Is responsible for merging action and media data
  with mini shelf segments and expose them via
  observables.
"]

MiniShelfNavigationService["
  Mini Shelf Navigation Service
  [Service]

  Given a rundown and the current mini
  shelf segment, sit calculates what the next
  should be.
"]

MiniShelfCycleService["
  Mini Shelf Cycle Service
  [Service]

  Is the mediator that can be called to orchestrate
  the cycling and the execution of it.
"]

ActionService["
  Action Service
  [Service]

  Is responsible for executing actions.
"]

KeyBindingService["
  Key Binding Service
  [Service]
"]

KeyBindingService --> |"triggers minishelf cycling via"| MiniShelfCycleService

MiniShelfCycleService --> |"asks for next/prev mini shelf from"| MiniShelfNavigationService
MiniShelfCycleService --> |"subscribes to mini shelf data from"| MiniShelfStateService
MiniShelfCycleService --> |"executes the mini shelf action via"| ActionService

MiniShelfStateService --> |"subscribes to actions from"| ActionStateService
MiniShelfStateService --> |"subscribes to rundowns from"| RundownStateService
MiniShelfStateService --> |"subscribes to media from"| MediaStateService
```

KvelaGorrrrnio (commented):
I tried to run it locally and noticed a bug, when I have taken and/or cycled a mini shelf as next, then if the page is reloaded, both the on air and set as next markings are gone.

src/scss/_sofie-color-vars.scss
```diff
@@ -9,6 +9,7 @@ body {
   --dark-color: #3a3a3a;
   --background-color: #343434;
   --green-color: green;
+  --bright-green-color: #00AC17;
```
LindvedKrvang, line 12: This doesn't seem to be used anywhere. If that's true, it should be deleted.

LindvedKrvang (commented):
Since this is technically my PR, I'm unable to approve it. So this is a pseudo approve

src/app/rundown-view/components/rundown/rundown.component.spec.ts
```diff
@@ -10,11 +10,210 @@ import { Logger } from '../../../core/abstractions/logger.service'
...
+
+    verify(mockedRundownEventObserver.subscribeToRundownSetNext(anything())).once()
+  })
+
+  it('should subscribe to RundownReset changes upon init', async () => {
+    const mockedMiniShelfStateService: MiniShelfStateService = createMockOfMiniShelfStateService()
+    const mockedRundownTimingContextStateService: RundownTimingContextStateService = createMockOfRundownTimingContextStateService()
+    const mockedActionStateService: ActionStateService = createMockOfActionStateService()
+    const mockedRundownEventObserver: RundownEventObserver = createMockOfRundownEventObserver()
+    const mockedRundown: Rundown = new TestEntityFactory().createRundown()
```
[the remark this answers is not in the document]
LindvedKrvang: I would be very mindfull of that since a little change to the "default mock" could potentially break a lot of tests because you lose the overview of what part of the mock is being used for which test.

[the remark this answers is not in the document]
LindvedKrvang: And when writing new tests you would be drawn to alter the "default mock" since that is where you mocks are and every test needs a slightly different mock setup. So in reality developers would constantly be updating the "default mock" requiring them to always be aware of all the test in the suite using that mock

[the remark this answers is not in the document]
KvelaGorrrrnio: If you alter the default mock in a way that breaks tests, then you should get quick feedback to indicate it. I have no strong feelings about it, so if this is disregarded, then fine :D

src/app/rundown-view/factories/key-binding.factory.ts
```diff
@@ -198,6 +204,24 @@ export class KeyBindingFactory {
...
+  }
+
+  private cycleMiniShelfBackward(rundown: Rundown): void {
+    try {
+      this.miniShelfCycleService.cycleMiniShelfBackward(rundown)
+    } catch (error) {
+      this.logger.data(error).error('Failed cycling backward in mini shelves.')
+      this.snackBar.open('Unable to cycle backward in mini shelves.', undefined, { panelClass: 'snackbar-danger', duration: 3000 })
+    }
+  }
```
LindvedKrvang, line 223: The error handling for cycling the miniShelf does not belong in the `KeyBindingFactory`. This factory should just bind a key to something and that something should handle the rest.

The responsibility of error handling for cycling MiniShelf belongs in the `MiniSehlfCycleService`.

This means that all the KeyBindingFactory should do is:
```
this.createRundownKeyBinding('Cycle MiniShelf', ['Tab'], () => this.miniShelfCycleService.cycleMiniShelfForward(rundown)),
this.createRundownKeyBinding('Cycle MiniShelf', ['Shift', 'Tab'], () => this.miniShelfCycleService.cycleMiniShelfBackward(rundown)),
```

By moving the error handling to `CycleMiniShelfService` also means that we don't need to re-implemt the error handling if we want to trigger the cycling from somewhere else in the system.

src/app/rundown-view/services/mini-shelf-cycle.service.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (!params?.miniShelfNavigationService) {
+    const miniShelfNavigationServiceMock: MiniShelfNavigationService = mock<MiniShelfNavigationService>()
+    miniShelfNavigationService = instance(miniShelfNavigationServiceMock)
+  }
+
+  let actionStateService: ActionStateService
+  if (!params?.actionStateService) {
+    const actionStateServiceMock: ActionStateService = mock<ActionStateService>()
+    actionStateService = instance(actionStateServiceMock)
+  }
```
LindvedKrvang: This can be simplified to:
```
const miniShelfNavigationService: MiniShelfNavigationService = params?.miniShelfNavigationService ?? instance(mock<MiniShelfNavigationService>())
const actionStateService: ActionStateService = params?.actionStateService ?? instance(mock<ActionStateService>())

...

return new MiniShelfCycleService(miniShelfNavigationService, actionStateService, params?.actionService ?? actionService)
```

LindvedKrvang (commented):
On my local machine the navgiation is not working correctly.
It doesn't seem to be able to find the first MiniShelf group. The other groups works fine, but when it should take the first MiniShelf group it takes from the other groups. See video.
I'm also not able to continue to cycle when I should be on the first group.

https://github.com/tv2/ng-sofie/assets/23049647/914d9169-18c8-4cd8-88d9-638dae728d4d

src/app/rundown-view/factories/key-binding.factory.ts
```diff
@@ -114,6 +118,8 @@ export class KeyBindingFactory {
         this.createRundownKeyBinding('Set Segment Below as Next', ['Shift', 'ArrowDown'], () => this.setSegmentBelowNextAsNext(rundown)),
         this.createRundownKeyBinding('Set Earlier Part as Next', ['Shift', 'ArrowLeft'], () => this.setEarlierPartAsNext(rundown)),
         this.createRundownKeyBinding('Set Later Part as Next', ['Shift', 'ArrowRight'], () => this.setLaterPartAsNext(rundown)),
+        this.createRundownKeyBinding('Cycle MiniShelf D->', ['Tab'], () => this.cycleMiniShelfForward(rundown)),
+        this.createRundownKeyBinding('Cycle MiniShelf D<-', ['Shift', 'Tab'], () => this.cycleMiniShelfBackward(rundown)),
```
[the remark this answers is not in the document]
k-primeholding-dim: Negative, so is removed.

src/app/rundown-view/components/rundown/rundown.component.spec.ts
```diff
@@ -10,11 +10,109 @@ import { Logger } from '../../../core/abstractions/logger.service'
...
+      mockedRundown: mockedRundown,
+      mockedRundownEventObserver: mockedRundownEventObserver,
+      mockedRundownTimingContextStateService: mockedRundownTimingContextStateService,
+      mockedActionStateService: mockedActionStateService,
+    })
+
+    verify(mockedRundownEventObserver.subscribeToRundownSetNext(anything())).once()
+  })
+
+  it('should subscribe to RundownReset changes upon init', async () => {
```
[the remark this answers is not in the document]
k-primeholding-dim: Implementation tests are dropped.

k-primeholding-dim (approved):

src/app/rundown-view/components/mini-shelf/mini-shelf.component.scss
```diff
@@ -3,46 +3,58 @@
...
+  box-shadow: 0 0 2px 2px var(--gradient-black-rgba-with-02-opacity-color);
 
   &:hover {
     cursor: pointer;
-    border: solid 1px var(--white-color);
-    box-shadow: 0 0 3px 3px var(--gradient-green-rgba-with-04-opacity-color);
+    box-shadow: 0 0 2px 2px var(--gradient-green-rgba-with-04-opacity-color);
+  }
+
+  &.is-on-air {
```
RasmusAlbrektsen, line 17: So far, in the scss files, we haven't prefixed things like these with `is`, so maybe these should just be renamed like so:
`is-on-air` => `on-air`
`is-next` => `next`

src/app/rundown-view/services/mini-shelf-navigation.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  private getCurrentMiniShelfActionId(rundown: Rundown): string | undefined {
+    const onAirPart: Tv2Part = this.getOnAirPart(rundown)
+    const nextPart: Tv2Part = this.getNextPart(rundown)
+    return nextPart?.metadata?.actionId ?? onAirPart.metadata?.actionId
+  }
+
+  private getOnAirPart(rundown: Rundown): Tv2Part {
+    const nextPart: Part | undefined = rundown.segments.find(segment => segment.isOnAir)?.parts.find(part => part.isOnAir)
+
```
RasmusAlbrektsen, line 27: This is just personal preference, but everywhere else i found we usually dont have the newline between the constant and the if-statement checking it.

RasmusAlbrektsen: If you change it, remember that it is also the case for the `getNextPart` method.

RasmusAlbrektsen (approved):
Just a few comment, but from the testing i managed to do it, it seems to work just fine for me :-)

LindvedKrvang (commented):
Since this is still technical my PR, I am still unable to approve it. So here is a "pseudo" approve again :)

PR #100, opened 2024-02-05
SOF-1786 - fix timeline Marker Fill Color to come from ts file

vivanov90:
Bug fixtimeline Marker Fill Color to come from ts file

### DIFF
--- a/src/app/rundown-view/components/timeline-markers/timeline-markers.component.ts
+++ b/src/app/rundown-view/components/timeline-markers/timeline-markers.component.ts
@@ -40,7 +40,6 @@ export class TimelineMarkersComponent implements AfterViewInit, OnChanges {
 
   private readonly secondsPerSection: number = 5
   private readonly subsectionsPerSection: number = 5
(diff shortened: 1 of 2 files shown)


src/app/rundown-view/components/timeline-markers/timeline-markers.component.ts
```diff
@@ -40,6 +40,7 @@ export class TimelineMarkersComponent implements AfterViewInit, OnChanges {
 
   private readonly secondsPerSection: number = 5
   private readonly subsectionsPerSection: number = 5
+  private readonly timelineMarkerFillColor: string = '#5f6164'
```
KvelaGorrrrnio, line 43: If we still want to use the css variables, then we can use the computed style of the canvas element like so:
```typescript
getComputedStyle(this.canvasElement.nativeElement).getPropertyValue('--dark-gray-color')
```

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

PR #101, opened 2024-02-06
SOF-1652 Show Piece Media Availability

RasmusAlbrektsen:
This PR introduces the media availability status of piece media by showing red stripes over the piece. This also includes the handling of media created and deleted events.


src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -166,44 +197,21 @@ export class ZodEntityParser implements EntityParser {
...
+  public parseMediaAsset(media: unknown): Media {
+    return this.mediaAssetParser.parse(media)
   }
 
-  private readonly systemInformationParser = zod.object({
-    name: zod.string(),
-  })
+  public parseMediaAssets(media: unknown): Media[] {
+    return this.mediaAssetsParser.parse(media)
+  }
```
LindvedKrvang, line 214: Seems like one of these two methods are redundant

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.scss
```diff
@@ -13,13 +13,32 @@
...
+      top: 0;
+      right: 0;
+      bottom: 0;
+      left: 0;
+      background: repeating-linear-gradient(
+                      45deg,
+                      #00000000,
+                      #00000000 12px,
+                      #9a0d0daa 12px,
+                      #9a0d0daa 18px
```
LindvedKrvang, line 36: Wild colors appears. We don't do that no more ;)

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -33,6 +36,51 @@ export class OffsetablePieceComponent {
   @ViewChild('labelTextElement')
   public labelTextElement: ElementRef<HTMLSpanElement>
 
+  public media: Media | undefined
+
+  public isMediaPiece: boolean = false
+
+  private mediaSubscription?: Subscription
```
LindvedKrvang, line 43: Since `mediaSubscription` is optional you could use that as your `isMediaPiece`.
If the subscription is `undefined` it is not a media piece. If it is not undefined, then it is. 
One less variable to maintain :)

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -33,6 +36,51 @@ export class OffsetablePieceComponent {
...
+      this.updatePieceMedia()
+    }
+  }
+
+  private doesPieceContainMedia(): boolean {
+    const piece: Tv2Piece = this.piece as Tv2Piece
+    if (!piece.metadata.sourceName) {
+      return false
+    }
+    return !!piece.metadata.sourceName
```
LindvedKrvang, line 81: I haven't checked, but `return !!piece.metadata.sourceName` should be enough.

src/app/shared/services/media-state.service.ts
```diff
@@ -1,33 +1,81 @@
...
+    if (mediaSubject) {
+      mediaSubject.next(media)
+      return
+    }
+    this.mediaSubjects.set(this.getMediaSubjectKey(media.sourceName), new BehaviorSubject<Media | undefined>(media))
+  }
+
+  private getMediaSubjectKey(sourceName: string): string {
+    return sourceName.toUpperCase()
+  }
```
LindvedKrvang, line 39: Here you uppercase the key of the `mediaSubjects` but further down on line 52 you just set the `media.sourceName` as the key directly.
That seems like it will lead to multiple mediaSubjects for the same Media.

src/app/shared/services/media-state.service.ts
```diff
@@ -1,33 +1,81 @@
...
+
+  private createMediaSubject(sourceName: string): BehaviorSubject<Media | undefined> {
+    const mediaSubjectKey: string = this.getMediaSubjectKey(sourceName)
+    const mediaSubject: BehaviorSubject<Media | undefined> | undefined = this.mediaSubjects.get(mediaSubjectKey)
+    if (mediaSubject) {
+      return mediaSubject
+    }
+    const subject: BehaviorSubject<Media | undefined> = new BehaviorSubject<Media | undefined>(undefined)
+    this.mediaSubjects.set(mediaSubjectKey, subject)
```
LindvedKrvang, line 74: This is the third time in this file where you set `this.mediaSubjects.set(..., ...)` i.e. this the third time where you create a new BehaviourSubject.

Why not have something like:
```
private getMediaSubject(sourceName: string): BehaviourSubject<Media | undefined> {
  if (!this.mediaSubjects.has(this.getMediaSubjectKey(sourceName)) {
    this.mediaSubjects.set(this.getMediaSubjectKey(sourceName), new BehaviourSubject<Media | undefined>(undefined)
  }
  return this.mediaSubjects(this.getMediaSubjectKey(sourceName)!
}
```

Then your `updateMedia` method becomes:
```
private updateMedia(media: Media): void {
  const mediaSubject = this.getMediaSubject(media.sourceName)
  mediaSubject.next(media)
}
```

And your `createMediaFromEvent` method becomes: 
```
private createMediaFromEvent(event: MediaCreatedEvent): void {
  const mediaSubject = this.getMediaSubject(event.media.sourceName)
  mediaSubject.next(event.media)
}
```

And `subscribteToMedia` becomes:
```
public subscribeToMedia(sourceName: string): Observable<Media | undefined> {
  return this.createMediaSubject(sourceName).asObservable()
}
```

LindvedKrvang: Notice how identical `updateMedia` and `createMediaFromEvent` is?
This would allow you to delete `createMediaFromEvent` and just use `updateMedia`:
```
this.subscriptions.push(this.mediaEventObserver.subscribeToMediaCreated((event: MediaCreatedEvent) => this.updateMedia(event.media)))
```

Further emphaszing why we shouldn't name stuff like `...FromEvent` because it is based on _where_/_how_ we want to use our methods instead of what they actually _do_.

src/app/core/parsers/zod-entity-parser.service.ts
RasmusAlbrektsen: They both are used, unless i don't understand what you are referencing. The plural `parseMediaAssets` is used when we gather all media assets from Sofie Server, and the `parseMediaAsset` for when we receive media through events such as created and deleted events.

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.scss
RasmusAlbrektsen: They are no more! Vars introduced!

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
RasmusAlbrektsen: It was enough!

src/app/shared/services/media-state.service.ts
RasmusAlbrektsen: Good catch, using it on line 34 was redundant since those values are the uppercased ones that we want to match when we create new subjects with undefined values from the piece metadata!

src/app/shared/services/media-state.service.ts
RasmusAlbrektsen: I made this refactor, a great suggestion and i must admit i should have thought more about the `FromEvent` naming  also!

k-primeholding-dim (approved):

src/app/shared/services/media-state.service.ts
KvelaGorrrrnio: I agree with the overall assessment, except with the statement that we shouldn't use `...FromEvent`. It does declare a premise for how it is used, but not _where_ it is used. The given alternative is to convert the method to a anonymous function. Personally I prefer `...FromEvent` over `on<Event>` syntax, since it describes it intent and what the responsibility is: Namely, unwrapping the event the event for the logic.

src/app/core/parsers/zod-entity-parser.service.ts
LindvedKrvang: I guess I got confused about `media` is both singular and plural :p

src/app/shared/services/media-state.service.ts
LindvedKrvang: It does declare a premise for _where_ it is used, but it does it by saying where it _can't_ be used.

The current implementation, after Rasmus refactor, has the method `updateMedia` which is used in two places. If you renamed it back to `updateMediaFromEvent` you can now only use it in _one_ of those places, because only one those places are from an event. So now you need a second method to do the same thing as `updateMediaFromEvent`.

But the `fromEvent` part still refers to something _outside_ of the scope of the method. The method does not and should not care about where it is called from as long as its contract is fullfilled (in this case, it gets a Media, doesn't matter where from). So the `fromEvent` is redundant and it limits the logical reuse of the method.

src/app/core/abstractions/entity-parser.service.ts
```diff
@@ -19,6 +19,7 @@ export abstract class EntityParser {
   public abstract parseShowStyleVariant(showStyleVariant: unknown): ShowStyleVariant
   public abstract parseStudioConfiguration(studioConfiguration: unknown): StudioConfiguration
   public abstract parseTv2Action(tv2Action: unknown): Tv2Action
-  public abstract parseMedia(media: unknown): Media
+  public abstract parseMediaAsset(media: unknown): Media
+  public abstract parseMediaAssets(media: unknown): Media[]
```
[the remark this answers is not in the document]
LindvedKrvang: Or the methods should be renamed to not include `assests` since we don't call them assets anywhere else.

I know that the methods then have the same name, but we could leverage method overloading for that.

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
```diff
@@ -46,15 +41,9 @@ export class MiniShelfComponent implements OnInit, OnDestroy, OnChanges {
...
-      .then(mediaObservable =>
-        mediaObservable.subscribe(media => {
-          this.setMedia(media)
-          this.calculateMediaDurationInMsWithoutPostroll()
-        })
-      )
-      .catch(error => this.logger.data(error).error(`Failed to update media for segment '${this.segment.name}' with id '${this.segment.id}'.`))
+    this.mediaStateService.subscribeToMedia(this.segment.metadata?.miniShelfVideoClipFile).subscribe(media => {
+      this.updateMedia(media)
+    })
```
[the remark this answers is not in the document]
LindvedKrvang: I would prefer that we didn't use the `.bind()` when we don't need to bind anything

src/app/shared/services/media-state.service.ts
```diff
@@ -26,42 +26,37 @@ export class MediaStateService implements OnDestroy {
   }
 
   private updateMedia(media: Media): void {
-    const mediaSubject: BehaviorSubject<Media | undefined> | undefined = this.mediaSubjects.get(media.sourceName)
-    if (mediaSubject) {
-      mediaSubject.next(media)
-      return
+    const mediaSubject: BehaviorSubject<Media | undefined> | undefined = this.getMediaSubject(media.sourceName)
```
LindvedKrvang, line 29: `this.getMediaSubject()` always returns a BehaviourSubject, so you can now write:
```
const mediaSubject: BehaviorSubject<Media | undefined>
```
instead of
```
const mediaSubject: BehaviorSubject<Media | undefined> | undefined
```

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -33,6 +36,43 @@ export class OffsetablePieceComponent {
   @ViewChild('labelTextElement')
   public labelTextElement: ElementRef<HTMLSpanElement>
 
+  public media: Media | undefined
```
LindvedKrvang, line 39: Can be shortened to `public media?: Media`

src/app/shared/services/media-state.service.ts
```diff
@@ -1,33 +1,76 @@
...
+        subject.next(undefined)
+      }
+    })
+  }
+
+  public subscribeToMedia(sourceName: string): Observable<Media | undefined> {
+    return this.createMediaSubject(sourceName).asObservable()
+  }
+
+  private createMediaSubject(sourceName: string): BehaviorSubject<Media | undefined> {
```
LindvedKrvang, line 62: You no longer need `createMediaSubject()`. 
`getMediaSubject()` does the exact same as `createMediaSubject()` so one of them should be deleted.

src/app/shared/services/media-state.service.ts
RasmusAlbrektsen: I followed Lindveds logic here because what he mentions about `updateMedia`, but also because the other method used for events is deleting a subject and that could easily become applicable in other cases where we would have the same issue. I understand wanting to signify the fact origin of something, but then we would have to accept that we need two methods that do really just do the same thing.

src/app/shared/services/http/http-media.service.ts
```diff
@@ -16,10 +16,23 @@ export class HttpMediaService implements MediaService {
     private readonly entityParser: EntityParser
   ) {}
 
-  public getMedia(id: string): Observable<Media> {
-    return this.http.get<HttpResponse<Media>>(`${environment.apiBaseUrl}/media/${id}`).pipe(
+  public getAllMedia(): Observable<Media[]> {
+    return this.http.get<HttpResponse<Media[]>>(`${environment.apiBaseUrl}/media`).pipe(
       catchError(error => this.httpErrorService.catchError(error)),
-      map(response => this.entityParser.parseMedia(response.data))
+      map(response => this.entityParser.parseMediaAssets(response.data))
```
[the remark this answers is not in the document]
RasmusAlbrektsen: I don't appreciate this issue either, but i must say that unless we find a completely different alternative for `media` i prefer using the `asset` suffix to differentiate. But the tech leads are more than welcome to set the line on this!

src/app/core/abstractions/entity-parser.service.ts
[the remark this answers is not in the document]
RasmusAlbrektsen: I have kept the `mediaAsset` approach for now. I think we need to figure out how we want this naming done in the code.

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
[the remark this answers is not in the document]
RasmusAlbrektsen: I just removed the unnecessary curly brackets and kept it as is.

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -33,6 +36,51 @@ export class OffsetablePieceComponent {
...
+      return
+    }
+    this.isMediaPiece = true
+    this.mediaSubscription = this.mediaStateService.subscribeToMedia(mediaSourceName).subscribe(this.updateMediaAvailabilityStatus.bind(this))
+  }
+
+  public ngOnChanges(changes: SimpleChanges): void {
+    const pieceChange: SimpleChange | undefined = changes['piece']
+    if (pieceChange) {
+      const oldMediaSourceName: string | undefined = pieceChange.previousValue ? this.getPieceMediaSourceName(pieceChange.previousValue) : undefined
```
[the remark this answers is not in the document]
RasmusAlbrektsen: A good suggestion :-D I missed this

src/app/shared/services/media-state.service.ts
RasmusAlbrektsen: Good catch!

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
RasmusAlbrektsen: Good point, i did that!

src/app/shared/services/media-state.service.ts
RasmusAlbrektsen: A very good point, i removed create!

src/app/shared/services/media-state.service.ts
KvelaGorrrrnio: `FromEvent` does not refer where it is used, but defines what type of data it works with. That will naturally limit the places where it can be used. We also use this pattern e.g. with `createPartFromAction` in Sofie server. `FromAction` specifies what data the method works on, in the same manner as `FromEvent` does it.
If it should be used here, then I agree, that the duplication in `updateMedia` and `updateMediaFromEvent` can be combined, but in that case it would become:
```typescript
private updateMediaFromEvent(event: MediaCreatedEvent): void {
  this.updateMedia(event.media)
}

private updateMedia(media: Media): void {
  const mediaSubject = this.getMediaSubject(media.sourceName)
  mediaSubject.next(media)
}
```

src/app/rundown-view/components/minishelf/mini-shelf.component.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: I am fine with using a lambda – but updateMedia does use the `this` context.

KvelaGorrrrnio (approved):

src/app/shared/services/media-state.service.ts
LindvedKrvang: > That will naturally limit the places where it can be used.

I.e. referencing _where_ it can and can't be used, not _what_ the method does...

Your `updateMediaFromEvent` method is redundant. It's just a parsethrough method that take something from the parameter and parses it along.
You want to write:
```
this.subscriptions.push(this.mediaEventObserver.subscribeToMediaCreated((event: MediaCreatedEvent) => this.updateMediaFromEvent(event)))
```
When you could just write:
```
this.subscriptions.push(this.mediaEventObserver.subscribeToMediaCreated((event: MediaCreatedEvent) => this.updateMedia(event.media)))
```

LindvedKrvang (approved):

PR #103, opened 2024-02-09
Sof 1752/listen for shelf configuration changes

LindvedKrvang:
NOTICE: This PR is dependent on https://github.com/tv2/sofie-server/pull/99

This PR includes:
- The Producer shelf is generating its ActionPanels from the ShelfConfiguration
- Able to listen for Configuration events
- The Producer shelf listens for the Shelf updated event and updates its ActionPanels accordingly
- The static buttons panel in the Producer Shelf is now showing the actual Actions for "static buttons". They are hardcoded for now, since we currently don't have a way to identify them...

Note: The filtering of Actions for the ActionPanels most likely will need an update when the settings page for the ActionPanel is done.


k-primeholding-dim (approved):

src/app/shared/models/shelf-configuration.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface ShelfConfiguration {
+  id: string
+  actionPanels: ShelfActionPanelConfiguration[]
+}
+
+export interface ShelfActionPanelConfiguration {
+  name: string
+  rank: number
+  actionFilter: Tv2ActionContentType[]
```
KvelaGorrrrnio, line 11: I wouldn't know what actionFilter contained without looking at the type. It could be renamed to `allowedContentTypes`.

src/app/shared/models/shelf-configuration.ts
```diff
@@ -0,0 +1,12 @@
+import { Tv2ActionContentType } from './tv2-action'
+
+export interface ShelfConfiguration {
+  id: string
+  actionPanels: ShelfActionPanelConfiguration[]
```
KvelaGorrrrnio, line 5: For the explicitness I would prefer this to be `actionPanelConfigurations`.

src/app/shared/services/zod-configuration-parser.service.ts
```diff
@@ -0,0 +1,35 @@
...
+  })
+
+  public parseStudioConfiguration(studioConfiguration: unknown): StudioConfiguration {
+    return this.studioConfigurationParser.parse(studioConfiguration)
+  }
+
+  private readonly shelfActionPanelConfigurationParser = zod.object({
+    name: zod.string(),
+    rank: zod.number(),
+    actionFilter: zod.array(zod.nativeEnum(Tv2ActionContentType)),
```
KvelaGorrrrnio, line 24: The parser schema should be `zod.nativeEnum(Tv2ActionContentType).array()`. The current schema only validates for the tuple `[string]`.

The same goes for `actionPanels` in the `shelfConfigurationParser`.

src/app/core/models/configuration-event.ts
```diff
@@ -0,0 +1,12 @@
...
+import { ShelfConfiguration } from '../../shared/models/shelf-configuration'
+
+export interface ConfigurationEvent extends TypedEvent {
+  type: ConfigurationEventType
+}
+
+export interface ShelfConfigurationUpdatedEvent extends ConfigurationEvent {
+  type: ConfigurationEventType.SHELF_UPDATED
+  shelf: ShelfConfiguration
+}
```
KvelaGorrrrnio, line 12: The attributes here can be made readonly to enforce the functional needed for data mutations to be reactive.

src/app/shared/models/shelf-configuration.ts
LindvedKrvang: I get your point. 
The filter needs to be updated once the Settings page is done. For now, I am only filtering on ContentType, but I am not sure that is going to be enough. 
ContentType won't, for example, allows us differentiate between different kinds of Graphics.

src/app/core/services/action-trigger-event-observer.service.ts
```diff
@@ -14,7 +14,7 @@ export class ActionTriggerEventObserver {
     private readonly actionTriggerEventParser: ActionTriggerEventParser,
     logger: Logger
   ) {
-    this.logger = logger.tag('ActionTriggerEventObserver')
+    this.logger = logger.tag(ActionTriggerEventObserver.name)
```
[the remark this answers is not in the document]
LindvedKrvang: I weren't aware.
It's been updated.

src/app/shared/models/shelf-configuration.ts
LindvedKrvang: Updated

src/app/shared/services/zod-configuration-parser.service.ts
LindvedKrvang: Updated

src/app/core/models/configuration-event.ts
LindvedKrvang: Updated

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
```diff
@@ -47,14 +78,40 @@ export class ProducerShelfComponent implements OnInit, OnDestroy {
       .then(actionsObservable => actionsObservable.subscribe(this.onActionsChanged.bind(this)))
       .then(actionsSubscription => (this.actionsSubscription = actionsSubscription))
       .catch(error => this.logger.data(error).error('Failed subscribing to actions.'))
+
+    this.configurationService.getShelfConfiguration().subscribe(shelf => {
+      this.shelfConfiguration = shelf
+      this.updateActionPanels()
+    })
```
vivanov90, line 85: I try to follow a practice witch all the time we unsubscribe for the observables. I know that will emit only one time, but we still can:

```
private readonly unsubscribe$: Subject<null> = new Subject<null>()

this.configurationService.getShelfConfiguration().pipe(takeUntil(this.unsubscribe$)).subscribe(shelf => {
    this.shelfConfiguration = shelf
    this.updateActionPanels()
  })
  
   public ngOnDestroy(): void {
    this.unsubscribe$.unsubscribe()
  }
```

Is that a good practice, or is overkill?

vivanov90 (approved):
I don't see any issues, only one question, It's more for my understanding

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
LindvedKrvang: I think it is good.
It makes it so we don't need to store _n_ subscriptions and unsubscribe all of them, but instead we just store the single `unsubscribe` and subscribe on that one instead.

KvelaGorrrrnio (approved):

src/app/core/parsers/zod-configuration-event-parser.service.ts
```diff
@@ -0,0 +1,26 @@
+import { ConfigurationEventParser } from '../abstractions/configuration-event-parser'
+import { ShelfConfigurationUpdatedEvent } from '../models/configuration-event'
+import { ConfigurationParser } from '../../shared/abstractions/configuration-parser.service'
+import { ConfigurationEventType } from '../models/configuration-event-type'
+import * as zod from 'zod'
+import { Injectable } from '@angular/core'
+
+@Injectable()
+export class ZodConfigurationEventParser extends ConfigurationEventParser {
+  constructor(private readonly configurationParser: ConfigurationParser) {
```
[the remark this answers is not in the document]
LindvedKrvang: All our constructors are public unless specifically said so.
We don't add `public` to constructors. (Atleast, there currently isn't a single constructor with the `public` keyword in `ng-sofie` and `sofie-server`)

src/app/core/services/configuration-event-observer.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ShelfConfigurationUpdatedEvent } from '../models/configuration-event'
+import { ConfigurationEventType } from '../models/configuration-event-type'
+import { ConfigurationEventParser } from '../abstractions/configuration-event-parser'
+import { Logger } from '../abstractions/logger.service'
+
+@Injectable()
+export class ConfigurationEventObserver {
+  private readonly logger: Logger
+
+  constructor(
```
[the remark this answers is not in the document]
LindvedKrvang: All our constructors are public unless specifically said so.
We don't add `public` to constructors. (Atleast, there currently isn't a single constructor with the `public` keyword in `ng-sofie` and `sofie-server`)

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
```diff
@@ -47,14 +78,40 @@ export class ProducerShelfComponent implements OnInit, OnDestroy {
       .then(actionsObservable => actionsObservable.subscribe(this.onActionsChanged.bind(this)))
       .then(actionsSubscription => (this.actionsSubscription = actionsSubscription))
       .catch(error => this.logger.data(error).error('Failed subscribing to actions.'))
+
+    this.configurationService.getShelfConfiguration().subscribe(shelf => {
+      this.shelfConfiguration = shelf
```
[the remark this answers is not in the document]
LindvedKrvang: Missed one...
Good catch. It has been updated :)

PR #109, opened 2024-02-20
SOF-1800 - Rundown View Follows Cursors

RasmusAlbrektsen:
This PR makes the rundown view scroll to either the on air or next segment (if they are outside the viewport) depending on the corresponding events.

### DIFF
--- a/src/app/rundown-view/components/rundown/rundown.component.ts
+++ b/src/app/rundown-view/components/rundown/rundown.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnChanges, OnDestroy, OnInit, SimpleChanges } from '@angular/core'
+import { Component, ElementRef, Input, OnChanges, OnDestroy, OnInit, SimpleChanges, ViewChild } from '@angular/core'
 import { Rundown } from '../../../core/models/rundown'
 import { Segment } from '../../../core/models/segment'
 import { RundownTimingContextStateService } from '../../../core/services/rundown-timing-context-state.service'
@@ -12,6 +12,7 @@ import { EventSubscription } from 'src/app/event-system/abstractions/event-obser
 import { ActionStateService } from '../../../shared/services/action-state.service'
 import { Action } from '../../../shared/models/action'
 import { Tv2Action, Tv2ActionContentType, Tv2VideoClipAction } from '../../../shared/models/tv2-action'
+import { PartEvent } from '../../../core/models/rundown-event'
 
 @Component({
   selector: 'sofie-rundown',
@@ -33,6 +34,9 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
   protected miniShelfSegmentActionMappings: Record<string, Tv2VideoClipAction> = {}
   private rundownActionsSubscription: Subscription
 
+  @ViewChild('segmentList')
+  public segmentListElement: ElementRef<HTMLCanvasElement>
+
   constructor(
     private readonly rundownTimingContextStateService: RundownTimingContextStateService,
     private readonly partEntityService: PartEntityService,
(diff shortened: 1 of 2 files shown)


k-primeholding-dim (approved):

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -72,6 +78,24 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
...
+      return
+    }
+
+    document.getElementById(event.segmentId)?.scrollIntoView({
+      behavior: 'smooth',
+      block: 'nearest',
+      inline: 'nearest',
+    })
+  }
+
```
vivanov90, line 93: You got `document.getElementById(event.segmentId)` in constant on to the top, why you don't use it here?
```
 document.getElementById(event.segmentId)?.scrollIntoView({
      behavior: 'smooth',
      block: 'nearest',
      inline: 'nearest',
    })
```
In addition you check you have a element and is no need to be optional  document.getElementById(event.segmentId)**?**.scrollIntoView

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -72,6 +78,24 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
...
+  public isInsideViewport(element: HTMLElement): boolean {
+    const elementRect: DOMRect = element.getBoundingClientRect()
+    return elementRect.top >= this.segmentListElement.nativeElement.clientHeight && elementRect.bottom <= (window.innerHeight || document.documentElement.clientHeight)
```
vivanov90, line 96: it's most likely a problem in me, but I personally don't understand what and why we're checking here

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -49,14 +53,16 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
...
   }
 
-  private subscribeForEventObserver(): void {
+  private subscribeToEventObserver(): void {
     this.rundownEventSubscriptions.push(this.rundownEventObserver.subscribeToRundownAutoNext(this.setAutoNextStartedToTrue.bind(this)))
     this.rundownEventSubscriptions.push(this.rundownEventObserver.subscribeToRundownSetNext(this.setAutoNextStartedToFalse.bind(this)))
     this.rundownEventSubscriptions.push(this.rundownEventObserver.subscribeToRundownReset(this.setAutoNextStartedToFalse.bind(this)))
     this.rundownEventSubscriptions.push(this.rundownEventObserver.subscribeToRundownDeactivation(this.setAutoNextStartedToFalse.bind(this)))
+    this.rundownEventSubscriptions.push(this.rundownEventObserver.subscribeToRundownTake(this.scrollViewToSegment.bind(this)))
+    this.rundownEventSubscriptions.push(this.rundownEventObserver.subscribeToRundownSetNext(this.scrollViewToSegment.bind(this)))
```
LindvedKrvang, line 65: This component already subscribes to `RundownSetNext` on line 61. No need to make two subscriptions for the same thing.
A potential way of doing it:
```
this.rundownEventSubscriptions.push(this.rundownEventObserver.subscribeToRundownSetNext(event => {
  this.setAutoNextStartedToFalse(event)
  this.scrollViewToSegment(event)
}))
```

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -72,6 +78,24 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
       .catch(error => this.logger.data(error).error('Failed subscribing to rundown actions changes.'))
   }
 
+  public scrollViewToSegment(event: PartEvent): void {
+    const element: HTMLElement | null = document.getElementById(event.segmentId)
+    if (!element || this.isInsideViewport(element)) {
+      return
+    }
+
+    document.getElementById(event.segmentId)?.scrollIntoView({
```
LindvedKrvang, line 87: You already fetch the `element` on line 82. No need to fetch it again

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -72,6 +78,24 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
...
+      behavior: 'smooth',
+      block: 'nearest',
+      inline: 'nearest',
```
LindvedKrvang, line 90: These are "almost" magic strings.
It would be nice if we could extract this block to something like:
```
const SCROLL_CONFIGURATION: ScrollIntoViewOptions = {
  behavior: 'smooth',
  block: 'nearest',
  inline: 'nearest',
}
```
so you can use it like:
`element.scrollIntoView(SCROLL_CONFIGURATION)`

LindvedKrvang (approved):

src/app/rundown-view/components/rundown/rundown.component.ts
RasmusAlbrektsen: A good point, i don't know why i didn't change that :-D 
I have changed it to use the element instead!

src/app/rundown-view/components/rundown/rundown.component.ts
RasmusAlbrektsen: We don't want to scroll the view unless the element is outside of what the producer is actually looking at it. The `segmentListElement` is used because we don't want to consider the height that goes behind the header, since that isn't actually visible to the producer.

src/app/rundown-view/components/rundown/rundown.component.ts
RasmusAlbrektsen: Changed to the suggestion, good point!

src/app/rundown-view/components/rundown/rundown.component.ts
RasmusAlbrektsen: Good point! Changed!

src/app/rundown-view/components/rundown/rundown.component.ts
RasmusAlbrektsen: It has been moved to a constant, i called it `SMOOTH_SCROLL_CONFIGURATION` to make it specific as to what the configuration is specifically for.

LindvedKrvang (approved):

PR #113, opened 2024-02-23
Sof 1560/update snackbar notification popup

LindvedKrvang:
Our Snackbar has been replaced by a NotificationService.
When called, this service will notify all the NotificationContainers (useally one per page, but there can be as many as your heart desire) and they will create a Notification that will popup where the container is placed.

The NotificationService is able to be called from anywhere in the system.
To use the NotificationService, simply inject it and call one of it functions; Like so:
```
constructor(private notificationService: NotificationService) {
  this.notificationService.createInfoNotification('write-your-message-here')
}
``` 

The introduction of the NotificationServcie means that all references to MatSnackbar has been removed.

Demo:

https://github.com/tv2/ng-sofie/assets/23049647/ff8eb3c3-5adb-47f8-8dc1-f1592e953c13

### DIFF
--- a/src/app/shared/components/header/header.component.ts
+++ b/src/app/shared/components/header/header.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit } from '@angular/core'
+import { Component, ElementRef, OnInit } from '@angular/core'
 import { Paths } from '../../../app-routing.module'
 import { Router } from '@angular/router'
 import { Logger } from '../../../core/abstractions/logger.service'
@@ -11,14 +11,18 @@ import { Title } from '@angular/platform-browser'
 })
 export class HeaderComponent implements OnInit {
   public title: string
+  public elementRef: ElementRef
+
   private readonly logger: Logger
 
   constructor(
+    private readonly elementReference: ElementRef,
     private readonly router: Router,
     private readonly titleService: Title,
     logger: Logger
   ) {
     this.logger = logger.tag('HeaderComponent')
+    this.elementRef = elementReference
   }
 
   public ngOnInit(): void {
--- a/src/app/shared/components/header/header.component.ts
+++ b/src/app/shared/components/header/header.component.ts
@@ -16,7 +16,7 @@ export class HeaderComponent implements OnInit {
   private readonly logger: Logger
 
   constructor(
-    private readonly elementReference: ElementRef,
+    elementReference: ElementRef,
     private readonly router: Router,
     private readonly titleService: Title,
     logger: Logger
--- a/src/app/shared/components/notification-container/notification-container.component.ts
+++ b/src/app/shared/components/notification-container/notification-container.component.ts
@@ -1,0 +1,101 @@
+import { AfterViewInit, ApplicationRef, Component, ComponentRef, createComponent, ElementRef, EmbeddedViewRef, Input, OnDestroy, OnInit } from '@angular/core'
+import { NotificationService } from '../../services/notification.service'
+import { Subject, takeUntil } from 'rxjs'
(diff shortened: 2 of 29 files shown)


k-primeholding-dim (approved):

src/app/shared/components/notification/notification.component.scss
```diff
@@ -0,0 +1,41 @@
+$animation-duration: 0.3s;
+$animation-function: ease;
+$animation-fill-mode: forwards;
+
+:host {
+  --gradient-degree: 90deg;
+  --status-color-width: 5%;
```
vivanov90, line 7: Do you have any specific reason to mix this this approach with create css variables?

src/app/shared/components/notification/notification.component.ts
```diff
@@ -0,0 +1,56 @@
...
+        return WARNING_COLOR
+      }
+      case StatusCode.BAD: {
+        return ERROR_COLOR
+      }
+      default: {
+        return INFO_COLOR
+      }
+    }
+  }
```
vivanov90, line 55: I will use getter here:
```
private get statusColor(): string {
    switch (this.notification.statusCode) {
      case StatusCode.WARNING: {
        return WARNING_COLOR
      }
      case StatusCode.BAD: {
        return ERROR_COLOR
      }
      default: {
        return INFO_COLOR
      }
    }
  }
```
and we can:
```
private setBackground(): void {
    this.elementRef.nativeElement.style.background = `linear-gradient(${GRADIENT_DEGREE}, ${this.statusColor}, ${this.statusColor} ${STATUS_COLOR_WIDTH}, ${BACKGROUND_COLOR} ${STATUS_COLOR_WIDTH}, ${BACKGROUND_COLOR})`
  }
```

vivanov90 (commented):
Before approve I wanna checkout the branch and test the notification, because I want to see how notification positioning working

LindvedKrvang:
Closing this PR because I made too many drastic changes in https://github.com/tv2/ng-sofie/pull/114.
I will address all comments on this PR in that one.

src/app/shared/components/header/header.component.ts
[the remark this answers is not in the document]
LindvedKrvang: I'm not aware of it is best practice or not.
It would be nice not to have and get the ElementRef.
How would you go about getting the `offsetHeight` without it?

src/app/shared/components/notification-container/notification-container.component.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  constructor(
+    private readonly elementRef: ElementRef,
+    private readonly applicationRef: ApplicationRef,
+    private readonly notificationService: NotificationService
+  ) {}
+
+  public ngAfterViewInit(): void {
+    this.elementRef.nativeElement.style.top = TOP_PADDING + this.topOffset + PIXEL_POSTFIX
+    this.elementRef.nativeElement.style.right = this.rightOffset + PIXEL_POSTFIX
```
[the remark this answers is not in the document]
LindvedKrvang: Personally not a fan. Mostly because then `px` becomes a magic string for all usages of it which in this file is already at four.

src/app/shared/components/notification/notification.component.scss
LindvedKrvang: I am not sure what you are referring to when you say "create" CSS variables. Could you elaborate?

src/app/shared/components/notification/notification.component.ts
LindvedKrvang: This comes down to preference. We haven't really been utilizing getters that much.

PR #115, opened 2024-02-27
Sof 1748/create page for shelf configuration

vivanov90:
- Create page for shelf configuration
- Redesign Keyboard mappings and Clear cache pages - just to look nice we need to make full redesign in this pages
- Add sofie-utility-classes.scss - to add some general spacing and display classes
- Add new components for table-headers, dialog and multi-select input.


RasmusAlbrektsen:
It looks good and works so well, good stuff! 
Before i go into the code i have some things that are easier to explain in terms of look and feel, rather than taking examples from the code. 

- The table headers such as `Panel name` and `Rank`, shouldn't be bold in my opinion. We should have a more faded gray color on those, much like the mockup. 

- The "Filtering-arrow" should be shown when hovering the related table header, so that i can that i have the option rather than not being there until i click it. 

- I wonder how it would look if we didn't extend the table all the way to the right and just keep it at e.g. a certain width or percentage of the screen since we stray far from looking at the name to then looking at where i edit or use the buttons.

vivanov90:
> It looks good and works so well, good stuff! Before i go into the code i have some things that are easier to explain in terms of look and feel, rather than taking examples from the code.
> 
> * The table headers such as `Panel name` and `Rank`, shouldn't be bold in my opinion. We should have a more faded gray color on those, much like the mockup.
> * The "Filtering-arrow" should be shown when hovering the related table header, so that i can that i have the option rather than not being there until i click it.
> * I wonder how it would look if we didn't extend the table all the way to the right and just keep it at e.g. a certain width or percentage of the screen since we stray far from looking at the name to then looking at where i edit or use the buttons.

@RasmusAlbrektsen Points 1 and 2 are done. For 3 I will add some screen shots, looks bad, I don't like it, but maybe I don't understand it what you want to do exactly.
Screenshot 2024-02-28 104258
Screenshot 2024-02-28 104323
Screenshot 2024-02-28 104353

k-primeholding-dim (approved):

LindvedKrvang:
> For 3 I will add some screen shots, looks bad,

It looks bad/off because the background is too dark. You are currently using the `--grey-color` variable. It looks a lot better with a slightly less dark background.
Here it is with the `--light-grey-color`, which looks quite nice:
image

The side menu also needs some work.
Right now, it takes way too much attentention with the same dark color as the header and it taking up the entire width of the side menu.
Adding some padding and giving it a ligther color is going to help this tremendiously:
image

Then the entire view looks like this, which looks really nice:
image

The only comment I have on that is the big blue "bootstraps" buttons that are extremely out of place in this UI. Both in color and shape.

src/app/settings/components/settings/settings-menu/settings-menu.component.scss
```diff
@@ -18,10 +17,36 @@
       font-size: medium;
       display: block;
       text-decoration: none;
-      color: var(--primary-color);
+      color: var(--black-color);
       &.active-tab {
+        color: var(--white-color);
         background-color: var(--background-color);
```
LindvedKrvang, line 23: This color is way too dark for the selected settings menu item. You should look at the mockup for how it should look

src/app/settings/components/settings/settings-menu/settings-menu.component.scss
```diff
@@ -18,10 +17,36 @@
...
+      padding: 12px 16px 12px 30px;
+      font-size: medium;
+      text-decoration: none;
+      color: var(--black-color);
+      &.active-tab {
+        font-weight: 400;
+        text-decoration: underline;
+        &::before {
+          content: ' \203A';
+          position: absolute;
```
LindvedKrvang, line 44: What are using position absolute and relative for here? Seems unnecessary

src/app/settings/components/settings/settings-menu/settings-menu.component.scss
```diff
@@ -18,10 +17,36 @@
...
+  }
+
+  &__nav-submenu {
+    width: 100%;
+    font-weight: lighter;
+    padding-top: 0;
+    a {
+      position: relative;
+      display: block;
+      padding: 12px 16px 12px 30px;
```
LindvedKrvang, line 35: You have introduced CSS classes for padding. You should use those. Goes for the `padding-top` four lines above here as well.

src/app/settings/components/settings/settings-menu/settings-menu.component.ts
```diff
@@ -1,11 +1,33 @@
...
   templateUrl: './settings-menu.component.html',
   styleUrls: ['./settings-menu.component.scss'],
 })
-export class SettingsMenuComponent {
+export class SettingsMenuComponent implements OnInit, OnDestroy {
   public settingsPaths = SettingsPaths
+  public shelfPaths = ShelfPaths
+  public isShelfRouteActive: boolean = false
+
+  private readonly unsubscribe$: Subject<void> = new Subject<void>()
```
LindvedKrvang, line 17: Minor thing: You don't need to declare the diamon brackets on the right side of equal sign because you have defined on the left side.

src/app/settings/components/settings/settings-menu/settings-menu.component.ts
```diff
@@ -1,11 +1,33 @@
...
 
 @Component({
   selector: 'sofie-settings-menu',
   templateUrl: './settings-menu.component.html',
   styleUrls: ['./settings-menu.component.scss'],
 })
-export class SettingsMenuComponent {
+export class SettingsMenuComponent implements OnInit, OnDestroy {
   public settingsPaths = SettingsPaths
+  public shelfPaths = ShelfPaths
```
LindvedKrvang, line 14: These are enums you are exposing to the HTML. Be nice if those were PascalCased i.e. `public SettingsPaths = SettingsPaths`.

src/app/settings/components/settings/settings-menu/settings-menu.component.ts
```diff
@@ -1,11 +1,33 @@
...
+  public isShelfRouteActive: boolean = false
+
+  private readonly unsubscribe$: Subject<void> = new Subject<void>()
+
+  constructor(private readonly router: Router) {}
+
+  public ngOnInit(): void {
+    this.isShelfRouteActive = this.router.url.includes(this.settingsPaths.SHELF)
```
LindvedKrvang, line 22: You are using enums here, so just use them directly to make it more clear that you are using enums:
`this.isShelfRouteActive = this.router.url.includes(SettingsPaths.SHELF)`

src/app/settings/components/settings/settings.component.scss
```diff
@@ -4,6 +4,7 @@
   display: flex;
   &__content {
     flex-grow: 1;
+    background-color: var(--gray-color);
```
LindvedKrvang, line 7: This background color is too dark. Look at the mockup to see how it should look like

src/app/settings/shelf/action-panel/action-panel-import/action-panel-import.component.ts
```diff
@@ -0,0 +1,69 @@
...
+import { ConfigurationService } from 'src/app/shared/services/configuration.service'
+import { ConfigurationParser } from 'src/app/shared/abstractions/configuration-parser.service'
+import { ShelfActionPanelConfiguration, ShelfActionPanelConfigurationWithId, ShelfConfiguration } from 'src/app/shared/models/shelf-configuration'
+
+@Component({
+  selector: 'sofie-action-panel-import',
+  templateUrl: './action-panel-import.component.html',
+  styleUrls: ['./action-panel-import.component.scss'],
+})
+export class ActionPanelImportComponent {
```
LindvedKrvang, line 12: This component duplicates a lot of the logic of `ActionTriggersImportComponent`.

Instead of duplicating the code we should have an import sevice that can load a text file and convert it to an object of a given type. Then these components can use that service for the actual logic of reading an parsing a file.

src/app/settings/shelf/action-panel/action-panel-list/action-panel-list.component.html
```diff
@@ -0,0 +1,90 @@
+<div class="c-action-panel-list">
```
LindvedKrvang, line 1: This components does a lof of the stuff that is supposed to be encapsulated by a reusable TableComponent.
Instead of making a reuseable component, you have made global CSS classes. We don't want global CSS classes, we want reusable components.

If we need to use another table somewhere else in the system, we need to reimplemt a lot of the markup/logic for a table. That goes against the point of reusable components.

LindvedKrvang (changes requested):
I have only made it through 1/3 of the PR, but I'm running out of time to review more today, so here are my comments so far.

Two things seems to stand out as a trend:
1. You are making global CSS classes instead of reusable components. We want to limit the use of global CSS classes because a global CSS class still needs to have all its markup and logic reimplemented everytime we need to use it. Instead, all of this should be encapsulated in reusable components, so we only have to write it once.
2. You are placing a lot of styling for a component outside of the component that needs it. I.e. it is the parent component that styles component.

I think we maybe should have a session together on what makes a component reusable and how to make them.

src/scss/_sofie-utility-classes.scss
```diff
@@ -0,0 +1,251 @@
...
+
+.ms-4 {
+  margin-left: 20px;
+}
+
+.ms-5 {
+  margin-left: 30px;
+}
+
+.me-0 {
```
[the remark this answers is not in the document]
vivanov90: ms- margin-start, me- margin-end. This is a new trend after the introduction of flex box from where we have flex-end and flex-start. Bootstrap and now I see places in Angular Material use that naming conventions. I think it was introduced because of possible element rotation and then left is no longer left and right is right

src/app/core/services/shelf-configuration-state.service.ts
```diff
@@ -0,0 +1,60 @@
...
+import { ConnectionStatusObserver } from './connection-status-observer.service'
+import { EventSubscription } from '../../event-system/abstractions/event-observer.service'
+import { MatSnackBar } from '@angular/material/snack-bar'
+import { ConfigurationEventObserver } from './configuration-event-observer'
+import { ShelfConfigurationUpdatedEvent } from '../models/configuration-event'
+import { ShelfActionPanelConfiguration, ShelfConfiguration } from 'src/app/shared/models/shelf-configuration'
+import { ConfigurationService } from 'src/app/shared/services/configuration.service'
+
+@Injectable()
+export class ShelfConfigurationStateService {
```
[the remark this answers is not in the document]
vivanov90: I use it because of the data centralization logic and I don't need to resync the data after adding/importing (or provide the data from child to parent component) or other operations. I consider it a good practice and a fairly familiar practice by all developers in the project. In addition, I do not know where it is expected to be used and how exactly the information is stored here.

src/scss/_sofie-utility-classes.scss
```diff
@@ -0,0 +1,251 @@
+.m-0 {
```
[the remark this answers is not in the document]
vivanov90: I don't mind, as it would be convenient I will write them like this, for now I added it here as a test concept. The names themselves come from bootstrap 4

src/scss/_sofie-utility-classes.scss
```diff
@@ -0,0 +1,251 @@
...
+  display: flex !important;
+}
+
+.d-block {
+  display: block !important;
+}
+
+.hidden {
+  display: none !important;
+}
```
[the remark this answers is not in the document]
vivanov90: Yes because most of the times you want to add this class from the script, but if you already add some display functionality in to the local scss file that will be more important from this generic class. For that reason I add "!important". We can remove it, but all develops must know this side effect and set the display value only from this generic classes.

src/scss/_sofie-table.scss
```diff
@@ -0,0 +1,70 @@
+.c-sofie-table {
```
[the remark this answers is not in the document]
vivanov90: Because I'm not quite sure how to create a generic table component without saying that the data this component will receive will be array<any>. Until now, when I have built such components, I give the table keys and the order of the columns as an object set from the parent component. Similar to the logic I have in sofie-table-header. However, this cannot be done without providing a table-data sheet which must be an array<any> as far as the sofie-table component is concerned.

src/app/settings/action-triggers/components/action-triggers-list/action-triggers-list.component.html
```diff
@@ -9,7 +9,7 @@
         (onSelect)="applyUserActionToSelectedActionTriggers($event)"
         label="Selected">
       </sofie-dropdown-button>
-      <sofie-button (onClick)="setSelectMode(false)" label="Disable multiselect"></sofie-button>
+      <sofie-button (onClick)="setSelectMode(false)" label="Disable multiselect" [buttonClasses]="'c-sofie-button c-sofie-button-primary ms-2'"></sofie-button>
```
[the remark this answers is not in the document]
vivanov90: 1. When we add class we add this class to native angular elements not to the button witch will be content of the container. I Want to apply it on <button> not on <sofie-button>. We can change the scss selectors and work with class applied to <sofie-button>. In reality the different is verry small
2 and 3. If it's used only there I will move in sofie-button scss

src/app/settings/action-triggers/components/action-triggers-list/action-triggers-list.component.scss
```diff
@@ -2,17 +2,18 @@
   width: 100%;
   min-width: 300px;
   &__search {
-    margin-bottom: 24px;
+    margin: 24px 30px;
```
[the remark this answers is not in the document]
vivanov90: This is true and if the concept is improved we will make a separate task to pass all the offsets in the project to use this offset. I was not sure we want to go to this route and for that reason don't spend a lot of time to move old spacings to this concept.

src/app/shared/models/shelf-configuration.ts
```diff
@@ -1,12 +1,16 @@
...
 
 export interface ShelfActionPanelConfiguration {
   name: string
   rank: number
   actionFilter: Tv2ActionContentType[]
 }
+
+export interface ShelfActionPanelConfigurationWithId extends ShelfActionPanelConfiguration {
+  id: string
+}
```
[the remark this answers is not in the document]
vivanov90: I didn't find any other way to save the selected components after reordering the table if I work with table index. I will check below what your suggestion is.

src/scss/_sofie-table.scss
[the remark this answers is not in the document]
vivanov90: I will remove c- while I am now rewriting the code as per the comments left here

src/app/settings/components/settings/settings-menu/settings-menu.component.ts
RasmusAlbrektsen: Also do we want to use the`unsubscribe$` naming convention? If so we haven't been doing that in other files, but i see that we have it in a couple now.

src/app/settings/action-triggers/components/action-triggers-list/action-triggers-list.component.html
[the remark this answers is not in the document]
vivanov90: @LindvedKrvang 2 and 3 are done

src/app/settings/action-triggers/components/action-triggers/action-triggers.component.html
```diff
@@ -1,17 +1,18 @@
-<div class="c-action-triggers c-settings-page">
-  <div class="c-settings-page__header">
-    <h2 i18n>settings.keyboard-mappings.label</h2>
-    <div>
-      <button class="c-sofie-button c-sofie-button-primary" (click)="createActionTrigger()" i18n>action-triggers.create-shortcut.button</button>
-      <button class="c-sofie-button c-sofie-button-primary" (click)="exportActionsTriggers()" i18n>action-triggers.export-shortcuts.button</button>
-      <sofie-action-triggers-import [actionTriggersWithAction]="actionTriggersWithAction" [isDisabled]="isLoading"></sofie-action-triggers-import>
-    </div>
+<div class="c-action-triggers-actions">
```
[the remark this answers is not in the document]
vivanov90: I think we can make that change when we redesign Keyboard Mappings components

src/app/settings/components/settings/settings-menu/settings-menu.component.html
```diff
@@ -5,4 +5,12 @@
   <div class="c-side-bar__nav-item">
```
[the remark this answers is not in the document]
vivanov90: We need to discuss that because in my understanding similar nesting and more correct for more straight file system structure.

src/app/settings/action-triggers/components/action-triggers/action-triggers.component.html
```diff
@@ -1,17 +1,18 @@
...
+    <button class="c-sofie-button c-sofie-button-primary me-2" (click)="createActionTrigger()" i18n>action-triggers.create-shortcut.button</button>
+    <button class="c-sofie-button c-sofie-button-primary me-2" (click)="exportActionsTriggers()" i18n>action-triggers.export-shortcuts.button</button>
+    <sofie-action-triggers-import [actionTriggersWithAction]="actionTriggersWithAction" [isDisabled]="isLoading"></sofie-action-triggers-import>
   </div>
+</div>
 
+<div class="c-action-triggers c-settings-page">
   <ng-container *ngIf="!isLoading; else loader">
     <div class="c-action-triggers__body">
       <div class="c-action-triggers__body--list" [ngClass]="{ 'c-action-triggers__body--list-open': selectedActionTrigger || isShowingCreateActionTriggerForm }">
```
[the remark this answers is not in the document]
vivanov90: In general, I understand where the problem is, but in the specific case it is logic whether to show the edit action trigger at all. In addition that will disappear after action trigger redesign and is good to not waste too much time rewriting this logic

src/app/settings/action-triggers/components/edit-action-triggers/edit-action-triggers.component.ts
```diff
@@ -25,7 +25,7 @@ export class EditActionTriggersComponent implements OnChanges {
   public readonly physicalMappingHelpInfo = $localize`action-triggers.physical-mapping-help.tooltip`
   public readonly triggerOnLabel = $localize`action-triggers.trigger-on.label`
   public readonly selectedActionLabel = $localize`action-triggers.selected-action.label`
-  public readonly submitButtonTooltipError = $localize`action-triggers.submit-tooltip.error`
+  public readonly submitButtonTooltipError = $localize`global.submit-tooltip.error`
```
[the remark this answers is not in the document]
vivanov90: I name global. information witch I expect to be use in more components not in specific one. That `global.submit-tooltip.error` is used in action-trigger and action-panel submit.

src/app/settings/components/settings/settings-menu/settings-menu.component.ts
vivanov90: Ok, I will remove it.
@RasmusAlbrektsen unsubscribe$ name is really unique and we will change verry easy when we decide the naming

src/app/settings/components/settings/settings-menu/settings-menu.component.scss
vivanov90: Not really it's remove it.

src/app/shared/models/shelf-configuration.ts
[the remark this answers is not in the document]
LindvedKrvang: In `action-panel-list.component.ts` you have the variable `selectedActionPanelIds: Set<string>`. 
This could simply be: `selectedActionPanels: Set<ShelfActionPanelConfiguration>` which makes the id redundant.

You need to get out of the mindset of sending the least amount of data around as possible. We are writing object-oriented programming, so we should utilize/take advantage of that and send the objects around.

But, my main point here _isn't_ that you added the id. The id might very well be needed/a good idea to add.

My main point here is that you created a separate interface _and_ made a previous interface generic _just to add a single attribute_.
This is what the code should look like after adding the id:
```
export interface ShelfConfiguration {
  id: string
  actionPanelConfigurations: ShelfActionPanelConfiguration[]
}

export interface ShelfActionPanelConfiguration {
  id: string // <-- This is all you had to add
  name: string
  rank: number
  actionFilter: Tv2ActionContentType[]
}
```

src/app/core/services/shelf-configuration-state.service.ts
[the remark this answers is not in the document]
LindvedKrvang: StateServices is a practice to be used when we have data that needs to go across multiple components/services that or to prevent duplicating complex logic maintaining state.

This stateService does neither.

What this stateService currently does falls under the responsibility of the "mother/smart" component in a "mother/child" (or smart/dumb) component hierarchy which is what we want in ngSofie.

This stateService currently only fetches the ShelfConfigurations and nothing else and it is only used by _a single place_ in the code (the `action-panel.component.ts`). The ActionPanelComponent should simply inject the `ConfigurationService` directly itself instead of the stateService.

Introducing a stateService introduces an extra layer of complexity. We need to have good arguments for introducing complexity. The arguments: "We do this elsewhere in the code" or "People know how to use this" are not good enough arguments.

src/scss/_sofie-utility-classes.scss
```diff
@@ -0,0 +1,251 @@
...
+
+.m-4 {
+  margin: 20px;
+}
+
+.m-5 {
+  margin: 30px;
+}
+
+.ms-0 {
```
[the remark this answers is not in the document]
LindvedKrvang: Ahh, that makes sense :)

src/app/settings/action-triggers/components/action-triggers-list/action-triggers-list.component.html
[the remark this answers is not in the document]
LindvedKrvang: > I Want to apply it on not on

I don't follow what you mean here.

1 & 2: You are still parsing the CSS class from the calling component to the SofieButton component each and every time you use the SofieButton. There is currently not a single usage of SofieButton where `c-sofie-button` isn't also been given to it. Looking at the CSS for `c-sofie-button` it looks like that this class is quite essential for the button to look correctly. If that's true, then we shouldn't need to remember to give the component this class every time we use it, it should simply always have the class.

So instead of writing: `<sofie-button buttonClasses="c-sofie-button" ... ></sofie-button>` you should just write `<sofie-button ... ></sofie-button>`

Speaking of `buttonClasses`. This attribute needs to be removed. It's used to parse CSS classes from the outside of the component to the button component (which we established above we shouldn't). In the Dialog component you are parsing `buttonClasses` throught _at least three_ components, just to give SofieButton a CSS class that it should already be setting itself.

3. You are still parsing it as a CSS class instead of a type exposed by on SofieButton:
```
# sofie-button.ts
@input() public type: 'PRIMARY' | 'WARNING' | 'DANGER' = 'PRIMARY'

#some component that wants a primary button
<sofie-button></sofie-button>

#some component that wants a warning button
<sofie-button type="WARNING"></sofie-button>

#some component that wants a danger button
<sofie-button type="DANGER"></sofie-button>

```
Above solution is typesafe and all calling components don't need to know an intimate detail of SofieButton (the CSS class). Instead they can just see the exposed types available for the button.

-----

Formulating the above answer made me take a closer look at the implementation of `SofieButton` which lead me to some comments about the SofieButton: 

The attribute `onClick: EventEmitter` is redundant and should be removed. All it does is letting us write `<sofie-button (onClick)="..."></sofie-button>`. On HTML elements there is already a `(click)` event and we should just use that. So it becomes: `<sofie-button (click)="..." ></sofie-button>`.

The attribute `label` should ideally be removed. Right now, SofieButton can only ever be a "text" button. Instead of having a label attribute, we should just content inject whatever needs to be in the button:
So instead of writing `<sofie-button label="Click me!"></sofie-button>` we should write `<sofie-button>Click me!</sofie-button>`. This makes the component a lot more flexible.

The attribute `buttonClasses` needs to be removed as discussed further above.

The attribute `type` needs to be more typesafe. Right now it is `public type: string = 'button'`. You use it to indicate if it is a submit button or not, so it needs to atleast be: `public type: 'button' | 'submit' = 'button'`. 
I would however make the argument that this attribute also should be deleted.

src/app/settings/action-triggers/components/action-triggers-list/action-triggers-list.component.scss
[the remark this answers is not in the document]
LindvedKrvang: The purpose of having helper classes for margins and paddings is defated if we just set margins and paddings without the helpers anyway

src/app/settings/components/settings/settings-menu/settings-menu.component.html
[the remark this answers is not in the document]
LindvedKrvang: We are currently applying a file/folder structure for our components that says that a module have a `components` folder. In this folder there are one folder for each component. These folders should only contain the `.ts`, `.html`, `.scss` and `.spec.ts` for that specific component. It should _not_ contain other components.

We can take it up for discussion if you want, but untill a different decision has been made, this is the file structure that you should be following.

src/app/settings/components/settings/settings-menu/settings-menu.component.ts
LindvedKrvang: @RasmusAlbrektsen are you reffering to the `unsubscribe` part or the `$` part?
Personnally, I think the name `unsubscribe` is a quite fitting name.

The `$` is a common convention in Angular to indicate the variable is a Subject/Observable. Personnally, I would prefer something like `unbsubscribeSubject` over `unsubscribe$`.

src/app/shared/components/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  public onOptionClick(option: MultiSelectOptions): void {
+    const selectedOptionIndex: number = this.selectedOptionsIds.findIndex(selectedOptionId => selectedOptionId === option.id)
+    if (selectedOptionIndex !== -1) {
+      this.unselectOption(selectedOptionIndex)
+    } else {
+      this.onItemSelectionChange.emit([...this.selectedOptionsIds, option.id])
+    }
+  }
+
+  public unselectOption(optionIndex: number): void {
```
RasmusAlbrektsen, line 45: This is very nerdy and i'm probably the only one who cares, so no hard feelings if this is how we do it. 

"Unselect", i believe, generally applies to things that has yet to be selected, where "deselect" refers to removing actual selections.

RasmusAlbrektsen (commented):
I took a look and made some initial comments :-)

src/app/shared/components/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  selector: 'sofie-multi-select',
+  templateUrl: './multi-select.component.html',
+  styleUrls: ['./multi-select.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MultiSelectComponent {
+  @Input() public options: MultiSelectOptions[]
+  @Input() public selectedOptionsIds: string[]
+  @Input() public placeholder: string
+  @Input() public showOnlyItemsCountLabel?: string
```
LindvedKrvang, line 21: What does this variable do?
The name indicates that it is a boolean, but it is a string??

src/app/shared/components/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,49 @@
...
+
+@Component({
+  selector: 'sofie-multi-select',
+  templateUrl: './multi-select.component.html',
+  styleUrls: ['./multi-select.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MultiSelectComponent {
+  @Input() public options: MultiSelectOptions[]
+  @Input() public selectedOptionsIds: string[]
```
LindvedKrvang, line 19: This makes so it is not the MultiSelect itself that keeps tracks of what is selected. That seems to be a big responsibility of the MultiSelect that you have removed from it. 
Why is it not the MultiSelect that keeps track of what the MultiSelect has selected?

src/app/shared/components/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  @Input() public placeholder: string
+  @Input() public showOnlyItemsCountLabel?: string
+
+  @Output() public readonly onItemSelectionChange: EventEmitter<string[]> = new EventEmitter<string[]>()
+
+  public readonly IconButton = IconButton
+  public readonly IconButtonSize = IconButtonSize
+
+  public getOptionById(optionId: string): MultiSelectOptions | undefined {
```
LindvedKrvang, line 28: This method is only required because you chose to let the responsibility of keeping track of the selected options _outside_ of the MultiSelect component _and_ because you are only giving the MultiSelect components the ids of the select options instead of giving them the entire option. We are writing object-oriented programming and we should utilize it.

src/scss/_sofie-dialog.scss
```diff
@@ -0,0 +1,36 @@
+$actions-size-height: 80px;
```
LindvedKrvang, line 1: Why is this a global CSS file instead of being styling encapsulated by the DialogComponent?

src/scss/_sofie-forms.scss
```diff
@@ -72,3 +71,12 @@ select {
     right: 10px;
   }
 }
+
+.c-multi-select-menu {
```
LindvedKrvang, line 75: Why is this a global CSS class instead of being encapsulated in the MultiSelect component?

src/scss/_sofie-dialog.scss
LindvedKrvang: Same question for having a global CSS file for Forms?

src/scss/_sofie-page.scss
```diff
@@ -53,28 +23,67 @@
...
+
+.c-settings-title {
+  margin: 30px;
+  font-size: 2rem;
+  font-weight: lighter;
 }
 
 .c-settings-page {
-  padding: 32px;
+  padding: 0;
```
LindvedKrvang, line 51: As mentioned in a previous comment, we should have reuseable components for the scafolding of our Settings pages instead of having global CSS classes. 
We should aim to limit our use of global CSS classes as much as possible.

src/scss/_sofie-utility-classes.scss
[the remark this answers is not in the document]
LindvedKrvang: We should use `xs`, `s` etc instead of using `0`, `1` etc

src/app/shared/pipes/translation-known-values.pipe.ts
```diff
@@ -0,0 +1,34 @@
+import { Pipe, PipeTransform } from '@angular/core'
+import { Tv2ActionContentType } from '../models/tv2-action'
+
+@Pipe({
+  name: 'translationKnownValues',
+})
+export class TranslationKnownValuesPipe implements PipeTransform {
```
LindvedKrvang, line 7: What constitues a `known value`? That's an extremely broad definition.
This pipe translate `Tv2ActionContentType` so a more accurate name would be `TranslateContentType`.

src/app/shared/pipes/translation-known-values.pipe.ts
```diff
@@ -0,0 +1,34 @@
...
+  public transform(value: string): string {
```
LindvedKrvang, line 8: Since we know that we are translating `Tv2ActionContentType` then the parameter should be `contentType: Tv2ActionContentType`

KvelaGorrrrnio:
Superseded by #121.

PR #117, opened 2024-02-28
SOF-1732 - fix open a link in a new tab/window with either context me…

vivanov90:
Fix links to use routerLink logic and a tags


k-primeholding-dim (approved):

src/app/shared/directives/sofie-link.directive.ts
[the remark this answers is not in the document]
LindvedKrvang: I agree with Anders. No need to reinvent something that is supported already by Angular.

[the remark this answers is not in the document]
vivanov90: I'm not sure why it didn't work for me on the first test and I went down a rabbit hole from there. I deleted the directive and fixed the rest of the components, please look at it again when you have time.

src/app/shared/components/header/header.component.html
```diff
@@ -1,15 +1,17 @@
...
+      <sofie-logo [text]="title"></sofie-logo>
+    </a>
   </mat-toolbar-row>
 
   <mat-toolbar-row class="c-header__main-menu">
     <mat-nav-list>
-      <a class="c-header__nav-item" (click)="navigateToRundown()" i18n>header-component.rundowns.label</a>
-      <a class="c-header__nav-item" (click)="navigateToStatus()" i18n>header-component.status.label</a>
+      <a class="c-header__nav-item" [routerLink]="['/', Paths.RUNDOWNS]" i18n>header-component.rundowns.label</a>
+      <a class="c-header__nav-item" [routerLink]="['/', Paths.STATUS]" i18n>header-component.status.label</a>
```
KvelaGorrrrnio, line 11: I can see that in the settings page, we use `routerLinkActive` to denote relative paths (if I am correct).
If we decide to use that for relative navigations, then `'/', ` is not necessary for the `routerLink` for absolute paths. I might be wrong though :smile:

KvelaGorrrrnio (commented):
Looks good! :+1:

It would be nice if the links from the Rundown overview page to the Rundown view page also used the implementation.

src/app/shared/components/header/header.component.html
vivanov90: @KvelaGorrrrnio Maybe I miss something but I don't see relation between routerLinkActive and route relativity. The only way to told Angular that route is absolute, not relative is to start with `"/"`. I can add routerLinkActive logic to follow the active route if we want. If I miss something I stay tuned to add it. 

Side Node: Rundown overview links are added too.

KvelaGorrrrnio: You are correct. There is no relation between them.

src/app/shared/components/header/header.component.html
```diff
@@ -1,15 +1,17 @@
 <mat-toolbar class="c-header">
   <mat-toolbar-row>
-    <sofie-logo [text]="title" (click)="navigateHome()"></sofie-logo>
+    <a class="logo-link" [routerLink]="[Paths.HOME]">
```
KvelaGorrrrnio, line 3: I am unsure whether this should start with `'/', `?
It seems to work without. Not sure why :)

KvelaGorrrrnio (approved):
Good job :rocket:

src/app/shared/components/header/header.component.html
vivanov90: It's work, and is not work with `'/'`

PR #125, opened 2024-03-11
SOF-1693/alternative - Clip piece labels depending on available space to next piece

k-primeholding-dim:
When several pieces with close/equal start time exist in a Part there is no enough space for every piece label to be presented in full width. This PR solves this by combining piece name for all pieces that have matching start time on an output layer.


src/app/rundown-view/services/tv2-piece-group.service.spec.ts
```diff
@@ -5,6 +5,80 @@ import { Tv2Piece } from '../../core/models/tv2-piece'
...
+        start: 123,
+      }) as Tv2Piece
+      const pieces: Tv2Piece[] = [pieceOne]
+
+      const testee: Tv2PieceGroupService = createTestee()
+      const result: Tv2Piece[] = testee.mergePiecesByStartOffset(pieces)
+
+      const expected: Tv2Piece[] = [pieceOne]
+
+      expect(result).toEqual(expected)
```
KvelaGorrrrnio, line 34: The test description and the implementation does not match. If `pieceOne` is modified within `mergePiecesByStartOffset`, then this test won't fail, because the expected output is based on the reference of the input. What you want to do is to create two instances of the pieces, which can be done by calling the `testEntityFactory.createPiece` twice or by spreading the instantiated piece.
The latter option does not take nested structures into account, so I would suggest using the test entity factory, e.g.:
```typescript
      const pieces: Tv2Piece[] = [
        testEntityFactory.createPiece({
          name: 'pieceName',
          start: 123,
        })
      ] as Tv2Piece[]
      const testee: Tv2PieceGroupService = createTestee()          

      const result: Tv2Piece[] = testee.mergePiecesByStartOffset(pieces)

      const expected: Tv2Piece[] = [
        testEntityFactory.createPiece({
          name: 'pieceName',
          start: 123,
        })
      ] as Tv2Piece[]
      expect(result).toEqual(expected)
```

src/app/rundown-view/services/tv2-piece-group.service.spec.ts
```diff
@@ -5,6 +5,84 @@ import { Tv2Piece } from '../../core/models/tv2-piece'
...
+
+      const pieceOne: Tv2Piece = testEntityFactory.createPiece({
+        name: 'pieceOneName',
+        start: 123,
+      }) as Tv2Piece
+      const pieceTwo: Tv2Piece = testEntityFactory.createPiece({
+        name: 'pieceTwoName',
+        start: 123,
+      }) as Tv2Piece
+      const pieces: Tv2Piece[] = [pieceOne, pieceTwo]
```
KvelaGorrrrnio, line 52: This might just be a matter of taste, but personally I think that the local variables are unnecessary in this case. Since the creation setup for the pieces are only a couple of lines, it makes it easier to grasp the linearity of the construction – I don't have to go back and forth between the `pieces` list of object references and the instantiation of the references. I would inline the piece creation inside the `pieces` instantiation like so:
```typescript
      const pieces: Tv2Piece[] = [
        testEntityFactory.createPiece({
          name: 'pieceOneName',
          start: 123,
        }),
        testEntityFactory.createPiece({
          name: 'pieceTwoName',
          start: 123,
        }),
      ] as Tv2Piece []
```

I would do the same for the `expected` result.

KvelaGorrrrnio (approved):
Good job 🚀

LindvedKrvang (approved):

PR #127, opened 2024-03-11
SOF-1566 Introduce RundownMode on Rundown

LindvedKrvang:
Breaking change: This changes `Rundown.isActive` to use the newly introduce `RundownMode`.
This PR must be merged at the same time as https://github.com/tv2/sofie-server/pull/105.

This PR does _not_ implement anything with the new REHEARSAL mode. It just make the frontend compatible with it.

### DIFF
--- a/src/app/core/enums/rundown-mode.ts
+++ b/src/app/core/enums/rundown-mode.ts
@@ -1,0 +1,5 @@
+export enum RundownMode {
+  ACTIVE = 'ACTIVE',
+  REHEARSAL = 'REHEARSAL',
+  INACTIVE = 'INACTIVE',
+}
--- a/src/app/core/services/rundown-timing-context-state.service.ts
+++ b/src/app/core/services/rundown-timing-context-state.service.ts
@@ -93,12 +93,12 @@ export class RundownTimingContextStateService {
 
   private rescheduleTimeTick(): void {
     clearTimeout(this.timeResolutionTimerId)
-    const timeTickDurationInMs: number = this.hasActiveRundown() ? HIGH_RESOLUTION_INTERVAL_DURATION_IN_MS : LOW_RESOLUTION_INTERVAL_DURATION_IN_MS
+    const timeTickDurationInMs: number = this.isAllRundownsInactive() ? LOW_RESOLUTION_INTERVAL_DURATION_IN_MS : HIGH_RESOLUTION_INTERVAL_DURATION_IN_MS
     this.timeResolutionTimerId = setTimeout(this.onTimeTick.bind(this), timeTickDurationInMs)
   }
 
-  private hasActiveRundown(): boolean {
-    return [...this.rundowns.values()].some(rundown => rundown.mode === RundownMode.ACTIVE)
+  private isAllRundownsInactive(): boolean {
+    return [...this.rundowns.values()].every(rundown => rundown.mode === RundownMode.INACTIVE)
   }
 
   public async subscribeToRundownTimingContext(rundownId: string): Promise<Observable<RundownTimingContext>> {
--- a/src/app/core/services/rundown-timing-context-state.service.ts
+++ b/src/app/core/services/rundown-timing-context-state.service.ts
@@ -5,6 +5,7 @@ import { RundownTimingService } from './rundown-timing.service'
 import { RundownTimingContext } from '../models/rundown-timing-context'
 import { Injectable } from '@angular/core'
 import { Segment } from '../models/segment'
+import { RundownMode } from '../enums/rundown-mode'
 
 const HIGH_RESOLUTION_INTERVAL_DURATION_IN_MS: number = Math.floor(1000 / 30)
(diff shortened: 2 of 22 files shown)


src/app/core/services/rundown-timing-context-state.service.ts
```diff
@@ -97,7 +98,7 @@ export class RundownTimingContextStateService {
   }
 
   private hasActiveRundown(): boolean {
-    return [...this.rundowns.values()].some(rundown => rundown.isActive)
+    return [...this.rundowns.values()].some(rundown => rundown.mode === RundownMode.ACTIVE)
```
KvelaGorrrrnio, line 101: Since this is used to determine if we should have a high or low time resolution for recalculating the timings, it would be more appropriate to have this check if there are any rundowns not inactive.
```typescript
    return [...this.rundowns.values()].some(rundown => rundown.mode !== RundownMode.INACTIVE)
```

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.html
```diff
@@ -20,7 +20,7 @@ <h2 i18n>rundown-overview-component.no-rundowns-available.message</h2>
       <ng-container matColumnDef="status">
         <mat-header-cell *matHeaderCellDef></mat-header-cell>
         <mat-cell *matCellDef="let basicRundown" >
-          <sofie-pulsating-dot *ngIf="basicRundown.isActive" [color]="Color.ON_AIR" [animationDurationSeconds]="3"></sofie-pulsating-dot>
+          <sofie-pulsating-dot *ngIf="basicRundown.mode === RundownMode.ACTIVE" [color]="Color.ON_AIR" [animationDurationSeconds]="3"></sofie-pulsating-dot>
```
KvelaGorrrrnio, line 23: I assume that having a differently styled indicator for rehearsal is not part of this task. But I think that the pulsating dot should show as is when in rehearsal mode.

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -166,4 +167,8 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
     this.rundownEventSubscriptions.forEach(subscription => subscription.unsubscribe)
     this.rundownActionsSubscription?.unsubscribe()
   }
+
+  public isRundownActive(): boolean {
+    return [RundownMode.ACTIVE, RundownMode.REHEARSAL].includes(this.rundown.mode)
+  }
```
KvelaGorrrrnio, line 173: Instead of evaluating this on every render call, it can be computed in `ngOnChange` and be cached.

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

k-primeholding-dim (approved):

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.html
LindvedKrvang: That is something to be handled in SOF-1568, but it is an acceptance criteria of the story. However, I agree with Rasmus, it should not be showing the ON_AIR color
image

src/app/rundown-view/components/rundown/rundown.component.ts
LindvedKrvang: I have changed it to only be updated in `ngOnChanges`.

Ideally, however, we should not have this attribute in this component at all. It is only used to be parsed down through _3_ components before it is being used. The components that wants to actually use this value should just get the information from the RundownStateService.
But doing that is a refactor a bit out of scope of the intention of this PR.

src/app/core/services/rundown-timing-context-state.service.ts
LindvedKrvang: But the method is called `hasActiveRundown`. The method implementation doesn't know what it is being used for. And it could easily be used for more than what it is currently being used for.
That it is being used to determine the time resolution for recalculating timings is out of scope of the method.

So either we do:
```
private hasActiveOrRehearsalRundown(): boolean {
  return [...this.rundowns.values()].some(rundown => [RundownMode.ACTIVE, RundownMode.REHEARSAL].includes(rundown.mode))
}
```
Or 
```
private isAllRundownsInactive(): boolean {
  return [...this.rundowns.values()].every(rundown => rundown.mode === RundownMode.INACTIVE)
}
```

I have opted for the last suggestion and switched the values returned in the ternary where the method is used.

src/app/core/enums/rundown-mode.ts
```diff
@@ -0,0 +1,5 @@
+export enum RundownMode {
```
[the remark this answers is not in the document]
LindvedKrvang: I can follow the logic.
Useally, when we talk about OnAir we talk about the actual Part being OnAir.
The way it is currently being used is:
- `ACTIVE`: The Rundown is being used for actively producing a show
- `REHEARSAL`: The Producer is testing that his Rundown works as expected.
- `INACTIVE`: The Rundown is not being used at all.

PR #129, opened 2024-03-13
SOF-1519 - add logic to activate and listen for rehearsal rundown mode

vivanov90:
- Add logic to can activate rehearsal rundown mode
- Remove isActive boolean flag from rundown models and add RundownMode enum
- Listen for put any rundown in rehearsal mode

### DIFF
--- a/src/app/core/abstractions/rundown.service.ts
+++ b/src/app/core/abstractions/rundown.service.ts
@@ -4,7 +4,7 @@ import { Rundown } from '../models/rundown'
 export abstract class RundownService {
   public abstract fetchRundown(rundownId: string): Observable<Rundown>
   public abstract activate(rundownId: string): Observable<void>
-  public abstract rehearsal(rundownId: string): Observable<void>
+  public abstract rehearse(rundownId: string): Observable<void>
   public abstract deactivate(rundownId: string): Observable<void>
   public abstract reset(rundownId: string): Observable<void>
   public abstract takeNext(rundownId: string): Observable<void>
--- a/src/app/core/abstractions/rundown.service.ts
+++ b/src/app/core/abstractions/rundown.service.ts
@@ -4,6 +4,7 @@ import { Rundown } from '../models/rundown'
 export abstract class RundownService {
   public abstract fetchRundown(rundownId: string): Observable<Rundown>
   public abstract activate(rundownId: string): Observable<void>
+  public abstract rehearsal(rundownId: string): Observable<void>
   public abstract deactivate(rundownId: string): Observable<void>
   public abstract reset(rundownId: string): Observable<void>
   public abstract takeNext(rundownId: string): Observable<void>
--- a/src/app/core/services/basic-rundown-state.service.ts
+++ b/src/app/core/services/basic-rundown-state.service.ts
@@ -61,7 +61,7 @@ export class BasicRundownStateService implements OnDestroy {
   private subscribeToRundownEvents(): EventSubscription[] {
     return [
       this.rundownEventObserver.subscribeToRundownActivation(this.activateBasicRundownFromEvent.bind(this)),
-      this.rundownEventObserver.subscribeToRundownRehearse(this.rehearsalBasicRundownFromEvent.bind(this)),
+      this.rundownEventObserver.subscribeToRundownRehearsal(this.rehearseBasicRundownFromEvent.bind(this)),
       this.rundownEventObserver.subscribeToRundownDeactivation(this.deactivateBasicRundownFromEvent.bind(this)),
       this.rundownEventObserver.subscribeToRundownCreation(this.createBasicRundownFromEvent.bind(this)),
       this.rundownEventObserver.subscribeToRundownUpdates(this.updateBasicRundownFromEvent.bind(this)),
@@ -79,12 +79,12 @@ export class BasicRundownStateService implements OnDestroy {
     this.basicRundownsSubject.next(this.basicRundowns)
   }
 
-  private rehearsalBasicRundownFromEvent(event: RundownEvent): void {
+  private rehearseBasicRundownFromEvent(event: RundownEvent): void {
     this.basicRundowns = this.basicRundowns.map(basicRundown => {
       if (basicRundown.id !== event.rundownId) {
         return basicRundown
       }
-      return this.basicRundownEntityService.rehearsal(basicRundown)
+      return this.basicRundownEntityService.rehearse(basicRundown)
     })
     this.basicRundownsSubject.next(this.basicRundowns)
   }
--- a/src/app/core/services/basic-rundown-state.service.ts
+++ b/src/app/core/services/basic-rundown-state.service.ts
@@ -61,6 +61,7 @@ export class BasicRundownStateService implements OnDestroy {
   private subscribeToRundownEvents(): EventSubscription[] {
     return [
       this.rundownEventObserver.subscribeToRundownActivation(this.activateBasicRundownFromEvent.bind(this)),
+      this.rundownEventObserver.subscribeToRundownRehearse(this.rehearsalBasicRundownFromEvent.bind(this)),
       this.rundownEventObserver.subscribeToRundownDeactivation(this.deactivateBasicRundownFromEvent.bind(this)),
       this.rundownEventObserver.subscribeToRundownCreation(this.createBasicRundownFromEvent.bind(this)),
       this.rundownEventObserver.subscribeToRundownUpdates(this.updateBasicRundownFromEvent.bind(this)),
@@ -74,6 +75,16 @@ export class BasicRundownStateService implements OnDestroy {
         return basicRundown
       }
       return this.basicRundownEntityService.activate(basicRundown)
+    })
+    this.basicRundownsSubject.next(this.basicRundowns)
+  }
+
+  private rehearsalBasicRundownFromEvent(event: RundownEvent): void {
+    this.basicRundowns = this.basicRundowns.map(basicRundown => {
+      if (basicRundown.id !== event.rundownId) {
+        return basicRundown
+      }
+      return this.basicRundownEntityService.rehearsal(basicRundown)
     })
     this.basicRundownsSubject.next(this.basicRundowns)
   }
--- a/src/app/core/services/http/http-rundown.service.ts
+++ b/src/app/core/services/http/http-rundown.service.ts
@@ -29,6 +29,10 @@ export class HttpRundownService implements RundownService {
     return this.http.put<void>(`${RUNDOWN_URL}/${rundownId}/activate`, null).pipe(catchError(error => this.httpErrorService.catchError(error)))
   }
 
+  public rehearsal(rundownId: string): Observable<void> {
+    return this.http.put<void>(`${RUNDOWN_URL}/${rundownId}/rehearse`, null).pipe(catchError(error => this.httpErrorService.catchError(error)))
+  }
+
   public deactivate(rundownId: string): Observable<void> {
     return this.http.put<void>(`${RUNDOWN_URL}/${rundownId}/deactivate`, null).pipe(catchError(err => this.httpErrorService.catchError(err)))
   }
--- a/src/app/core/services/http/http-rundown.service.ts
+++ b/src/app/core/services/http/http-rundown.service.ts
@@ -29,7 +29,7 @@ export class HttpRundownService implements RundownService {
     return this.http.put<void>(`${RUNDOWN_URL}/${rundownId}/activate`, null).pipe(catchError(error => this.httpErrorService.catchError(error)))
   }
 
-  public rehearsal(rundownId: string): Observable<void> {
+  public rehearse(rundownId: string): Observable<void> {
     return this.http.put<void>(`${RUNDOWN_URL}/${rundownId}/rehearse`, null).pipe(catchError(error => this.httpErrorService.catchError(error)))
   }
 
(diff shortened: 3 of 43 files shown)


src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -78,6 +79,16 @@ export class BasicRundownStateService implements OnDestroy {
     this.basicRundownsSubject.next(this.basicRundowns)
   }
 
+  private rehearsalBasicRundownFromEvent(event: RundownEvent): void {
```
LindvedKrvang, line 82: it should be `rehearse` not `rehearsal` here.
`rehearsal` is the noun, but we use verbs when naming methods, because methods _does something_.

Same for the `basicEntityService`

src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -61,6 +61,7 @@ export class BasicRundownStateService implements OnDestroy {
   private subscribeToRundownEvents(): EventSubscription[] {
     return [
       this.rundownEventObserver.subscribeToRundownActivation(this.activateBasicRundownFromEvent.bind(this)),
+      this.rundownEventObserver.subscribeToRundownRehearse(this.rehearsalBasicRundownFromEvent.bind(this)),
```
LindvedKrvang, line 64: Here it should be `rehearsal` and not `rehearse` because we subscribe to the noun, not the verb.

src/app/core/services/http/http-rundown.service.ts
```diff
@@ -29,6 +29,10 @@ export class HttpRundownService implements RundownService {
     return this.http.put<void>(`${RUNDOWN_URL}/${rundownId}/activate`, null).pipe(catchError(error => this.httpErrorService.catchError(error)))
   }
 
+  public rehearsal(rundownId: string): Observable<void> {
```
LindvedKrvang, line 32: This is a method name should we should use the verb `rehearse`

src/app/core/services/models/basic-rundown-entity.service.ts
```diff
@@ -17,4 +17,11 @@ export class BasicRundownEntityService {
       mode: RundownMode.INACTIVE,
     }
   }
+
+  public rehearsal(basicRundown: BasicRundown): BasicRundown {
```
LindvedKrvang, line 21: This a method name so we should use the verb `rehearse`

src/app/core/services/models/rundown-entity.service.ts
```diff
@@ -20,6 +20,14 @@ export class RundownEntityService {
     }
   }
 
+  public rehearsal(rundown: Rundown): Rundown {
```
LindvedKrvang, line 23: This a method name so we should use the verb `rehearse`

src/app/core/services/rundown-event-observer.service.ts
```diff
@@ -43,6 +44,10 @@ export class RundownEventObserver {
     return this.eventObserver.subscribe(RundownEventType.ACTIVATED, this.createEventValidatingConsumer(onActivated, this.rundownEventParser.parseRundownActivatedEvent.bind(this.rundownEventParser)))
   }
 
+  public subscribeToRundownRehearse(onActivated: (event: RundownRehearseEvent) => void): EventSubscription {
```
LindvedKrvang, line 47: Here we subscribe to the noun so it should be `rehearsal`.

Also, it looks like you forgot to change the parameter. It should be `onRehearsal`

src/app/core/services/rundown-state.service.ts
```diff
@@ -111,6 +113,15 @@ export class RundownStateService implements OnDestroy {
     rundownSubject.next(activeRundown)
   }
 
+  private rehearsalRundownFromEvent(event: RundownRehearseEvent): void {
```
LindvedKrvang, line 116: This a method name so we should use the verb `rehearse`

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -48,8 +52,38 @@ export class RundownHeaderContextMenuComponent {
...
+    },
+    {
+      label: 'Reset Rundown',
+      contextAction: (): void => this.openResetRundownDialog(),
+    },
+    {
+      label: 'Reingest data',
+      contextAction: (): void => this.reingestData(),
+    },
+  ]
```
LindvedKrvang, line 76: Instead of duplicating the same entries as in the Active menu options, you can just add them to this array:
```
private readonly contextMenuOptionsForRehearseRundown: ContextMenuOption[] = [
    {
      label: 'Activate (On Air)',
      contextAction: (): void => this.openActivateRundownDialog(),
    },
    ...this.contextMenuOptionsForActiveRundown,
  ]
```

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -65,6 +99,14 @@ export class RundownHeaderContextMenuComponent {
     this.rundownService.activate(this.rundown.id).subscribe()
   }
 
+  public openRehearseRundownDialog(): void {
```
LindvedKrvang, line 102: Here the verb of the method is `open` and what `open` refers to is a noun, so it should be `Rehearsal`.

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -65,6 +99,14 @@ export class RundownHeaderContextMenuComponent {
...
+    this.dialogService.createConfirmDialog(this.rundown.name, 'Are you sure you want to rehearsal the Rundown?', 'Rehearse', () => this.rehearsalRundown())
```
LindvedKrvang, line 103: "Are you sure you want to _rehearse_ the Rundown?"

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -65,6 +99,14 @@ export class RundownHeaderContextMenuComponent {
...
+  }
+
+  public rehearsalRundown(): void {
```
LindvedKrvang, line 106: This is a method name so it should use the verb `rehearse`

src/app/core/abstractions/rundown.service.ts
```diff
@@ -4,6 +4,7 @@ import { Rundown } from '../models/rundown'
 export abstract class RundownService {
   public abstract fetchRundown(rundownId: string): Observable<Rundown>
   public abstract activate(rundownId: string): Observable<void>
+  public abstract rehearsal(rundownId: string): Observable<void>
```
LindvedKrvang, line 7: This is a method name so it should use the verb `rehearse`

src/app/rundown-view/components/rundown/rundown.component.html
```diff
@@ -4,7 +4,7 @@
                 id="{{segment.id}}"
                 *ngIf="!segment.isHidden"
                 [segment]="segment"
-                [isRundownActive]="isRundownActive"
+                [isRundownActiveOrRehearsal]="rundown ? isActiveOrRehearsal : false"
```
LindvedKrvang, line 7: No need for the ternary here since `isActiveOrRehearsal` already returns a boolen.

My guess is that you are using the ternary to guard against the Rundown being loaded. You should just put this "guard" within the `isActiveOrRehearsal` method:
```
public get isActiveOrRehearsal(): boolean {
  if (!this.rundown) {
    return false
  }
  ... // rest of method
}
```

LindvedKrvang (commented):
It looks good. Main thing is that you flipped the noun (`rehearsal`) with the verb (`rehearse`) and vice versa a lot of places :)

src/app/core/abstractions/rundown.service.ts
vivanov90: Fixed

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
vivanov90: Fixed

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
vivanov90: Fixed

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
vivanov90: Fixed

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
vivanov90: Done

src/app/core/services/rundown-state.service.ts
vivanov90: Fixed

src/app/core/services/rundown-event-observer.service.ts
vivanov90: Fixed

src/app/core/services/models/basic-rundown-entity.service.ts
vivanov90: Fixed

src/app/core/services/http/http-rundown.service.ts
vivanov90: Fixed

src/app/core/services/basic-rundown-state.service.ts
vivanov90: Fixed

src/app/core/services/basic-rundown-state.service.ts
vivanov90: Fixed

LindvedKrvang (approved):

src/app/rundown-view/components/rundown-header/rundown-header.component.scss
```diff
@@ -17,6 +17,11 @@
     background-color: var(--rundown-header-background-active-color);
   }
 
+  &--rehearsal {
+    --scrollbar-thumb-color: var(--rehearsal-color);
+    background-color: var(--rehearsal-color);
```
KvelaGorrrrnio, line 22: Making the scrollbar thumb the same color as the background means that the user cannot see the scrollbar thumb. The scrollbar thumb should have a different color than scrollbar track.

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -170,6 +170,13 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
     return action ? { ...actionMap, [segment.id]: action } : actionMap
   }
 
+  public get isActiveOrRehearsal(): boolean {
+    if (!this.rundown) {
+      return false
+    }
+    return (this.rundown && this.rundown.mode === RundownMode.ACTIVE) || this.rundown.mode === RundownMode.REHEARSAL
```
KvelaGorrrrnio, line 177: `this.rundown &&` can be removed. The truthiness of the rundown value has already been checked in the guard on line 174-176.

To condense the condition a bit more, you can use `Array.includes`:
```typescript
    return [RundownMode.ACTIVE, RundownMode.REHEARSAL].includes(this.rundown.mode)
```

k-primeholding-dim (approved):

src/app/rundown-view/components/rundown-header/rundown-header.component.scss
vivanov90: Fix

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
```diff
@@ -48,8 +52,23 @@ export class RundownHeaderContextMenuComponent {
     },
   ]
 
+  private readonly contextMenuOptionsForRehearseRundown: ContextMenuOption[] = [
+    {
+      label: 'Activate (On Air)',
+      contextAction: (): void => this.openActivateRundownDialog(),
+    },
+    ...this.contextMenuOptionsForActiveRundown,
```
RasmusAlbrektsen, line 60: This makes use of `resetRundown()` which does not consider the fact that we might be in Rehearsal prior to a reset. So we will have a Rehearsal rundown, then reset and end up with an Activated rundown instead.

RasmusAlbrektsen (changes requested):
I believe you are already aware of this but i'm mentioning it here just in case :-)

src/app/rundown-view/components/rundown-header-context-menu/rundown-header-context-menu.component.ts
vivanov90: @RasmusAlbrektsen We check that, but that bug is related to back-end code, no front-end. After Reset rundown back-end set rundown to deactivate and after that push it to active again, front-end just fetch again the rundown data and that is the reason to see it in front-end too. That is discuss it with @LindvedKrvang  and he will make a back-end task for this issue.

RasmusAlbrektsen (approved):

PR #131, opened 2024-03-19
SOF-1747/active rundown prompt

k-primeholding-dim:
With this PR, oc Rundown activation a confirmation dialog will open that will warn if currently active rundown will be deactivated before activating the one that we intend to.

In order to get the info - a broad getBasicRundowns is introduced to BasicRundownStatService which results are filtered by mode.
Also a new DialogConfirmationService is added to accommodate switchActivateRundownDialog() and switchRehearsalRundownDialog() that are used for user interaction when manipulating a Rundown to On-Air and to Rehearsal mode.


RasmusAlbrektsen (commented):
I think it would be smart to include the rehearsal changes from develop and add this functionality for `REHEARSAL` mode since this is needed as well. I can see it is not explicitly written in the task but it was, however, blocked by those changes so i guess it was the intention.

LindvedKrvang:
> I think it would be smart to include the rehearsal changes from develop and add this functionality for `REHEARSAL` mode since this is needed as well. I can see it is not explicitly written in the task but it was, however, blocked by those changes so i guess it was the intention.

Rehearsal it's part of the acceptance criteria, så the task isn't done before it also works for rehearsal
image

src/app/rundown-view/factories/key-binding.factory.ts
```diff
@@ -142,7 +145,17 @@ export class KeyBindingFactory {
...
+    if (currentlyNonIdleBasicRundown) {
+      this.dialogService.createConfirmDialog(
+        rundown.name,
+        `Are you sure you want to activate the Rundown?\n\nThis will deactivate the Rundown "${currentlyNonIdleBasicRundown.name}"`,
+        'Activate',
+        () => this.rundownService.deactivate(currentlyNonIdleBasicRundown.id).subscribe(() => this.rundownService.activate(rundown.id).subscribe())
+      )
+    } else {
+      this.dialogService.createConfirmDialog(rundown.name, 'Are you sure you want to activate the Rundown?', 'Activate', () => this.rundownService.activate(rundown.id).subscribe())
+    }
```
LindvedKrvang, line 158: This is duplicated with the code in the RundownHeaderContextMenu.
Whenever you start writing duplicated code, something in you should trigger and you should try to see if it's possible to avoid it. In  this case, there is not really anything that is component specific, so the code could be extracted to a service. Just to give an example.

LindvedKrvang (changes requested):
I haven't reviewed the code, but testing the changes reveals some bugs:
- The user is no longer being asked to confirm if he wants to activate a Rundown, when there are no other Rundowns activated or in Rehearsal.
- There is no confirmation box _whatsoever_ when trying to put a Rundown in Rehearsal. If there is any other Rundown that is Active or in Rehearsal, they are just deactivated with no warning or notification...

LindvedKrvang (commented):
Overall it looks good :)

Main thing is that you have given `DialogService` a responsibility that it shouldn't have.

Also, a nice to have, would be to make the Confirmation Dialog red when another Rundown needs to be deactivated. Right now, it's quite easy to miss because the only thing that indicates is a text string. So having something visual different would be a good thing.
All you need to do is to provide `DialogServerity.DANGER` to the confirm dialog when another Rundown would be deactivated.

src/locale/messages.da.xlf
```diff
@@ -402,6 +402,30 @@
...
+        <source>dialog.button.rehearse</source>
+        <target state="new">Rehearse</target>
+      </trans-unit>
+      <trans-unit id="5333590585565329092" datatype="html">
+        <source>dialog.message.activation</source>
+        <target state="new">Are you sure you want to activate the Rundown?</target>
+      </trans-unit>
+      <trans-unit id="3626510894964414637" datatype="html">
+        <source>dialog.message.activation.warning</source>
+        <target state="new">Are you sure you want to activate the Rundown?\n\nThis will deactivate the Rundown</target>
```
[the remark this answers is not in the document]
k-primeholding-dim: You are correct - `\n` does not break the line, but simple multi-line translation resource string can be represented as is written in the resource file (with new line) if we use 
```
.mat-dialog-content {
    white-space: pre-line;
}
```
which is what I did.

src/app/rundown-view/factories/key-binding.factory.ts
```diff
@@ -108,8 +110,13 @@ export class KeyBindingFactory {
   }
 
   public createRundownKeyBindings(rundown: Rundown): StyledKeyBinding[] {
+    let keybindings: StyledKeyBinding[] = [
```
KvelaGorrrrnio: The complexity of this method is slowly increasing due to the overwrites of values and conditional updates. I suggest that the creation of key bindings are split up into groups which are based on the rundown state. My suggestion is to have the creation of rundown mode transitions key bindings (activation, entering rehearsal and deactivation) and the rundown interaction key bindings (reset, take, set next, mini shelf cycle) into their own categories and then use composition to chain them together. I have provided a suggestion below with this categorization.

Sidenote: It also looks like that entering rehearsal have no key binding yet, so I have added it in the suggestion with the key binding `Control + Backquote`.

```typescript
class KeyBindingFactory {
  public createRundownKeyBindings(rundown: Rundown): StyledKeyBinding[] {
    return [
      ...this.createRundownModeTransitionKeyBindings(rundown),
      ...this.createRundownInteractionKeyBindings(rundown),
    ]
  }

  private createRundownModeTransitionKeyBindings(rundown: Rundown): StyledKeyBinding[] {
    switch (rundown.mode) {
      case RundownMode.ACTIVE:
        return [
          this.createRundownKeyBinding('Deactivate Rundown', ['Control', 'Shift', 'Backquote'], () => this.deactivateRundown(rundown)),
        ]
      case RundownMode.REHEARSAL:
        return [
          this.createRundownKeyBinding('Activate Rundown', ['Backquote'], () => this.dialogConfirmationService.switchActivateRundownDialog(rundown)),
          this.createRundownKeyBinding('Deactivate Rundown', ['Control', 'Shift', 'Backquote'], () => this.deactivateRundown(rundown)),
        ]
      case RundownMode.INACTIVE:
        return [
          this.createRundownKeyBinding('Activate Rundown', ['Backquote'], () => this.dialogConfirmationService.switchActivateRundownDialog(rundown)),
          this.createRundownKeyBinding('Enter Rehearsal', ['Control', 'Backquote'], () => this.dialogConfirmationService.switchRehearsalRundownDialog(rundown)),
        ]
    }
  }

  private createRundownInteractionKeyBindings(rundown: Rundown): StyledKeyBinding[] {
    switch (rundown.mode) {
      case RundownMode.ACTIVE:
      case RundownMode.REHEARSAL:
        return [
          this.createRundownKeyBinding('Take', ['AnyEnter'], () => this.takeNext(rundown)),
          this.createRundownKeyBinding('Reset Rundown', ['Escape'], () => this.resetRundown(rundown)),
          this.createRundownKeyBinding('Set Segment Above as Next', ['Shift', 'ArrowUp'], () => this.setSegmentAboveNextAsNext(rundown)),
          this.createRundownKeyBinding('Set Segment Below as Next', ['Shift', 'ArrowDown'], () => this.setSegmentBelowNextAsNext(rundown)),
          this.createRundownKeyBinding('Set Earlier Part as Next', ['Shift', 'ArrowLeft'], () => this.setEarlierPartAsNext(rundown)),
          this.createRundownKeyBinding('Set Later Part as Next', ['Shift', 'ArrowRight'], () => this.setLaterPartAsNext(rundown)),
          this.createRundownKeyBinding('Cycle MiniShelf', ['Tab'], () => this.miniShelfCycleService.cycleMiniShelfForward(rundown)),
          this.createRundownKeyBinding('Cycle MiniShelf', ['Shift', 'Tab'], () => this.miniShelfCycleService.cycleMiniShelfBackward(rundown)),
        ]
      case RundownMode.INACTIVE:
        return [
          this.createRundownKeyBinding('Reset Rundown', ['Escape'], () => this.resetRundown(rundown)),
        ]
    }
  }
}
```
In the suggestion I have used a switch statement to differentiate between what system actions are available for the given rundown modes. It makes it easy to extend and have a low complexity.

It should be noted that this file is deleted in #142. Just so we are aware of it.

src/app/shared/services/dialog-confirmation.service.ts
```diff
@@ -0,0 +1,72 @@
...
+  private readonly rehearseMessage: string = $localize`dialog.message.rehearse`
+  private readonly rehearseWarningMessage: string = $localize`dialog.message.rehearse.warning`
+
+  constructor(
+    private readonly dialogService: DialogService,
+    private readonly basicRundownStateService: BasicRundownStateService,
+    private readonly rundownService: RundownService
+  ) {}
+
+  public switchActivateRundownDialog(rundown: Rundown): void {
```
KvelaGorrrrnio: `switch` feels like an odd word to use here in my opinion. We use `open` for dialogs other places in code.

src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -129,4 +130,8 @@ export class BasicRundownStateService implements OnDestroy {
   public ngOnDestroy(): void {
     this.eventSubscriptions.forEach(eventSubscription => eventSubscription.unsubscribe())
   }
+
+  public getNonIdleRundown(): BasicRundown | undefined {
+    return this.basicRundowns.find(rundown => [RundownMode.ACTIVE, RundownMode.REHEARSAL].includes(rundown.mode))
+  }
```
KvelaGorrrrnio: Instead of having a specific get method on the state service, I would suggest to change to `getBasicRundowns()` and only have it return the basic rundowns array. Then where the `getNonIdleRundown` is used, can do the analysis that it needs. It will make the state service usable by more.

src/app/shared/services/dialog-confirmation.service.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+
+      verify(mockedDialogService.createConfirmDialog).once()
+    })
+
+    it('should show dialog when another rundown is already in active mode when going for rehearsal', () => {
+      const testEntityFactory: TestEntityFactory = new TestEntityFactory()
+
+      const mockedRunownService = mock<RundownService>()
+
+      const mockedRehearseSubscription: Observable<void> = mock<Observable<void>>()
```
KvelaGorrrrnio: To create a test that resembles "real-life" as close as possible, we should not mock value objects like observables. You can use the `EMPTY` observable exposed by RxJS for this purpose.

src/app/shared/services/dialog-confirmation.service.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+      })
+      testee.switchRehearsalRundownDialog(testEntityFactory.createRundown({ id: 'this', mode: RundownMode.INACTIVE }))
+
+      verify(mockedDialogService.createConfirmDialog).once()
+    })
+
+    it('should show dialog when another rundown is already in active mode when going for rehearsal', () => {
+      const testEntityFactory: TestEntityFactory = new TestEntityFactory()
+
+      const mockedRunownService = mock<RundownService>()
```
KvelaGorrrrnio: Rundown is misspelled – a `d` is missing :)

src/app/shared/services/dialog-confirmation.service.ts
```diff
@@ -0,0 +1,72 @@
...
+    const nonIdleRundown: BasicRundown | undefined = this.basicRundownStateService.getNonIdleRundown()
+    if (!nonIdleRundown) {
+      this.dialogService.createConfirmDialog(rundown.name, this.activationMessage, this.activationOkButtonText, () => this.rundownService.activate(rundown.id).subscribe())
+      return
+    }
+
+    if (nonIdleRundown.mode === RundownMode.REHEARSAL && nonIdleRundown.id === rundown.id) {
+      this.dialogService.createConfirmDialog(rundown.name, this.activationMessage, this.activationOkButtonText, () => this.rundownService.activate(nonIdleRundown.id).subscribe())
+      return
+    }
```
KvelaGorrrrnio: Instead of duplicating the code with these checks, I suggest that the conditions are combined and simplified. The given rundown has already confirmed that it is not active (on air), so it must be inactive or in rehearsal. Therefore in the second condition, it is enough to check that the ids match, which leaves us with the code:
```typescript
    if (!nonIdleRundown || nonIdleRundown.id === rundown.id) {
      this.dialogService.createConfirmDialog(rundown.name, this.activationMessage, this.activationOkButtonText, () => this.rundownService.activate(rundown.id).subscribe())
      return
    }
```

src/app/shared/services/dialog-confirmation.service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    this.dialogService.createConfirmDialog(
+      rundown.name,
+      `${this.activationWarningMessage} "${nonIdleRundown.name}"`,
+      `${this.activationOkButtonText}`,
+      () => this.rundownService.deactivate(nonIdleRundown.id).subscribe(() => this.rundownService.activate(rundown.id).subscribe()),
+      DialogColorScheme.DARK,
+      DialogSeverity.DANGER
```
KvelaGorrrrnio: Is there a reason that the color scheme and severity are set here, but not in the other createConfirmDialog calls?

src/app/shared/services/dialog-confirmation.service.ts
```diff
@@ -0,0 +1,72 @@
...
+    )
+  }
+
+  public switchRehearsalRundownDialog(rundown: Rundown): void {
+    if (rundown.mode === RundownMode.REHEARSAL) {
+      return
+    }
```
KvelaGorrrrnio, line 55: It is not expressed in the task, but from my understanding we don't won't to be able to go from an active (on air) rundown to rehearsal. Shouldn't that apply for cross-rundown mode transitions as well?
@LindvedKrvang what are your thoughts?

src/app/rundown-view/factories/key-binding.factory.ts
LindvedKrvang: This method has been changed/deleted in https://github.com/tv2/ng-sofie/pull/142
I highly suggest to not start refactoring it in this PR

src/app/shared/services/dialog-confirmation.service.ts
k-primeholding-dim: When 'deactivate' is involved a 'DANGER' is used.

src/app/rundown-view/factories/key-binding.factory.ts
k-primeholding-dim: > 
> Sidenote: It also looks like that entering rehearsal have no key binding yet, so I have added it in the suggestion with the key binding `Control + Backquote`.
> 
I was on this same page and introduced (a different 'Shift+Backquote' in resemblance of the MiniShelf cycling in 'opposite' ) key binding and later removed it here https://github.com/tv2/ng-sofie/pull/131/commits/f0338066955a749ebc0e33c5c998ca86751f6fd9 
So at the end this appears to be a really missing feature.

src/app/shared/services/dialog-confirmation.service.ts
LindvedKrvang: That's a good question. I can see the argument for why it shouldn't be allowed. But this is a PO question, so we need to ask Jacob.

src/app/core/services/basic-rundown-state.service.ts
```diff
@@ -131,7 +130,7 @@ export class BasicRundownStateService implements OnDestroy {
     this.eventSubscriptions.forEach(eventSubscription => eventSubscription.unsubscribe())
   }
 
-  public getNonIdleRundown(): BasicRundown | undefined {
-    return this.basicRundowns.find(rundown => [RundownMode.ACTIVE, RundownMode.REHEARSAL].includes(rundown.mode))
+  public getBasicRundowns(): BasicRundown[] {
```
LindvedKrvang, line 133: We already have a method in this file that basically does the same as this one: `subscribeToBasicRundowns()`. We should ideally only have one method that gives us the BasicRundowns.

LindvedKrvang (approved):

src/app/core/services/basic-rundown-state.service.ts
KvelaGorrrrnio: I think there is a great enough of difference between getting a snapshot of the state and subscribing to the state to have this method.

src/app/shared/services/dialog-confirmation.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  private readonly rehearseMessage: string = $localize`dialog.message.rehearse`
+  private readonly rehearseWarningMessage: string = $localize`dialog.message.rehearse.warning`
+
+  constructor(
+    private readonly dialogService: DialogService,
+    private readonly basicRundownStateService: BasicRundownStateService,
+    private readonly rundownService: RundownService
+  ) {}
+
+  public openActivateRundownDialog(rundown: Rundown): void {
```
RasmusAlbrektsen, line 26: If i have a rundown in `REHEARSAL` and i then try to activate it (as "on air"), i will now be asked if that same rundown should be deactivated when i try to do so. I don't know if this is a huge issue, but it is for sure not the way it should be.

src/app/shared/services/dialog-confirmation.service.ts
KvelaGorrrrnio: It is actually a product team question, since we don't have a PO ;)

src/app/shared/services/dialog-confirmation.service.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+    it('should show dialog when another rundown is already in rehearsal mode when going for rehearsal', () => {
+      const testEntityFactory: TestEntityFactory = new TestEntityFactory()
+
+      const mockedRundownService: RundownService = mock<RundownService>()
+
+      const mockedRehearseSubscription: Observable<void> = EMPTY
+      when(mockedRundownService.rehearse('this')).thenReturn(instance(mockedRehearseSubscription))
+
+      const mockedDeactivateSubscription: Observable<void> = EMPTY
+      when(mockedRundownService.deactivate('another')).thenReturn(instance(mockedDeactivateSubscription))
```
KvelaGorrrrnio: Instead of using the magic strings `this` and `another`, they should be extracted to local variables. That way if we change the id of one or both of the rundowns, it won't break the test. It also makes the connection between the mock and the test rundowns clear.

src/app/shared/services/dialog-confirmation.service.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+
+      const mockedDialogService: DialogService = createDialogServiceMock()
+
+      const mockedBasicRundoStateService: BasicRundownStateService = mock<BasicRundownStateService>()
```
KvelaGorrrrnio: `Rundo` should be `Rundown`. This applies to other places in the file as well :)

src/app/shared/services/dialog-confirmation.service.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+      when(mockedBasicRundoStateService.getBasicRundowns()).thenReturn([testEntityFactory.createRundown({ id: 'another', mode: RundownMode.REHEARSAL })])
```
KvelaGorrrrnio: Shouldn't the rundown with id `this` be returned as well?

src/app/shared/services/dialog-confirmation.service.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+  })
+
+  describe(DialogConfirmationService.prototype.openActivateRundownDialog.name, () => {
+    it('should show dialog when all rundowns are idle when going for on-air', () => {
+      const testEntityFactory: TestEntityFactory = new TestEntityFactory()
+      const rundown: Rundown = testEntityFactory.createRundown({ id: 'this', mode: RundownMode.INACTIVE })
+
+      const mockedRundownService: RundownService = mock<RundownService>()
+
+      const mockedSubscription: Observable<void> = EMPTY
```
KvelaGorrrrnio: This is not a mock and is an observable, so perhaps this should be renamed to `emptyObservable`.

This should also be used for the other places in the file where the observables are mocked.

src/app/shared/services/dialog-confirmation.service.ts
k-primeholding-dim: Great catch.

RasmusAlbrektsen (approved):

src/app/shared/services/dialog-confirmation.service.ts
k-primeholding-dim: So should I wait before I merge in develop, or are we good to go?

src/app/shared/services/dialog-confirmation.service.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+    expect(testee).toBeTruthy()
+  })
+
+  describe(DialogConfirmationService.prototype.openRehearsalRundownDialog.name, () => {
+    it('should show dialog when all rundowns are idle when going for rehearsal', () => {
+      const testEntityFactory: TestEntityFactory = new TestEntityFactory()
+
+      const mockedRundownService: RundownService = mock<RundownService>()
+
+      const mockedSubscription: Observable<void> = mock<Observable<void>>()
```
KvelaGorrrrnio, line 24: This should be named as the others and use `EMPTY` instead of mocking :)

src/app/shared/services/dialog-confirmation.service.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    it('should show dialog when another rundown is already in active mode when going for rehearsal', () => {
+      const testEntityFactory: TestEntityFactory = new TestEntityFactory()
+
+      const mockedRundownService: RundownService = mock<RundownService>()
+
+      const emptyRehearseObservable: Observable<void> = EMPTY
+      when(mockedRundownService.rehearse(anyString())).thenReturn(instance(emptyRehearseObservable))
+
+      const mockedDeactivateSubscription: Observable<void> = mock<Observable<void>>()
```
KvelaGorrrrnio, line 53: This should be named as the others and use `EMPTY` instead of mocking :)

src/app/shared/services/dialog-confirmation.service.ts
KvelaGorrrrnio: We are good to go. We'll add it as a feature/bug task if necessary :)

KvelaGorrrrnio (approved):

PR #133, opened 2024-03-19
SOF-1829 Make TooltipComponent more reuseable

LindvedKrvang:
Refactor TooltipComponent to not require an implementation of mouse events to use our tooltips.

Refactor `TooltipComponent` so it is no longer dependent on the calling component making an implementation for listening on mouse events.
Instead, the directive `sofieTooltip` has been introduced.
This directive takes a template which it will show as a tooltip. So all we have do to make a tooltip now is this:
image
Or:
image

Which will give the following result:
image
image

For common tooltips that just display a text, the component `SofieTextTooltip` has been introduced. This component doesn't actually know anything about tooltips, but this is where the styling for our "text" tooltips is being kept. This also makes it extremely convenient to add tooltips to our application (see images above).

For hoverscrub, which is a tooltip edgecase, we need to do a little extra work because the hoverscrub needs to know the current position of the cursor.
To solve this, the tooltip directive emits an event with `TooltipMetadata` which currently only contains the x position of the cursor and the width of the tooltip.

Some refactoring of the Hoverscrub component has been made. This is mainly removing all references to tooltip/cursors because the Hoverscrub has no idea that it _might_ be in a tooltip.

Notice that `SofiePieceTooltipComponent` has had its "tooltip responsibilities" removed and that `SofieMiniShelfTooltipComponent` has been deleted.
Their responsibilites was to implement the mouse events for the hoverscrub which is now the responsibility of the new tooltip directive.


src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export interface TooltipMetadata {
+  xPosition: number
+  width?: number
+}
+
+@Directive({
+  selector: '[sofieTooltip]',
+})
+export class TooltipDirective {
```
RasmusAlbrektsen, line 12: Just a reminder that we need an `OnDestroy` implemented for this to remove the tooltip if we have a `tooltipComponentRef`.

src/app/rundown-view/components/rundown-tooltips/video-hover-scrub/video-hover-scrub.component.html
```diff
@@ -1,4 +1,4 @@
-<video #videoElementRef [src]="wasVideoVisible ? hoverScrubVideoSource : ''"></video>
+<video #videoElementRef [src]="isVideoVisible ? getVideoPreviewUrl() : ''"></video>
```
RasmusAlbrektsen, line 1: Is this correct renaming, i just to make sure that it has been given a second thought. `wasVideoVisible` sounds like if the video has been made visible prior to this (like if we have the element reference from before). In that case it could be named accordingly, if not i apologize for the bother.

RasmusAlbrektsen (approved):
I really like these changes! I approve on the assumption that the `onDestroy` is implemented in my prior comment.

k-primeholding-dim (approved):

src/app/rundown-view/components/rundown-tooltips/video-hover-scrub/video-hover-scrub.component.html
LindvedKrvang: `wasVideoVisible` indicates for me that we used to have the video visible, but now it no longer is, since `was` is past tense.

LindvedKrvang: The attribute is now completely gone.

src/app/shared/directives/tooltip.directive.ts
LindvedKrvang: OnDestroy has been implemented

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
+import { ApplicationRef, ComponentRef, createComponent, Directive, EventEmitter, HostListener, Input, OnDestroy, Output, TemplateRef } from '@angular/core'
+import { TooltipComponent } from '../components/tooltip/tooltip.component'
+
+export interface TooltipMetadata {
+  xPosition: number
+  width?: number
```
KvelaGorrrrnio, line 6: I am a bit confused regarding this property, and where it is set and what it is the width of.

KvelaGorrrrnio (commented):
I like the idea of using a directive for it :)
I have used `inPixels` a lot in here. If we prefer `inPx`, then that is fine by me.

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  @HostListener('mousemove', ['$event'])
+  public moveTooltip(event: MouseEvent): void {
+    if (!this.tooltipComponentRef) {
+      return
+    }
+
+    const top: number = event.clientY - event.offsetY
+    const tooltipWidth = this.tooltipComponentRef.instance.tooltipContainer.nativeElement.offsetWidth
+    const left: number = Math.min(Math.max(event.clientX - Math.ceil(tooltipWidth / 2), 0), window.innerWidth - tooltipWidth)
```
KvelaGorrrrnio, line 47: A fair bit of math is happening here. To make it understandable for us in the future, we could extract some of the parts out and name them. A suggestion:
```typescript
const MIN_HORIZONTAL_OFFSET_IN_PIXELS: number = 0
// ...
const horizontalOffsetInPixels: number = event.clientX - Math.ceil(tooltipWidth / 2)
const maxHorizontalOffsetInPixels: number = window.innerWidth - tooltipWidth
const constrainedHorizontalOffsetInPixels = Math.min(
  Math.max(horizontalOffsetInPixels, MIN_HORIZONTAL_OFFSET_IN_PIXELS),
  maxHorizontalOffsetInPixels
)

File: src/app/rundown-view/components/rundown-tooltips/video-hover-scrub/video-hover-scrub.component.ts [hunk 5]
```diff
@@ -12,15 +11,13 @@ import { TooltipMousePosition } from 'src/app/core/models/tooltips'
 })
 export class VideoHoverScrubComponent implements OnInit, OnDestroy, OnChanges {
   @Input() public filename: string
-  @Input() public tooltipHoverElementWidth: number
-  @Input() public tooltipElementHoverMousePosition?: TooltipMousePosition
+  @Input() public videoLengthInPixels: number
+  @Input() public positionInVideoInPixels: number
```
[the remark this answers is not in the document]
LindvedKrvang: If you merge those three you also move the logic of converting the position of the cursor (which is in pixels) into the "timestamp" of the video depending on how wide is the video element is (which is also in pixels) and how much of the video has already been played.

The above logic would need to be implemented in all consumers of `VideoHoverScrubComponent` which, at the time of writing, is two components.

By not merging the three inputs, we can keep the logic in one place. The place where it is needed. 
The alternative would be to make a wrapping component with the responsibility to convert pixel position into the timestamp in the video, but seems like overkill.

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface TooltipMetadata {
+  xPosition: number
+  width?: number
+}
+
+@Directive({
+  selector: '[sofieTooltip]',
+})
+export class TooltipDirective implements OnDestroy {
+  @Input() public sofieTooltip: TemplateRef<unknown>
```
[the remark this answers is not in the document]
LindvedKrvang: We could do that, but that will mean that instead of writing:
image
We would have to write:
image
And I think it make sense that when we declare that an element has a tooltip, we immediatly give it the template used for the tooltip.

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
+import { ApplicationRef, ComponentRef, createComponent, Directive, EventEmitter, HostListener, Input, OnDestroy, Output, TemplateRef } from '@angular/core'
+import { TooltipComponent } from '../components/tooltip/tooltip.component'
+
+export interface TooltipMetadata {
+  xPosition: number
```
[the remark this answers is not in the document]
LindvedKrvang: Updated to `horizontalOffsetInPixels`

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+  width?: number
+}
+
+@Directive({
+  selector: '[sofieTooltip]',
+})
+export class TooltipDirective implements OnDestroy {
+  @Input() public sofieTooltip: TemplateRef<unknown>
+  @Input() public shouldFollowCursor: boolean = false
```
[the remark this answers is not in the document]
LindvedKrvang: It is unused. I started an implementation where the tooltip would not follow the cursor (as it does for hoverscrub) but just always be displayed in the middle and top of the element (or to the right, left, below depending on input/screen position). But it turned out to be a bigger task than what I wanted to allocate time to it and then I forgot to clean up after myself.

It has been removed :)

src/app/shared/components/text-tooltip/text-tooltip.component.scss
```diff
@@ -0,0 +1,10 @@
+div {
```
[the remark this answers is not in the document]
LindvedKrvang: Moved styling to `:host` and removed the div.

src/app/shared/components/notification-icon/notification-icon.component.html
```diff
@@ -1 +1,5 @@
-<sofie-icon-button [ngClass]="{ error: hasNotification() }" [iconButton]="IconButton.CIRCLE_EXCLAMATION" [iconButtonSize]="IconButtonSize.XL" (click)="toggleNotificationPanel()"></sofie-icon-button>
+<sofie-icon-button [sofieTooltip]="tooltip" [ngClass]="{ error: hasNotification() }" [iconButton]="IconButton.CIRCLE_EXCLAMATION" [iconButtonSize]="IconButtonSize.XL" (click)="toggleNotificationPanel()"></sofie-icon-button>
```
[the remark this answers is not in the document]
LindvedKrvang: No big preference. I have updated it.

src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.html
```diff
@@ -1,36 +1,30 @@
-  <div #tooltipHoverElement class="tooltip-hover-element"></div>
+<div class="tooltip" *ngIf="tooltipContentFields.length > 0">
+  <sofie-video-hover-scrub *ngIf="shouldShowHoverScrub && !isMediaUnavailable"
+                           [filename]="media?.sourceName ?? piece.name"
+                           [playedDurationInMs]="playedDurationForPartInMs"
+                           [videoLengthInPixels]="tooltipMetadata ? tooltipMetadata.width ?? 0 : 0"
+                           [positionInVideoInPixels]="tooltipMetadata ? tooltipMetadata.xPosition : 0"
```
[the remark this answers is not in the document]
LindvedKrvang: Updated

src/app/rundown-view/components/rundown-tooltips/video-hover-scrub/video-hover-scrub.component.ts
```diff
@@ -31,57 +28,59 @@ export class VideoHoverScrubComponent implements OnInit, OnDestroy, OnChanges {
...
+    this.currentVideoTimeInMs = this.getCurrentTimeInVideo()
     this.videoElementRef.nativeElement.currentTime = this.currentVideoTimeInMs / 1000
   }
 
-  private createVideoSource(): string {
-    return this.isJingle
-      ? `${this.studioConfiguration.settings.mediaPreviewUrl}/media/preview/${this.studioConfiguration.blueprintConfiguration.JingleFolder}/${this.filename}`
-      : `${this.studioConfiguration.settings.mediaPreviewUrl}/media/preview/${this.filename}`
+  private getCurrentTimeInVideo(): number {
+    const playedDurationInMs: number = this.playedDurationInMs ? this.playedDurationInMs : 0
```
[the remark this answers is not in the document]
LindvedKrvang: Updated

src/app/rundown-view/components/rundown-tooltips/video-hover-scrub/video-hover-scrub.component.ts
```diff
@@ -31,57 +28,59 @@ export class VideoHoverScrubComponent implements OnInit, OnDestroy, OnChanges {
...
 
-  private setNewTimeForVideoElement(mousePosition?: TooltipMousePosition): void {
+  private setNewTimeForVideoElement(): void {
     if (!this.videoElementRef?.nativeElement) {
       return
     }
-    if (!mousePosition) {
+    if (this.positionInVideoInPixels === 0) {
       return
     }
```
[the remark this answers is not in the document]
LindvedKrvang: I'm not sure. The value is only used to get the position in the video in percentage. And if `positionInVideoInPixels` is zero, the percentage is also zero, which should be fine.

Removing it seems to have no effect, so it has been removed.

src/app/rundown-view/components/rundown-tooltips/video-hover-scrub/video-hover-scrub.component.ts
```diff
@@ -31,57 +28,59 @@ export class VideoHoverScrubComponent implements OnInit, OnDestroy, OnChanges {
...
+    return Math.round((this.positionInVideoInPixels / this.videoLengthInPixels) * 100)
   }
 
-  private getCurrentTimeBasedOnCursor(videoDurationInMs: number, userCursorInPercent: number, playedDurationInMs: number): number {
-    const videoDurationWithoutPlayedDuration = videoDurationInMs - playedDurationInMs
-    return videoDurationWithoutPlayedDuration > 0 ? Math.round(playedDurationInMs + (videoDurationWithoutPlayedDuration / 100) * userCursorInPercent) : playedDurationInMs
+  public setVideoPreviewUrl(): void {
+    if (!this.studioConfiguration) {
+      return
+    }
```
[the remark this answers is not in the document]
LindvedKrvang: This check has been removed.
This method was only called once, and that is when the `studioConfiguration` is fetched or it has changed. So the instance variable `studioConfiguration` has been removed and is now just a parameter of the the method.

src/app/rundown-view/components/mini-shelf/mini-shelf.component.html
```diff
@@ -19,6 +20,14 @@
...
-        <sofie-mini-shelf-tooltip [durationInMs]="mediaDurationInMsWithoutPostroll" [filename]="mediaFilename"></sofie-mini-shelf-tooltip>
     </div>
 </div>
+
+<ng-template #miniShelfTooltip>
+    <sofie-video-hover-scrub
+            [filename]="mediaFilename"
+            [videoLengthInPixels]="tooltipMetadata ? tooltipMetadata.width ?? 0 : 0"
+            [durationInMs]="mediaDurationInMsWithoutPostroll"
+            [positionInVideoInPixels]="tooltipMetadata ? tooltipMetadata.xPosition : 0">
```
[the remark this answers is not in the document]
LindvedKrvang: Updated

src/app/shared/directives/tooltip.directive.ts
LindvedKrvang: It is the width of the element that the tooltip is a tooltip for. So I guess you could call it `hostElementWidth`?

The reason it's on this interface is because the Hoverscrub needs to know the width of the element displaying the video in order to convert the mouse position into the specific timestamp in the video. And the element displaying the video is not necessarily available where we create the markup for the tooltip (this is the case for PieceTooltipComponent).

I would say the width of the host element for the tooltip make sense to have on TooltipMetadata.

I have updated it to be called `hostElementWidth`

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  @Output() public tooltipMetadata: EventEmitter<TooltipMetadata> = new EventEmitter()
```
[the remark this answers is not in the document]
LindvedKrvang: Updated to `tooltipMetadataChange`

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+    const tooltipWidth = this.tooltipComponentRef.instance.tooltipContainer.nativeElement.offsetWidth
+    const left: number = Math.min(Math.max(event.clientX - Math.ceil(tooltipWidth / 2), 0), window.innerWidth - tooltipWidth)
+    this.tooltipComponentRef.instance.updatePosition(top, left)
+
+    this.tooltipMetadata.emit({
+      xPosition: event.offsetX,
+    })
+  }
+
+  @HostListener('mouseleave', [])
```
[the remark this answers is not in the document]
LindvedKrvang: No it is not. It has been removed.

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+    document.body.append(this.tooltipComponentRef.location.nativeElement)
+  }
+
+  @HostListener('mousemove', ['$event'])
+  public moveTooltip(event: MouseEvent): void {
+    if (!this.tooltipComponentRef) {
+      return
+    }
+
+    const top: number = event.clientY - event.offsetY
```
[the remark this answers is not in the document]
LindvedKrvang: It's the top position of the tooltip, so I would be fine with calling it just that. 
My suggestion is `topPositionInPixels`

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+    const tooltipWidth = this.tooltipComponentRef.instance.tooltipContainer.nativeElement.offsetWidth
```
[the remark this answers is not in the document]
LindvedKrvang: Updated

src/app/shared/directives/tooltip.directive.ts
LindvedKrvang: Updated. Made some slight changes to your suggestion.

src/app/rundown-view/components/rundown-tooltips/video-hover-scrub/video-hover-scrub.component.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: Currently the logic for computing the video frame to show is based on how the width of OffsetablePiece is implemented. `playedDurationInMs` is only there because of how OffsetablePiece is rendered. 
MiniShelf does not make use of `playedDurationInMs` because of this. So it is only offsetable piece which would have to do some extra calculating before sending the data to hover scrub component.

src/app/shared/directives/tooltip.directive.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: I think having the `Template` suffix is nice to have. If I were to create a component programmatically, I would from the attribute know that I just can't pass in the component that I created.

src/app/shared/components/tooltip/tooltip.component.html
```diff
@@ -1,5 +1,3 @@
-<div class="c-sofie-tooltip-wrapper hidden" #tooltipElementWrapper>
-  <div class="c-sofie-tooltip-content" #tooltipElementContent>
-    <ng-content></ng-content>
-  </div>
+<div #tooltipContainer class="tooltip" [ngStyle]="{ 'top': topPosition, 'left': leftPosition }">
```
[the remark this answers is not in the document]
LindvedKrvang: Updated

src/app/shared/directives/tooltip.directive.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: I don't think that _top_ is the correct word. Yes, it is the top value we set, but we also offset the entire tooltip such that it becomes the bottom. Therefore I suggest `verticalPositionInPixels` – the `ForWindow` suffix can be omitted.

src/app/shared/components/tooltip/tooltip.component.ts
```diff
@@ -1,64 +1,22 @@
...
   selector: 'sofie-tooltip',
   templateUrl: './tooltip.component.html',
   styleUrls: ['./tooltip.component.scss'],
 })
-export class TooltipComponent implements OnChanges, OnDestroy {
-  @Input() public tooltipElementHoverMousePosition?: TooltipMousePosition
-  private isTooltipAppended: boolean = false
+export class TooltipComponent {
+  @HostBinding('style.top.px')
+  public topPosition: number = 0
```
KvelaGorrrrnio, line 10: It would be nice with a unit on this and `leftPosition`.

src/app/shared/directives/tooltip.directive.ts
KvelaGorrrrnio: I don't think that it should be a part of the `tooltipMetadata`, since hoverscrub has nothing to do with the tooltip directive or the tooltip component. It is content that is inserted. What the tooltip directive outputs is the option to listen to changes in the position within the element.

The width is set in the same components as the value is extracted again. Instead of having MiniShelfComponent and OffsetablePieceComponent adding to the tooltipMetadata object, they could compute the hoverscrub percent to use directly and then discard the tooltip metadata.

src/app/shared/directives/tooltip.directive.ts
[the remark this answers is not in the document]
LindvedKrvang: I have updated to have the suffix.

src/app/shared/directives/tooltip.directive.ts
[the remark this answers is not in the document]
LindvedKrvang: Updated to `verticalPositionInPixels`

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -0,0 +1,70 @@
...
+    document.body.append(this.tooltipComponentRef.location.nativeElement)
+  }
+
+  @HostListener('mousemove', ['$event'])
+  public moveTooltip(event: MouseEvent): void {
+    if (!this.tooltipComponentRef) {
+      return
+    }
+
+    const verticalPositionInPixels: number = event.clientY - event.offsetY
```
KvelaGorrrrnio, line 46: If we compute the bounding rectangle of the element, and use the top value, then the jitter is removed:
```typescript
const triggerElement: Element | undefined = event.target as Element | undefined
const verticalPositionInPixels: number = triggerElement?.getBoundingClientRect().top ?? 0

File: src/app/shared/components/tooltip/tooltip.component.scss [hunk 23]
```diff
@@ -1,15 +1,5 @@
-.c-sofie-tooltip-wrapper {
+:host {
```
KvelaGorrrrnio, line 1: The shadow and rounded corners have been removed by this change. By reinstating them, the file would look like:
```scss
:host {
  position: absolute;
  border-radius: 8px;
  background-color: var(--background-color);
  overflow: hidden;
  box-shadow: 1px 1px 2px var(--black-color); transform: translate(0, -100%);
  z-index: 9999;
}
```

src/app/shared/directives/tooltip.directive.ts
LindvedKrvang: It has been removed

src/app/shared/components/tooltip/tooltip.component.ts
LindvedKrvang: Updated

src/app/shared/directives/tooltip.directive.ts
LindvedKrvang: Updated

src/app/shared/components/tooltip/tooltip.component.scss
LindvedKrvang: Updated

src/app/rundown-view/components/mini-shelf/mini-shelf.component.scss
```diff
@@ -31,37 +31,37 @@
...
-  }
+.note {
+  display: flex;
+  padding: 2px;
+  height: 18px;
+  align-items: center;
+  justify-content: space-between;
+  text-shadow: var(--black-color) 0 0 2px;
+  font-weight: 450;
+  font-size: 12px;
```
KvelaGorrrrnio, line 44: font size should be in rem. The equivalent value is `0.75rem`.

src/app/rundown-view/components/mini-shelf/mini-shelf.component.ts
```diff
@@ -140,4 +146,10 @@ export class MiniShelfComponent implements OnInit, OnDestroy, OnChanges {
   public get mediaFilename(): string {
     return this.segment.metadata?.miniShelfVideoClipFile ?? ''
   }
+
+  public updatePositionInVideo(tooltipMetadata: TooltipMetadata): void {
+    const videoLengthInPixels: number = this.thumbnailRef.nativeElement.offsetWidth
+    const positionInVideoInPercent: number = Math.round((tooltipMetadata.horizontalOffsetInPixels / videoLengthInPixels) * 100)
+    this.positionInVideoInMs = Math.round((this.mediaDurationInMsWithoutPostroll / 100) * positionInVideoInPercent)
```
KvelaGorrrrnio, line 153: In mathematics, percent is almost always expressed as floating point [0,1]. So the rounding and `* 100` can be removed, resulting in:
```typescript
    const positionInVideoInPercent: number = tooltipMetadata.horizontalOffsetInPixels / videoLengthInPixels
    this.positionInVideoInMs = Math.round(this.mediaDurationInMsWithoutPostroll * positionInVideoInPercent)
```

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -37,6 +38,8 @@ export class OffsetablePieceComponent implements OnChanges, OnDestroy {
 
   public media?: Media
 
+  public positionInVideoInMs: number
```
KvelaGorrrrnio, line 41: It would be nice with a default value of `0`.

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -48,6 +51,15 @@ export class OffsetablePieceComponent implements OnChanges, OnDestroy {
     }
   }
 
+  public updatePositionInVideo(tooltipMetadata: TooltipMetadata): void {
+    const videoLengthInPixels: number = this.widthInPixels
+    const positionInVideoInPercent: number = Math.round((tooltipMetadata.horizontalOffsetInPixels / videoLengthInPixels) * 100)
+
+    const videoDurationWithoutPlayedDurationInMs: number = Math.max(this.partDuration - this.playedDurationForPartInMs, 0)
+
+    this.positionInVideoInMs = Math.round(this.playedDurationForPartInMs + (videoDurationWithoutPlayedDurationInMs * positionInVideoInPercent) / 100)
```
KvelaGorrrrnio, line 60: Video length here might be a bit confusing, since it describes the displayed width of the piece. I don't think it is necessary to assign `this.widthInPixels` to a constant. It can be used directly.

And the upscaling to 100 can also be removed here, resulting in:
```typescript
    const positionInVideoInPercent: number = tooltipMetadata.horizontalOffsetInPixels / this.widthInPixels
    const videoDurationWithoutPlayedDurationInMs: number = Math.max(this.partDuration - this.playedDurationForPartInMs, 0)
    this.positionInVideoInMs = Math.round(this.playedDurationForPartInMs + videoDurationWithoutPlayedDurationInMs * positionInVideoInPercent)
```

KvelaGorrrrnio (approved):
Nice work! 🪨

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
LindvedKrvang: Updated

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
LindvedKrvang: Updated

src/app/rundown-view/components/mini-shelf/mini-shelf.component.scss
LindvedKrvang: Updated

src/app/rundown-view/components/mini-shelf/mini-shelf.component.ts
LindvedKrvang: Updated

PR #134, opened 2024-03-20
SOF-1676 - add dark and light theme to confirm dialogs

vivanov90:
Add dark and light theme to confirm dialog modals, light theme is default one and only if we set dark theme something is changed


RasmusAlbrektsen (approved):

src/app/shared/components/confirmation-dialog/confirmation-dialog.component.ts
```diff
@@ -8,6 +8,11 @@ export enum DialogSeverity {
   DANGER = 'DANGER',
 }
 
+export enum DialogColorTheme {
```
LindvedKrvang, line 11: `color` seems to be implied by `theme`, so you could just name it `DialogTheme` :)

LindvedKrvang (approved):

src/app/shared/components/confirmation-dialog/confirmation-dialog.component.ts
KvelaGorrrrnio: I would say that either `DialogTheme` or `DialogColorScheme`. Personally I am favoring `DialogColorScheme` (or `ColorScheme`), since theme often include other stylistic attributes than color.

src/app/shared/services/dialog.service.ts
```diff
@@ -15,7 +15,7 @@ export class DialogService {
     return this.dialog.open(component, config)
   }
 
-  public createConfirmDialog(title: string, message: string, okButtonText: string, onOk: () => void, severity?: DialogSeverity): void {
+  public createConfirmDialog(title: string, message: string, okButtonText: string, onOk: () => void, theme?: DialogColorTheme, severity?: DialogSeverity): void {
```
KvelaGorrrrnio: I don't think we need to have `theme` and `severity` as optional properties. When opening a dialog, it is fair to assume that the user of the method knows what theme and severity they desire.

src/app/shared/components/confirmation-dialog/confirmation-dialog.component.html
```diff
@@ -6,11 +6,11 @@ <h2 mat-dialog-title>{{title}}</h2>
         <span [mat-dialog-close]="false" class="material-icons c-close-dialog-icon">close</span>
     </div>
 </div>
-<div class="c-dialog-body">
+<div class="c-dialog-body" [class.dark]="theme === DialogColorTheme.DARK">
```
KvelaGorrrrnio: Just a suggestion: To easily accommodate other color schemes, the `[class.dark]` attribute could be set as `[ngClass]="theme.toLowerCase()"`.

src/app/shared/components/confirmation-dialog/confirmation-dialog.component.ts
vivanov90: Rename

KvelaGorrrrnio (approved):
Looks good :+1:

src/app/shared/services/dialog.service.ts
vivanov90: This way we have default one, only down side of the is we can't set `severity ` without `theme`, when they are set it this way

src/app/shared/components/confirmation-dialog/confirmation-dialog.component.html
vivanov90: Done

PR #137, opened 2024-03-25
Sof 1835/bug aligned table sort arrows

vivanov90:
Aline Sofie table sort arrows to be in to the center of the table header row.

### DIFF
--- a/src/app/shared/components/table/table.component.ts
+++ b/src/app/shared/components/table/table.component.ts
@@ -71,6 +71,10 @@ export class TableComponent<T> implements OnChanges {
     return header.sortDirection === SortDirection.ASC ? Icon.UPWARD_TRIANGLE : Icon.DOWNWARD_TRIANGLE
   }
 
+  public getSortIconTopPostion(header: SofieTableHeader<T>): number {
+    return header.sortDirection === SortDirection.ASC ? 3 : -3
+  }
+
   public toggleAllEntities(isSelected: boolean): void {
     this.entities.forEach(actionPanel => this.toggleEntity(isSelected, actionPanel))
   }
--- a/src/app/shared/components/table/table.component.ts
+++ b/src/app/shared/components/table/table.component.ts
@@ -12,6 +12,9 @@ export interface SofieTableHeader<T> {
   sortDirection?: SortDirection
 }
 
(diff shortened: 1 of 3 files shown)


src/scss/_sofie-utility-classes.scss
```diff
@@ -270,6 +270,14 @@
   visibility: hidden !important;
 }
 
+.relative {
+  position: relative !important;
+}
+
+.absolute {
+  position: absolute !important;
+}
```
LindvedKrvang, line 279: Are you using this class?
We should try to refrain from adding code that we are not using. We can always add it once we actually need it.

src/app/shared/components/table/table.component.ts
```diff
@@ -71,6 +71,10 @@ export class TableComponent<T> implements OnChanges {
     return header.sortDirection === SortDirection.ASC ? Icon.UPWARD_TRIANGLE : Icon.DOWNWARD_TRIANGLE
   }
 
+  public getSortIconTopPostion(header: SofieTableHeader<T>): number {
+    return header.sortDirection === SortDirection.ASC ? 3 : -3
```
LindvedKrvang, line 75: `3` and `-3` are magic numbers. Would be nice if they atleast got a variable with a name.

LindvedKrvang (commented):
I know the icons are technically in the same position now, but given the nature of the two icons we are using, it still "look" like the icons are jumping up and down.

https://github.com/tv2/ng-sofie/assets/23049647/a8a4839d-d2ec-4d67-b989-07e0c852d546

The "base" of the icon (the horizontal line of the icon) should be aligned veritcallyl. This makes it look like that the icon just "flips" instead of looking like the icon is moving. 
Furthermore, the "base" should be placed so it is veritically centered with the label to the left.
To summarize: the "arrow-up" icon is centered correctly. The "arrow-down" icon needs to be moved further down.

src/scss/_sofie-utility-classes.scss
vivanov90: Removed

src/app/shared/components/table/table.component.ts
vivanov90: Done

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #139, opened 2024-03-25
Fix/unstable tooltip trigger element

KvelaGorrrrnio:
The tooltip position was based on the emitted mouse event, which would cause it to use the sub elements instead of the element with the tooltip directive.
By using the `ElementRef` as position reference, the position becomes stable.

A tiny bit of spacing between the triangle icon and the label has been added:
Skærmbillede 2024-03-25 kl  15 17 45

Also fixs video hover scrub not showing time code.

### DIFF
--- a/src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.scss
+++ b/src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.scss
@@ -27,10 +27,6 @@
     sofie-icon-button {
       color: var(--black-color);
     }
-
-    span {
-      margin-left: 4px;
-    }
   }
 
   .content-fields {
--- a/src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.scss
+++ b/src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.scss
@@ -27,6 +27,10 @@
     sofie-icon-button {
       color: var(--black-color);
     }
+
+    span {
+      margin-left: 4px;
+    }
   }
 
   .content-fields {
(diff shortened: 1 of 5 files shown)


src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.scss
```diff
@@ -27,6 +27,10 @@
     sofie-icon-button {
       color: var(--black-color);
     }
+
+    span {
+      margin-left: 4px;
+    }
```
LindvedKrvang, line 33: Instead of setting margins in CSS we should start to utilize our classes for this. So we should add `ms-1` on the span element instead of writing a class for it.

src/app/shared/directives/tooltip.directive.ts
```diff
@@ -17,7 +17,10 @@ export class TooltipDirective implements OnDestroy {
 
   private tooltipComponentRef: ComponentRef<TooltipComponent>
 
-  constructor(private readonly applicationRef: ApplicationRef) {}
+  constructor(
+    private readonly applicationRef: ApplicationRef,
+    private readonly triggerElementRef: ElementRef
```
LindvedKrvang, line 22: Since this is the `ElementRef` injected into the constructor, we should just call it `elementRef`.

LindvedKrvang (approved):

src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.scss
KvelaGorrrrnio: I have updated it to use `ms-1`.

src/app/shared/directives/tooltip.directive.ts
KvelaGorrrrnio: Changed to `elementRef`.

LindvedKrvang (approved):

k-primeholding-dim (approved):

RasmusAlbrektsen (approved):

vivanov90 (approved):

PR #142, opened 2024-03-26
SOF-1743 Don't generate all keyboard bindings on every single update

LindvedKrvang:
'SystemKeybindings' are only updated when there is a Rundown change. 'ActionTriggerKeybindings' are only updated when there is an update to ActionTriggers or Actions.

Renamed `KeyBindingFactory` to `SystemKeyBindingFactory`.

Updated deafult value for `mappedKeys` when creating an ActionTrigger.


src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -94,26 +96,16 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
...
-      return keyBindings
-    }
-
+  private updateActionTriggerKeybindings(): void {
+    const keyBindings: KeyBinding[] = []
     this.actionTriggersWithAction.forEach((action: Tv2Action, actionTrigger: ActionTrigger<KeyboardTriggerData>) => {
       keyBindings.push(this.createBinding(action, actionTrigger, this.rundown!.id))
     })
-    return keyBindings
+    this.actionTriggerKeyBindings = keyBindings
```
KvelaGorrrrnio: The complexity of this can be reduced with by making this into a single expression either with `Object.entries`, `[...mapVariable]` or `Array.from`. I have provided a suggestion with `Array.from`.
```typescript
this.actionTriggerKeyBindings = Array.from(
  this.actionTriggersWithAction,
  ([actionTrigger, action]) => this.createBinding(action, actionTrigger, this.rundown!.id)
)
```

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -185,7 +178,25 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
...
+
+    if (this.rundown.mode === RundownMode.INACTIVE) {
+      this.systemKeyBindings = this.systemKeyBindingFactory.createInactiveRundownKeyBindings(this.rundown)
+      return
+    }
+
+    this.systemKeyBindings =
+      this.rundown.mode === RundownMode.ACTIVE
+        ? this.systemKeyBindingFactory.createActiveRundownKeyBindings(this.rundown)
+        : this.systemKeyBindingFactory.createRehearsalRundownKeyBindings(this.rundown)
```
KvelaGorrrrnio: This seems like the perfect case for a switch statement :D
To reduce the complexity, the generation and assignment can be separated to only have one place where we assign. E.g.

```typescript
private updateSystemKeyBindings(): void {
  this.systemKeyBindings = this.createSystemKeyBindings()
}

private createSystemKeyBindings(): KeyBinding[] {
  if (!this.rundown) {
    return []
  }

  switch (this.rundown.mode) {
    case RundownMode.INACTIVE:
      return this.systemKeyBindingFactory.createInactiveRundownKeyBindings(this.rundown)
    case RundownMode.ACTIVE:
      return this.systemKeyBindingFactory.createActiveRundownKeyBindings(this.rundown)
    case RundownMode.REHEARSAL:
      return this.systemKeyBindingFactory.createRehearsalRundownKeyBindings(this.rundown)
  }
}

File: src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts [hunk 3]
```diff
@@ -33,7 +33,7 @@ export class EditKeyboardMappingComponent implements OnInit {
       label: [this.keyboardMapping?.actionTrigger.data.label ?? ''],
       keys: [this.keyboardMapping?.actionTrigger.data.keys, [Validators.required]],
       triggerOn: [KeyEventType.RELEASED, [Validators.required]],
-      mappedToKeys: [this.keyboardMapping?.actionTrigger.data.mappedToKeys],
+      mappedToKeys: [this.keyboardMapping?.actionTrigger.data.mappedToKeys ?? []],
```
KvelaGorrrrnio, line 36: Why is this added?

LindvedKrvang: Without it, the InputField for MappedKeys would display the "x" to clear the field, but the field was already "clear". So this makes the field behave correctly.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
LindvedKrvang: I wasn't aware of the override of `Array.from()`. I'm a fan :) 
It has been updated.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts

RasmusAlbrektsen (approved):

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
KvelaGorrrrnio: I just tried removing this and nothing changed about when the clear button was shown in the keyboard input field. But it changes the runtime type of mappedToKeys from `Keys | undefined` to `Keys | string[] | undefined`.

LindvedKrvang: Sorry, I got it a bit messed up. If you don't have the `?? []` then when you create a new KeyboardMapping, the `mappedKeys` is going to have the value `null`, which is not allowed in our Zod parser.

Try creating a new KeyboardMapping and check your console.

KvelaGorrrrnio: If we use `[]` as the default value for the mappedToKeys input field, then we should check for an empty array and convert it to `undefined` when saving.
Another solution is to make the default value of the KeyboardInputComponent `undefined` instead of `null`.

LindvedKrvang: The default value of KeyboardInputComponent is already `undefined`. It's Angular forms that puts it as `null` when you don't provide a value.
If you give the input field a value and then clear it with the "x", then the value is going to be `undefined`.

KvelaGorrrrnio: The exposed default value should be `undefined` instead of `null`, such that the input field only provides valid values.

LindvedKrvang: As mentioned in my comment above. The default value in the InputField is already `undefined`. It's Angular forms that changes the value to `null` when you don't provide a value to the `FormControl`.

KvelaGorrrrnio: We'll leave it as is and I'll create a task for fixing it.

KvelaGorrrrnio (approved):

k-primeholding-dim (approved):

vivanov90 (approved):

PR #143, opened 2024-03-26
SOF-1838 - Changes the looks of the rundown overview and tables

RasmusAlbrektsen:
The rundown overview is changed to look less like links, and with a bit of styling that looks more like the old overview page. 
Skærmbillede 2024-03-27 kl  12 54 59

Furthermore this has become a joint effort to change the appearance of the generic tables in general, with focus on table spacing, and content size interaction (that content adheres to spacing).

The table checkbox column header now indicates partial selection of rows.
Skærmbillede 2024-03-27 kl  12 54 12


src/app/settings/shelf-settings/components/shelf-action-panel-settings-page/shelf-action-panel-settings-page.component.html
```diff
@@ -27,13 +27,15 @@
                 <td>
                     {{ actionPanel.rank }}
                 </td>
-                <td>
-                    <sofie-rounded-box *ngFor="let filter of actionPanel.actionFilter" class="{{ filter | actionColorClass }} me-2">
+                <td >
+                  <div class="flex" style="flex-wrap: wrap; gap: 8px;">
```
vivanov90: I personally don't see the point of inline styles since we already have a class defined and we can add specific one for this inline styles. And in general I don't think they are a good practice.

src/app/shared/components/page/page.component.scss
```diff
@@ -1,7 +1,6 @@
 :host {
-  background: var(--light-gray-color);
-  width: 100%;
-  height: 100%;
+  display: block;
+  padding: 20px;
```
vivanov90, line 3: Padding value can come from utility classes `p-4` === `padding: 20px`

vivanov90 (approved):

src/app/settings/shelf-settings/components/shelf-action-panel-settings-page/shelf-action-panel-settings-page.component.html
KvelaGorrrrnio: The inline styling has been moved to the dedicated class `action-filters`.

LindvedKrvang: Just to note for everyone: We want to avoid inline styling whereever possible.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
```diff
@@ -1,3 +1,20 @@
...
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
+td {
+  height: 55px;
+  border-bottom: solid 1px var(--gray-color);
+  background-color: var(--light-gray-color);
```
LindvedKrvang: I must admit, it makes me sad to see the grey background color returning on the RundownOverview table.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
```diff
@@ -1,3 +1,20 @@
...
+  text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
+td {
+  height: 55px;
+  border-bottom: solid 1px var(--gray-color);
```
LindvedKrvang, line 12: You are styling our table here. The styling of our table should happen in the TableComponent, so we can keep a consistent look throughout the application.

LindvedKrvang (approved):
Looks good.
The approve is with the assumption that you move the styling of the Table away from RundownOverviewComponent and into the TableComponent.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
RasmusAlbrektsen: The only argument was to keep it "close to home" for the first release. But i really want to rethink this general page design and change it to better fit a new Sofie-styling we come up with. Rather than just white pages for the overview and settings and then a completely different looking rundown view as we have it now.

KvelaGorrrrnio (approved):

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
LindvedKrvang: That seems like a new task would need to be created for that

RasmusAlbrektsen: Well, after some discussion this has been removed anyway!

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
RasmusAlbrektsen: Since the styling is due to us wanting a different feel of the overview page rather than the one we provide in the settings, this seemed the only way at the moment. In a future iteration we should redesign the front page, and not make use of this table component.

PR #151, opened 2024-03-28
SOF-1832 - add UI form validation

vivanov90:
KeyboardMapping and ShelfActionPanelConfiguration forms:

- add red border to touched and invalid form controls
- add tooltip to save button if form is invalid
- add error labels bellow every required field
- add asterisk to every required field


src/app/settings/keyboard-mapping-settings/components/action-selector/action-selector.component.ts
```diff
@@ -11,6 +11,8 @@ import { Tv2Action } from '../../../../shared/models/tv2-action'
 export class ActionSelectorComponent implements OnInit {
   @Input() public selectedAction?: Tv2Action
   @Output() public actionSelected: EventEmitter<Tv2Action> = new EventEmitter()
+  @Input() public hasError: boolean
+  @Input() public errorMessage: string
```
KvelaGorrrrnio: `hasError` seems redundant. If the errorMessage is made optional, then the presence of `errorMessage` will yield the same information.

src/app/shared/components/text-input/text-input.component.html
```diff
@@ -1,7 +1,15 @@
...
-        <label *ngIf="label">{{ label }}</label>
-        <sofie-icon-button *ngIf="helpText" [tooltip]="helpText" [icon]="Icon.CIRCLE_QUESTION" [iconSize]="IconSize.M"></sofie-icon-button>
-    </span>
-    <input [placeholder]="placeholder ?? label" type="text" [ngModel]="value" (ngModelChange)="onChange($event)" [disabled]="isDisabled"/>
+  <span class="mb-1">
+    <label class="flex" *ngIf="label">{{ label }}<sofie-form-required-mark *ngIf="isRequired"></sofie-form-required-mark></label>
+    <sofie-icon-button *ngIf="helpText" [tooltip]="helpText" [icon]="Icon.CIRCLE_QUESTION" [iconSize]="IconSize.M"></sofie-icon-button>
+  </span>
+  <input
+    [ngClass]="{ 'border-danger': hasError }"
```
KvelaGorrrrnio: An alternative syntax is `[class.border-danger]="hasError"`.

src/app/shared/components/form-required-mark/form-required-mark.component.html
```diff
@@ -0,0 +1 @@
+<span>*</span>
```
KvelaGorrrrnio: The span seems redundant.

src/app/shared/components/keyboard-input/keyboard-input.component.scss
```diff
@@ -17,6 +17,10 @@ input {
   font-weight: lighter;
 }
 
+sofie-row {
+  justify-content: space-between;
+}
```
KvelaGorrrrnio: This goes a bit against the reusable component approach, since `justify-content` couples to the implementation detail that it is a `flex` (or `grid`) container. If we change how `sofie-row` produces the row to e.g. a new CSS API or use `float` on all immediate children, then all the places that do overwrites like this, are likely to be broken.
So either this functionality should be embedded into the RowComponent as an input or we should evaluate if we need the row component.

KvelaGorrrrnio (commented):
Here are some first thoughts. I haven't read it all yet.

src/app/settings/keyboard-mapping-settings/components/action-selector/action-selector.component.ts
vivanov90: The idea behind this logic is to: 
- Allow an error to occur without an error message.
- Use translate logic in to the html template, not in to the TS
```
i18n-errorMessage
errorMessage="action-panel.action-filter.error"
```
That is not working and we need to move the translation it to the TS file:
```
i18n-errorMessage
[errorMessage]="checkControlForErrors('actionFilter') ? 'action-panel.action-filter.error' : undefined"
```

src/app/shared/components/keyboard-input/keyboard-input.component.scss
vivanov90: I added it as Boy Scout rule, but maybe we need separate task and more logic for that. For that reason I remove it for now.

src/app/shared/components/form-required-mark/form-required-mark.component.html
vivanov90: I added it for semantics, but will remove it

src/app/shared/components/text-input/text-input.component.html
vivanov90: Syntax updated

src/app/settings/keyboard-mapping-settings/components/action-selector/action-selector.component.html
```diff
@@ -1,9 +1,18 @@
-<sofie-text-input i18n-label label="action-selector.select-action" i18n-placeholder placeholder="action-selector.search-action" [(value)]="actionSearchQuery"></sofie-text-input>
+<sofie-text-input
```
LindvedKrvang: The ActionSelector is _not_ part of the form and should not be marked as required.
It is perfectly valid to not put any data in this input field since you can just scroll the Actions to find your desired Action:
image
What is required however, is the `Selected Action` right beneath the `Trigger On` input:
image


In short: There should not be validation in the ActionSelector, since the ActionSelector is a tool to fill out the form, not the form itself. But the `Selected Action` should be marked as required.

src/app/settings/keyboard-mapping-settings/components/action-selector/action-selector.component.ts
```diff
@@ -11,6 +11,8 @@ import { Tv2Action } from '../../../../shared/models/tv2-action'
 export class ActionSelectorComponent implements OnInit {
   @Input() public selectedAction?: Tv2Action
   @Output() public actionSelected: EventEmitter<Tv2Action> = new EventEmitter()
+  @Input() public hasError?: boolean
+  @Input() public errorMessage?: string
```
LindvedKrvang: As mentioned in my other comment, the ActionSelector is not what should be marked as required, so these fields needs to be removed again.

src/app/shared/components/text-input/text-input.component.ts
```diff
@@ -22,6 +22,9 @@ export class TextInputComponent implements ControlValueAccessor {
   @Input() public placeholder?: string
 
   @Input() public helpText: string
+  @Input() public isRequired: boolean
+  @Input() public hasError?: boolean
```
LindvedKrvang: This comment applies to all Component where you have added `hasError`.

`hasError` is redundant and should be removed. You only use it to check whether you should apply the error styling and whether you need to show the error messages.

The error messages are already optional so instead of doing:
```
<sofie-form-error-message *ngIf="hasError && errorMessage">
```
it is enough to just do:
```
<sofie-form-error-message *ngIf="errorMessage">
```
There should only be error messages to show, if the input have an actual error, so it is implicit that there is an error, if there is error messages.

For the styling: Angular already applies CSS classes to form controls for us that we can utilize https://angular.io/guide/form-validation#control-status-css-classes
So instead of doing:
```
<div class="select" [class.border-danger]="hasError">
```
we can simply in the CSS do:
```
:host.ng-invalid {
  input {
    border-color: red;
  }
}
```

Which leaves the `hasError` attribute completely redundant. And we want to minimize the amount of attributes on our components, since a component with lots of attributes is very cumbersome to work with.

src/app/shared/components/keyboard-input/keyboard-input.component.ts
```diff
@@ -33,6 +33,9 @@ export class KeyboardInputComponent implements ControlValueAccessor {
   @Input() public placeholder?: string
 
   @Input() public helpText: string
+  @Input() public errorMessage?: string
+  @Input() public isRequired?: boolean
+  @Input() public hasError?: boolean
```
LindvedKrvang: This attribute should be deleted. Refer to my comment on `TextInputComponent`.

src/app/shared/components/multi-select/multi-select.component.ts
```diff
@@ -29,6 +29,10 @@ export class MultiSelectComponent<T> implements OnInit, ControlValueAccessor {
   @Input() public label: string
   @Input() public placeholder?: string
 
+  @Input() public hasError?: boolean
```
LindvedKrvang: This attribute should be deleted. Refer to my comment on `TextInputComponent`.

src/app/shared/components/number-input/number-input.component.ts
```diff
@@ -22,6 +22,9 @@ export class NumberInputComponent implements ControlValueAccessor {
   @Input() public placeholder?: string
 
   @Input() public helpText: string
+  @Input() public isRequired?: boolean
+  @Input() public hasError?: boolean
```
LindvedKrvang: This attribute should be deleted. Refer to my comment on `TextInputComponent`.

src/app/shared/components/select/select.component.ts
```diff
@@ -20,6 +20,10 @@ export class SelectComponent<T> implements ControlValueAccessor {
   @Input() public label: string
   @Input() public placeholder?: string
 
+  @Input() public errorMessage?: string
+  @Input() public isRequired?: boolean
+  @Input() public hasError?: boolean
```
LindvedKrvang: This attribute should be deleted. Refer to my comment on `TextInputComponent`.

src/app/shared/components/text-input/text-input.component.ts
```diff
@@ -22,6 +22,9 @@ export class TextInputComponent implements ControlValueAccessor {
...
+  @Input() public errorMessage?: string
```
LindvedKrvang: I must admit, I'm very on the fence about these attributes. They are all used for form validation which happens _outside_ the scope of the component, so this component shouldn't know anything about it at all (same applies for all the other components you did this to).

Since validation is part of the form, it should ideally be taken care of by the form.

So instead of baking form validation into components that know nothing about it (because it is outside the scope of the components) we should "build" the form validation when we also build the form.

If we take a look at `EditShelfActionPanelConfigurationComponent`, then right now it looks like:
```
<form ...>
  <sofie-text-input
    ...
    [formControlName]="'name'"
    [isRequired]="true"
    [hasError]="checkControlForError('name')
    errorMessage="..."
  >
  </sofie-text-input>
  <sofie-multi-select
    ...
    [formControlName]="'actionFilter'"
    [isRequired]="true"
    [hasError]="checkControlForError('actionFilter')
    errorMessage="..."
  >
  </sofie-multi-select>
  <sofie-number-input
    ...
    [formControlName]="'rank'"
    [isRequired]="true"
    [hasError]="checkControlForError('rank')
    errorMessage="..."
  >
  </sofie-number-input>
</form>
```
If we "build" the form validation here instead it of giving the validation to components that doesn't know anything about the form, it could look something like this:
```
<form ...>
  <sofie-text-input
    ...
    class="required"
    [formControlName]="'name'"
  >
  </sofie-text-input>
  <sofie-form-error-message *ngIf="checkControlForErrors('name')">Your error message here</sofie-form-error-message>
  <sofie-multi-select
    ...
    class="required"
    [formControlName]="'actionFilter'"
  >
  </sofie-multi-select>
  <sofie-form-error-message *ngIf="checkControlForErrors('actionFiler')">Your error message here</sofie-form-error-message>
  <sofie-number-input
    ...
    class="required"
    [formControlName]="'rank'"
  >
  </sofie-number-input>
  <sofie-form-error-message *ngIf="checkControlForErrors('rank')">Your error message here</sofie-form-error-message>
</form>
```

The above moves the responsibility of form validation away from components that doesn't know what form validation it is. It makes our input components less bloated, because the simply have fewer attributes. And it gives a clearer overview of how the validation of a form works, when looking at the form.

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
```diff
@@ -51,4 +55,46 @@ export class EditShelfActionPanelConfigurationComponent implements OnInit {
...
     this.onCancel.emit()
   }
+
+  public get formValidationTooltip(): string | undefined {
+    const invalidControls = this.findInvalidControls()
+    return invalidControls.length > 0 ? $localize`common.form-required-field.button-tooltip` + `: ${invalidControls.join(', ')}` : undefined
+  }
+
+  private findInvalidControls(): string[] {
+    const invalidControlLabels = []
```
LindvedKrvang: Missing to specify type of `invalidControlLabels`. Since you have omitted it, it defaults to `any[]`, and we don't want to use `any`.

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
```diff
@@ -51,4 +55,46 @@ export class EditShelfActionPanelConfigurationComponent implements OnInit {
...
+    const controls = this.form.controls
+    for (const name in this.form.controls) {
+      if (controls[name].valid) {
+        continue
+      }
+      const translateControl = this.translateControl(name)
```
LindvedKrvang: Missing type for `translateControl`, it is not clear form this line, what the type is, so it should be included.

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
```diff
@@ -51,4 +55,46 @@ export class EditShelfActionPanelConfigurationComponent implements OnInit {
   public cancel(): void {
     this.onCancel.emit()
   }
+
+  public get formValidationTooltip(): string | undefined {
+    const invalidControls = this.findInvalidControls()
```
LindvedKrvang: Missing type for `invalidControls`. It is not clear from the line what the type is, so it should be included.
I would think that `findInvalidControls` returns `AbstractControl[]` since that is the type of FormControls, but it returns a `string[]`.

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
```diff
@@ -51,4 +55,46 @@ export class EditShelfActionPanelConfigurationComponent implements OnInit {
   public cancel(): void {
     this.onCancel.emit()
   }
+
+  public get formValidationTooltip(): string | undefined {
```
LindvedKrvang: This gets the tooltip for the save button when there is errors. That is not the same as `formValidationTooltip` so the name should be updated to reflect this

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,4 +102,60 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+    }
+  }
+
+  public checkControlForErrors(name: string, groupName?: string): boolean {
+    const control = this.getFormControl(name, groupName)
+    if (!!groupName) {
+      return control ? control.invalid && control.touched : false
+    } else {
+      return control ? control.invalid && control.touched : false
+    }
```
LindvedKrvang: The body of the positive and negative outcome is identical making the if-statement redundant.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,4 +102,60 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+  }
+
+  private getFormControl(name: string, groupName?: string): AbstractControl | undefined | null {
```
LindvedKrvang: We don't want to return `null`. Ideally, we also don't want to return `undefined`, but that's not always an option.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,4 +102,60 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+      if (controls[name].valid) {
+        continue
+      }
+      const translateControl = this.translateControl(name)
+      if (translateControl) {
+        invalidControlLabels.push(translateControl)
+      }
+    }
+    return invalidControlLabels
+  }
```
LindvedKrvang: Same comments in here as for EditShelfActionPanelConfigurationComponent.

LindvedKrvang (changes requested):
A lot of the attributes you have added to the input components are not needed.
You have also given them the responsibility of showing form validation, when these components have no idea of what form validation is and the form validation is _outside_ the scope of our input components.

src/app/settings/keyboard-mapping-settings/components/action-selector/action-selector.component.ts
LindvedKrvang: `hasError` is still redundant. See https://github.com/tv2/ng-sofie/pull/151#discussion_r1550994757

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: Fixed

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: This is remove it after code changes

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: This is remove it after code changes

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
vivanov90: I rename it to `saveButtonErrorsTooltip`

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
vivanov90: Fixed

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
vivanov90: Fixed

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
vivanov90: Fixed

src/app/shared/components/text-input/text-input.component.ts
vivanov90: @LindvedKrvang That way if we have some margins or/and paddings in to the components as `sofie-text-input`, `sofie-multi-select` the error message will go bellow this spacings, аnd it will make space between end of the component and the error message. In addition I'm not sure I understand what is the idea behind `class="required"`, we want to push the star `(*)` in :after element or am I wrong?

src/app/shared/components/select/select.component.ts
vivanov90: Fixed

src/app/shared/components/number-input/number-input.component.ts
vivanov90: Fixed

src/app/shared/components/multi-select/multi-select.component.ts
vivanov90: Fixed

src/app/shared/components/keyboard-input/keyboard-input.component.ts
vivanov90: Fixed

src/app/shared/components/text-input/text-input.component.ts
vivanov90: Fixed

src/app/settings/keyboard-mapping-settings/components/action-selector/action-selector.component.ts
vivanov90: Fixed

src/app/settings/keyboard-mapping-settings/components/action-selector/action-selector.component.html
vivanov90: Fixed

src/locale/messages.en-US.xlf
```diff
@@ -402,6 +402,42 @@
...
+        <source>action-triggers.action.label</source>
+        <target state="new">Action</target>
+      </trans-unit>
+      <trans-unit id="719177318571875364" datatype="html">
+        <source>common.form-required-field.button-tooltip</source>
+        <target state="new">You have not filled in a required field</target>
+      </trans-unit>
+      <trans-unit id="5742386463554387927" datatype="html">
+        <source>action-triggers.shortcut.error</source>
+        <target state="new">Shortcut is required field</target>
```
RasmusAlbrektsen: I'd change this to:
`A shortcut is required`

Otherwise this should at least be:
`Shortcut is a required field`

src/locale/messages.en-US.xlf
```diff
@@ -402,6 +402,42 @@
...
+      </trans-unit>
+      <trans-unit id="6359779760013296434" datatype="html">
+        <source>action-triggers.trigger-on.error</source>
+        <target state="new">Trigger on is required field</target>
```
RasmusAlbrektsen: I don't see why this one i necessary? At the moment you cannot have this field be empty. 
In any case it should probably say:
`Please select how the shortcut should trigger`

src/locale/messages.en-US.xlf
```diff
@@ -402,6 +402,42 @@
...
+      </trans-unit>
+      <trans-unit id="6127544094154745108" datatype="html">
+        <source><x id="PH" equiv-text="this.selectedAction?.argument?.name!"/> action-triggers.action-arguments.error</source>
+        <target state="new"><x id="PH" equiv-text="this.selectedAction?.argument?.name!"/> is required field</target>
+      </trans-unit>
+      <trans-unit id="979335411678822239" datatype="html">
+        <source>action-panel.name.error</source>
+        <target state="new">Name is required field</target>
```
RasmusAlbrektsen: `Please fill out a name`

src/locale/messages.en-US.xlf
```diff
@@ -402,6 +402,42 @@
...
+      </trans-unit>
+      <trans-unit id="5202318571157878044" datatype="html">
+        <source>action-panel.action-filter.error</source>
+        <target state="new">Action filter is required field</target>
```
RasmusAlbrektsen: `Select at least one filter`

src/locale/messages.en-US.xlf
```diff
@@ -402,6 +402,42 @@
...
+      </trans-unit>
+      <trans-unit id="770599904666863258" datatype="html">
+        <source>action-panel.rank.error</source>
+        <target state="new">Rank is required field</target>
```
RasmusAlbrektsen: `The panel must have a rank provided`

RasmusAlbrektsen (commented):
I didn't really go through the code yet because i have some concerns about the general design and feel that is related to this PR. 

**My thoughts:**
1. I think we should make some space between the text and symbols. There should be at least a space between them, this goes for all forms and also the “?” symbol.
Shortcut

2. We never set the “Trigger on” field to not have a value so that validation never happens, I don’t see the point in marking that as required.

3. Generally, i think we need the functionality to be that whenever you click away from something required, it should show the message. This is not the general behaviour. It works like that for the `Panel Name` in Shelf Configuration but not for the fields anywhere.

4. I have no indication that I need to select an action, or that it is required of me. Once I have selected an action, I actually see the header saying “Selected action”. This header should always be visible and the label for not having selected an action should be shown below it. Right now there is no way to get that error text. See pictures below for reference:
Select an Action
Selected action

src/locale/messages.en-US.xlf
KvelaGorrrrnio: And should end with a period, since it is a sentence. That goes for all our messages :D

vivanov90:
@RasmusAlbrektsen 

> I think we should make some space between the text and symbols. There should be at least a space between them, this goes for all forms and also the “?” symbol.

At first I had fixed it, but a faster way for me, but from @KvelaGorrrrnio comment I decided that we need to think more about the that and decided that it was not a part of the current task and removed it. If we want to fix it in this task I will think about how best to do it according to @KvelaGorrrrnio comment

> We never set the “Trigger on” field to not have a value so that validation never happens, I don’t see the point in marking that as required.

Trigger on is auto filled field, but still is required and in my head that must to be marked.

> Generally, i think we need the functionality to be that whenever you click away from something required, it should show the message. This is not the general behavior. It works like that for the Panel Name in Shelf Configuration but not for the fields anywhere.

It should be so, and if it isn't, something has been left out somewhere. I fix that in `KeyboardInputComponent` with focusout logic, but for `MultiSelectComponent` and `SelectComponent`, we need first to approve onclick #153 and #159 . Everything else must work.

> I have no indication that I need to select an action, or that it is required of me. Once I have selected an action, I actually see the header saying “Selected action”. This header should always be visible and the label for not having selected an action should be shown below it. Right now there is no way to get that error text.

I think you are right, but in a comment above we agreed with @LindvedKrvang that rather there should not be such a marking in component `SelectAction`, but it should be in `SelectedAction` form component. Maybe we can come up with some middle ground between the two marks, but I'm not sure what that should be

src/locale/messages.en-US.xlf
vivanov90: Changed

src/locale/messages.en-US.xlf
vivanov90: Changed

src/locale/messages.en-US.xlf
vivanov90: Changed

src/locale/messages.en-US.xlf
vivanov90: Changed

src/locale/messages.en-US.xlf
vivanov90: Changed

src/locale/messages.en-US.xlf
LindvedKrvang: > I don't see why this one i necessary? At the moment you cannot have this field be empty.

You can't set it to empty, but it starts empty.

LindvedKrvang:
> Trigger on is auto filled field, but still is required and in my head that must to be marked.

Agreed, it should still be marked as required.

> I think you are right, but in a comment above we agreed with @LindvedKrvang that rather there should not be such a marking in component SelectAction, but it should be in SelectedAction form component. Maybe we can come up with some middle ground between the two marks, but I'm not sure what that should be.

It sounds like you are misunderstanding what Rasmus is talking about. He saying the header `Selected Action` does not appear until an Action is selected.
image
That header should always be visible with an required asterisk.

RasmusAlbrektsen:
> It sounds like you are misunderstanding what Rasmus is talking about. He saying the header `Selected Action` does not appear until an Action is selected.

Exactly! It only shows the header once i've already selected an Action, not when i have yet to do so.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -1,39 +1,65 @@
...
 
 <ng-template #actionArguments>
-    <div class="selected-action mb-1">Selected action</div>
-    <sofie-row>
-        <sofie-action-box [action]="selectedAction!"></sofie-action-box>
-        <div class="selected-action-description ps-1">
-            {{ selectedAction!.description }}
-        </div>
-    </sofie-row>
+  <div class="selected-action mb-1"><ng-container i18n>action-triggers.selected-action.label</ng-container> <sofie-form-required-mark></sofie-form-required-mark></div>
```
LindvedKrvang: Please newline nested componenets. It's way to easy to miss components when they are formatted on one line.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,4 +102,43 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+      case FORM_TRIGGER_ON_CONTROL_LABEL:
+        return $localize`action-triggers.trigger-on.label`
+      case FORM_KEYS_CONTROL_LABEL:
+        return $localize`action-triggers.shortcut.label`
+      case FORM_ACTION_CONTROL_LABEL:
+        return $localize`action-triggers.action.label`
+      case FORM_ACTION_ARGUMENTS_CONTROL_LABEL:
+        return this.selectedAction?.argument?.name
+      default:
+        return undefined
```
LindvedKrvang: I think we should just return the untranslated control name if we can't find anything.
That way it's easier for us to spot, that we are missing an translation and we don't have to have a method that returns undefined.

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
```diff
@@ -51,4 +55,37 @@ export class EditShelfActionPanelConfigurationComponent implements OnInit {
   public cancel(): void {
     this.onCancel.emit()
   }
+
+  public get saveButtonErrorsTooltip(): string | undefined {
+    const invalidControlsLabels: string[] = this.findInvalidControlsLabels()
+    return invalidControlsLabels.length > 0 ? $localize`common.form-required-field.button-tooltip` + `: ${invalidControlsLabels.join(', ')}` : undefined
```
LindvedKrvang: A lot of stuff is happening in this ternary.
Ternaries are awesome and they should be used, but they are prone to become difficult to read quite easily.

Since you have so much going on in this one, I would suggest, that in this case, we simply use an if-statement for better readability:
```
if (invalidControlsLabels.length <= 0) {
  return
}
return $localize`common.form-required-field.button-tooltip` + `: ${invalidControlsLabels.join(', ')}`
```

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
```diff
@@ -51,4 +55,37 @@ export class EditShelfActionPanelConfigurationComponent implements OnInit {
...
+  private translateControl(controlName: string): string | undefined {
+    switch (controlName) {
+      case FORM_NAME_CONTROL_LABEL:
+        return $localize`action-panel.panel-name.label`
+      case FORM_ACTION_FILTER_CONTROL_LABEL:
+        return $localize`global.filters.label`
+      case FORM_RANK_CONTROL_LABEL:
+        return $localize`action-panel.rank.label`
+      default:
+        return undefined
```
LindvedKrvang: Same comment about returning the control name instead of undefined when translating.

src/app/shared/components/button/button.component.html
```diff
@@ -1,5 +1,7 @@
-<button class="button p-2 pe-3 ps-3"
-        [ngClass]="type.toLowerCase()"
-        [disabled]="isDisabled">
+<button class="button p-2 pe-3 ps-3" sofieTooltip [sofieTooltipTemplate]="tooltipRef" [ngClass]="type.toLowerCase()" [disabled]="isDisabled">
   <ng-content></ng-content>
 </button>
+
+<ng-template #tooltipRef>
+  <sofie-text-tooltip *ngIf="tooltip" [text]="tooltip"></sofie-text-tooltip>
```
LindvedKrvang: I think it would be nice to have the `*ngIf` inside of `sofie-text-tooltip` instead of here.
Then `sofie-text-tooltip` could take an optional string and only show it if it receives a non-empty string.
This way we don't need to write this if-statement everytime we need a text-tooltip some of the time.

Thoughts on the above is welcome.

src/app/shared/components/form-error-message/form-error-message.component.ts
```diff
@@ -0,0 +1,8 @@
+import { Component } from '@angular/core'
+
+@Component({
+  selector: 'sofie-form-error-message',
+  templateUrl: './form-error-message.component.html',
+  styleUrls: ['./form-error-message.component.scss'],
+})
+export class FormErrorMessageComponent {}
```
LindvedKrvang: This comment applies for both this component and `FormRequiredMark`.

I am not sure that `Form` should be in the name. Right now, this component is used within our inputs. Our inputs don't really know anything about forms, and it is possible to use our inputs without forms. There is also nothing in these components that references forms at all.

Now, if you did as suggested in another comment and did not put the error fields within our inputs, but instead added the error fields in the markup where we build our form, then the naming make much more sense. (This last part only applies for this component)

LindvedKrvang (approved):

src/app/shared/components/text-input/text-input.component.ts
LindvedKrvang: The idea behind `class="required"` was more so we didn't need an entire component, `FormRequiredMarkComponent`, for adding an *.

Is the padding/spacing a big issue if the error message isn't baked into our inputs? Ideally, our inputs shouldn't have a lot of padding/margins on themselves, but rather, it's possible to give them padding when building the form.
So you would do something like:
```
<text-input></text-input>
<error-message></error-message>
<text-input class="mt-3"></text-input>
<error-message></error-message>
```

src/app/shared/components/form-error-message/form-error-message.component.html
```diff
@@ -0,0 +1 @@
+<small><ng-content></ng-content></small>
```
KvelaGorrrrnio: Why is `ng-content` used here, e.g. over an input? Do we expect to add complex structures inside the error message?

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
vivanov90: @LindvedKrvang, for here not, but for Keyboard Mapping when any of the controls in `actionTriggerDataForm ` has a error data control in `actionTriggerForm` has a error too. That way we return only controls witch want to be shown in case in data control (witch is control group) we return undefined and we skip it. If we want to rework that we need to find another solution for that row:
`const controls = { ...this.actionTriggerForm.controls, ...this.actionTriggerDataForm.controls }`

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
```diff
@@ -51,4 +55,37 @@ export class EditShelfActionPanelConfigurationComponent implements OnInit {
...
+      if (translateControl) {
+        invalidControlLabels.push(translateControl)
+      }
+    }
+    return invalidControlLabels
+  }
+
+  private translateControl(controlName: string): string | undefined {
+    switch (controlName) {
+      case FORM_NAME_CONTROL_LABEL:
```
KvelaGorrrrnio: From the contents of the `X_CONTROL_LABEL` variables it does not seem to be labels, but rather attribute names or identifiers. Perhaps they should be renamed to reflect that, e.g. `PANEL_NAME_CONTROL_ID` (The `FORM_` prefix seems redundant).

src/app/shared/components/form-required-mark/form-required-mark.component.html
```diff
@@ -0,0 +1 @@
+*
```
KvelaGorrrrnio: It can be considered whether we need a file for this or if it can be set in the `@Component` template in the TypeScript file.

src/app/shared/components/keyboard-input/keyboard-input.component.ts
```diff
@@ -61,7 +63,7 @@ export class KeyboardInputComponent implements ControlValueAccessor {
     this.onChangeCallback?.(this.value)
   }
 
-  private markAsTouched(): void {
+  public markAsTouched(): void {
```
KvelaGorrrrnio: I assume this is changed to make it available in the template. If that is true, then it can be marked as protected to avoid external users to call it.

src/app/shared/components/select/select.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="container pb-2 pt-2">
-    <label *ngIf="label">{{ label }}</label>
+    <label class="flex" *ngIf="label">{{ label }}<sofie-form-required-mark *ngIf="isRequired"></sofie-form-required-mark></label>
```
KvelaGorrrrnio: Why is this marked as flex?

src/scss/_sofie-utility-classes.scss
```diff
@@ -277,3 +277,7 @@
 .width-full {
   width: 100%;
 }
+
+.border-danger {
+  border-color: var(--danger-color) !important;
+}
```
KvelaGorrrrnio: The name is a bit misleading. If I were to use this class, I would assume that it would assign a border to the element that I put the class on. It only changes the color of the border.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,4 +102,43 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+      if (controls[name].valid) {
+        continue
+      }
+      const translateControl: string | undefined = this.translateControl(name)
+      if (translateControl) {
+        invalidControlLabels.push(translateControl)
+      }
+    }
+    return invalidControlLabels
+  }
```
KvelaGorrrrnio: A suggestion for reducing the amount of code used to achieve this. If `Object.entries` is used, then we can get rid og the imperative updates to the `invalidControlLabels` variable. And by reading into what happens in the "filter", we can deduce that invalid control labels are defined as a label on an invalid control that has a translation. This can be expressed with the condition `!control.valid && !!this.translateControl(name)`.
Thus this can be reduced to:
```typescript
class EditKeyboardMappingComponent {
  private findInvalidControlsLabels(): string[] {
    return Object.entries({ ...this.actionTriggerForm.controls, ...this.actionTriggerDataForm.controls })
      .filter(([name, control]) => !control.valid && !!this.translateControl(name))
      .map(([name]) => name)
  }
}
```

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,4 +102,43 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+        return $localize`action-triggers.action.label`
+      case FORM_ACTION_ARGUMENTS_CONTROL_LABEL:
+        return this.selectedAction?.argument?.name
+      default:
+        return undefined
+    }
+  }
+
+  public get actionArgumentErrorMessage(): string {
+    return $localize`${this.selectedAction?.argument?.name!} action-triggers.action-arguments.error`
```
KvelaGorrrrnio: Why is `!` suffixed to `name`?

And a suggestion for the naming of the method: `selectedActionErrorMessage`.

src/app/shared/components/form-error-message/form-error-message.component.html
vivanov90: I'm not sure, that way it seemed more flexible to me in the particular case

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
vivanov90: @LindvedKrvang I filter the `data` control and that fix the issue.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
vivanov90: Fix

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: Done

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
vivanov90: Fix

src/app/shared/components/button/button.component.html
vivanov90: Fixed. I need to add one span to move css styles to this span, because `:host` elements is always visible and we apply black background and small padding (looks like a big black dot)

src/app/shared/components/form-error-message/form-error-message.component.ts
vivanov90: Removed

src/app/settings/shelf-settings/components/edit-shelf-action-panel-configuration/edit-shelf-action-panel-configuration.component.ts
vivanov90: Fixed

src/app/shared/components/form-required-mark/form-required-mark.component.html
vivanov90: Removed

src/app/shared/components/keyboard-input/keyboard-input.component.ts
vivanov90: Changed to protected in all places.

src/app/shared/components/select/select.component.html
vivanov90: It was because the `*` was moving to the new row, but after we changed the `sofie-required-mark` structure it is no longer necessary

src/scss/_sofie-utility-classes.scss
vivanov90: That is removed after another change, I will remove this class

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: Refactored

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: Fixed

src/app/shared/components/form-error-message/form-error-message.component.html
LindvedKrvang: You don't need to have a requirement of complex structure for using `ng-content`.

A perfectly valid usecase for `ng-content` is for a button where the content injection is just plain text:
```
// button.html
<div>
  <ng-content></ng-content>
</div>

// someOther.html
<button>Hello world</button>
```

src/app/shared/components/form-required-mark/form-required-mark.component.html
LindvedKrvang: > It can be considered whether we need a file for this or if it can be set in the @Component template in the TypeScript file.

I would advise against that. We follow a pattern where we have the html in their separate file. You are suggesting to break that pattern.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -55,11 +52,16 @@
...
     <sofie-action-box [action]="selectedAction!"></sofie-action-box>
     <div class="selected-action-description ps-1">
       {{ selectedAction!.description }}
     </div>
   </sofie-row>
 </ng-template>
+
+<ng-template #noSelectedAction>
+  <sofie-row>
+    <p i18n>action-triggers.no-selected-action.label</p>
```
LindvedKrvang: What's the reasing for using a `p`? 
The `p` puts quite a bit space between the title and the message
image
Using a `span` gives it a nicer "distance"
image

src/app/shared/components/text-tooltip/text-tooltip.component.html
```diff
@@ -1 +1,3 @@
-{{ text }}
+<span *ngIf="text">
```
LindvedKrvang, line 1: You don't need a `span` for this. 
If you need to add an "element" just for conditionals then `ng-container` is your friend, since `ng-container` does not leave an element in the dom.

LindvedKrvang (approved):

src/app/shared/components/button/button.component.html
LindvedKrvang: You can't use a `ng-container` instead of a `span`?

src/app/shared/components/text-tooltip/text-tooltip.component.html
vivanov90: @LindvedKrvang `ng-container` dose not leave element, but if we apply the css to 
```
:host {
  display: flex;
  background: var(--rundown-scrollbar-color);
  color: var(--light-gray-color);
  padding: 4px;
  border-radius: 3px;
  font-size: 0.8rem;
  box-shadow: 0 0 1px var(--dark-gray-color);
  white-space: nowrap;
}
```
The host element is there and we got this dot bug:
Screenshot 2024-04-09 132337

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
vivanov90: @LindvedKrvang I use the `p`, because I like the spacing, but now is changed to `span`

RasmusAlbrektsen (approved):

src/app/shared/components/form-required-mark/form-required-mark.component.html
vivanov90: @LindvedKrvang @KvelaGorrrrnio What to do here? With or without template?

src/app/shared/components/keyboard-input/keyboard-input.component.scss
```diff
@@ -1,3 +1,18 @@
...
+:host.ng-invalid {
+  &.ng-touched {
+    input {
+      border-color: var(--danger-color);
+    }
+    sofie-error-message {
+      display: block;
+    }
+  }
+}
```
KvelaGorrrrnio, line 14: Just a suggestion that can be used or not :)
Instead of marking the error message as hidden by default and block when it has been touched, the query can be reversed, such that we want to hide it only when we haven't toucher it. E.g.:
```scss

:host.ng-invalid {
  &.ng-touched {
    input {
      border-color: var(--danger-color);
    }
  }

  &:not(.ng-touched) {
    sofie-error-message {
      display: none;
    }
  }
}
```
It might be that it is not beneficial in some cases.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,4 +103,38 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+
+  private findInvalidControlsLabels(): string[] {
+    return Object.entries({ ...this.actionTriggerForm.controls, ...this.actionTriggerDataForm.controls })
+      .filter(([name, control]) => !control.valid && name !== KEYBOARD_DATA_CONTROL_ID)
+      .map(([name]) => this.translateControl(name))
+  }
+
+  private translateControl(controlName: string): string {
+    switch (controlName) {
+      case KEYBOARD_TRIGGER_ON_CONTROL_ID:
```
KvelaGorrrrnio, line 124: There is still a semantic mismatch between control name and control id. Only one term should be used :)

src/app/shared/components/text-tooltip/text-tooltip.component.html
LindvedKrvang: Ahh, then leave it as is :)

LindvedKrvang (approved):

src/locale/messages.en-US.xlf
```diff
@@ -402,6 +402,46 @@
         <source>notification-icon.toggle-panel.tooltip</source>
         <target state="new">Click to toggle notification panel</target>
       </trans-unit>
+      <trans-unit id="3299373246271290710" datatype="html">
+        <source>action-triggers.action.label</source>
+        <target state="new">Action</target>
+      </trans-unit>
+      <trans-unit id="719177318571875364" datatype="html">
+        <source>common.form-required-field.button-tooltip</source>
+        <target state="new">You have not filled in a required field</target>
```
KvelaGorrrrnio, line 411: Our messages should not "blame" the user. An alternate phrasing that guides the user could be: `Please fill out the required field(s).`

If we want to just point out that is not been correctly filled out, then it could be something like: `One or more required fields have not been filled out.`

src/app/shared/components/text-tooltip/text-tooltip.component.ts
```diff
@@ -6,5 +6,5 @@ import { Component, Input } from '@angular/core'
   styleUrls: ['./text-tooltip.component.scss'],
 })
 export class TextTooltipComponent {
-  @Input() public text: string
+  @Input() public text?: string
```
KvelaGorrrrnio, line 9: Why is this made optional?

src/app/shared/components/form-error-message/form-error-message.component.html
KvelaGorrrrnio: I don't see the use case for supporting it, but it is fine :)

src/app/shared/components/form-required-mark/form-required-mark.component.html
KvelaGorrrrnio: > I would advise against that. We follow a pattern where we have the html in their separate file. You are suggesting to break that pattern.

There is no HTML – just a single character in a string. But either way is fine for me. I'll let it be up to @LindvedKrvang.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,4 +103,38 @@ export class EditKeyboardMappingComponent implements OnInit {
   public cancel(): void {
     this.onCancel.emit()
   }
+
+  public get saveButtonErrorsTooltip(): string | undefined {
```
KvelaGorrrrnio, line 107: Instead of naming the getter method based on where it is used, it should state what it does/contains. A suggestion could be `invalidControlsErrorMessage`.

The undefined type can be removed as well. An empty string should be fine use when there are no errors.

The same goes for the other form component(s).

src/app/shared/components/keyboard-input/keyboard-input.component.scss
vivanov90: In that specific example we set `display: none;`, only on invalid and not touched inputs, but if the inputs are valid that css is not applied. In similar sematic I can propose that:

```
:host {
  &.ng-invalid {
    &.ng-touched {
      .input {
        border-color: var(--danger-color);
      }
    }
  }

  &.ng-valid,
  &.ng-untouched {
    sofie-error-message {
      display: none;
    }
  }
}
```

src/app/shared/components/text-tooltip/text-tooltip.component.ts
vivanov90: Because on many places we have conditional tooltip. That way we provide or not the tooltip and when we have a text we apply the styles to the div and show the tooltip if we don't have a text we don't show anything. That was discuss it with @LindvedKrvang in to the comments above.

```
<span *ngIf="text">
  {{ text }}
</span>

```

src/locale/messages.en-US.xlf
vivanov90: Changed to `Please fill out the required field(s). {Fields names}`

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: Fixed

src/app/shared/components/form-required-mark/form-required-mark.component.html
vivanov90: Revert to the template

PR #152, opened 2024-03-28
SOF-1843 - remove title from Sofie page

vivanov90:


src/app/rundown-overview/components/rundown-overview/rundown-overview.component.html
```diff
@@ -1,5 +1,5 @@
 <sofie-header></sofie-header>
-<sofie-page i18n-title title="rundown-overview.rundown-overview.title">
+<sofie-page>
```
LindvedKrvang: Why is the title removed?
We still want to have the "title" that says `Rundowns` on the overview page.

I see that the unwanted "tooltip" is now gone from the RundownOverview page. But it is still there on the Settings page. So this does not solve SOF-1843. (The video on the task shows both pages...)

It looks like the issue comes from `i18n-title` where `title` seems to be a keyword that `i18n` applies to a lot of things. Changing the input variable `title` on `SofiePage` to a different name fixes the bug.

Sidenote: In the solution you have removed the usage of a translation key, but you haven't removed the translation from the translation files which leaves us with unused translations. We should keep our translation files up-to-date as well. In this case, the translation should stay, because we still want the "title" of the RundownOverview page.

LindvedKrvang (changes requested):
This solution does not solve SOF-1843. Look at my comment to see why.

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.html
vivanov90: @LindvedKrvang Sorry, I didn't understand the assignment properly. I think now the solution is the way you want it.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.html
```diff
@@ -1,4 +1,4 @@
-<sofie-page i18n-title title="settings.keyboard-mappings.label">
+<sofie-page i18n-pageTitle pageTitle="settings.keyboard-mappings.label">
```
KvelaGorrrrnio: It has nothing to do with your change, but the changes made it quite obvious that we are using `.title` for some of the page titles and `.label` for others. It would be nice to have streamlined. Not that it has to be done in this PR :)

KvelaGorrrrnio (approved):
Looks good!

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.html
vivanov90: @KvelaGorrrrnio Here we refer to title of the page, but in most places we mean some label on button, input or something else. I only think we need only to change the name of the translate key from `settings.keyboard-mappings.labe` to `settings.keyboard-mappings.title`, for more consistency

KvelaGorrrrnio: And `settings.shelf.action-panels.label` :)

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #153, opened 2024-04-01
Sof 1833/open multi select downdown on click

vivanov90:
- Remove logic to open and close multiselect dropdown
- Add click service to emit all new body clicks and to listen with interested components
- Add custom close multiselect functionality based on existing of the className in all parentElement


src/app/shared/components/multi-select/multi-select.component.ts
```diff
@@ -49,6 +55,29 @@ export class MultiSelectComponent<T> implements OnInit, ControlValueAccessor {
...
+    const allClassNames = []
+    let clickTarget: Element = clickEvent.target as Element
+    while (clickTarget) {
+      allClassNames.push(...clickTarget.classList.value.split(' '))
+      if (!clickTarget.parentElement) {
+        break
+      }
+      clickTarget = clickTarget.parentElement
+    }
+    return !!allClassNames.find(className => className === this.className)
```
KvelaGorrrrnio: Having the click service seems overly complex and makes the multi select depend on external components to work. I am not sure what the classnames are for.

I suggest an alternative solution, where when opening the options overlay, an event listener is added to the document. The event listener is removed once it is triggered (with the `once` option). This way the multi-select component just needs to stop the event bubbling. I have tried implementing it based on `develop`. The result is that the following should be added to the component file (TypeScript):
```typescript
export class MultiSelectComponent implements ... {

  // Stop event bubbling
  @HostListener('click', ['$event'])
  protected stopEventPropagation(event: MouseEvent): void {
    event.stopPropagation()
  }

  // State variable for showing the options overlay
  protected isShowingOptions: boolean = false

  // Toggle the state variable and register document event listener, that is triggered at most once.
  protected toggleIsShowingOptions(): void {
    this.isShowingOptions = !this.isShowingOptions
    if (this.isShowingOptions) {
      document.addEventListener('click', () => (this.isShowingOptions = false), { once: true })
    }
  }
}
```

In the template, the `div.display-area` gets `(click)="toggleIsShowingOptions()"` to toggle the visibility of the options overlay and `div.multi-select-dropdown` gets `*ngIf="isShowingOptions"` to react to the toggling. And the hover selectors in the css should be removed.

vivanov90: My approach was with a  ClickService to avoid each component creating its own `HostListener`. I made the desired change, but you should keep in mind going forward that if you have a bunch of components on one page and each one has a `HostListener`, it will slow down the overall performance of the page. But certainly the logic seems much cleaner that way!

LindvedKrvang: It should only be an issue if we have multiple components registered at the same time.
But since each component should unregister itself after it closes (and each component should close itself whenever another opens), we should never have more than one listening at a time.

src/app/shared/components/multi-select/multi-select.component.ts
```diff
@@ -40,6 +41,11 @@ export class MultiSelectComponent<T> implements OnInit, ControlValueAccessor {
 
   private isTouched: boolean = false
 
+  @HostListener('click', ['$event'])
+  protected stopEventPropagation(event: MouseEvent): void {
```
LindvedKrvang: It would be very nice if the first method of each class is the constructor. This makes it easy to always find the constructor (which is the first thing run in a classs) and it gives a nice separation between instance variables and methods.

src/app/shared/components/multi-select/multi-select.component.scss
```diff
@@ -17,10 +17,6 @@ label {
   border: 1px solid var(--gray-color);
 }
 
-.multi-select:hover .multi-select-dropdown {
```
LindvedKrvang, line 20: What is the reasoning for removing this?

LindvedKrvang (commented):
Your solution is close but not entirely there.
There should only ever be one dropdown open at a time. So if you open a dropdown and click _anywhere_ else, the dropdown should close.
Right now, if you open a dropdown and then click to open another, both dropdowns are open. That should not be the case.

src/app/shared/components/multi-select/multi-select.component.scss
vivanov90: To remove the logic witch show the dropdown on hover and add this type of logic to be only in click.

LindvedKrvang: Ahh, yeah, I missed that :D

src/app/shared/components/multi-select/multi-select.component.ts
vivanov90: Fixed

src/app/shared/components/multi-select/multi-select.component.ts
KvelaGorrrrnio: Valid points :) Though it would require a quite hefty amount of listeners for it to become an issue.
As @LindvedKrvang mentions in a later comment – I did not take the scenario of the interaction between multiple selects into account.

src/app/shared/components/multi-select/multi-select.component.ts
```diff
@@ -42,6 +44,14 @@ export class MultiSelectComponent<T> implements OnInit, ControlValueAccessor {
 
   constructor() {}
 
+  @HostListener('click', ['$event'])
+  protected stopEventPropagation(): void {
+    this.isStopOptionsClosing = true
```
KvelaGorrrrnio: I am not entirely sure what `isStopOptionsClosing` is indicating :)
From the logic based on it, I assume that it is whether the component `hasBeenClicked`.
If that is the case, then the method could also be renamed to something like `registerClick`.

src/app/shared/components/multi-select/multi-select.component.scss
KvelaGorrrrnio, line 1: It would be nice if the same styling is used for both multi-select and select: https://github.com/tv2/ng-sofie/pull/159#discussion_r1555603520

KvelaGorrrrnio (approved):

src/app/shared/components/multi-select/multi-select.component.ts
vivanov90: Renamed. The idea behind this variable is to let the click event propagate (to can close all another open dropdowns), but still if the click is on the current component not to close the dropdown.

src/app/shared/components/multi-select/multi-select.component.scss
vivanov90: I will apply it.

src/app/shared/components/multi-select/multi-select.component.ts
KvelaGorrrrnio: Thanks, that makes sense.

I might not have given the best description. My suggestion was to rename `isStopOptionsClosing` to `hasBeenClicked` and `stopEventPropagation` to `registerClick`.

vivanov90: Renamed

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

PR #156, opened 2024-04-02
Fix(staging): Inserting piece into on air part breaks UI

KvelaGorrrrnio:
Changes from #149 based on `staging`. `develop` is bypassed to have it ready for stakeholder demo tomorrow.


LindvedKrvang (approved):
It would be nice if we somehow could have some test for this, so we don't all of a sudden get the bug again.
Not sure how to write those and I don't think that should block the merge of this PR

PR #159, opened 2024-04-05
SOF-1834 - add on click logic to select component

vivanov90:
- Add on click logic to select component
- Remove hover display css logic

### DIFF
--- a/src/app/shared/components/select/select.component.html
+++ b/src/app/shared/components/select/select.component.html
@@ -1,13 +1,13 @@
 <div class="container pb-2 pt-2">
     <label *ngIf="label">{{ label }}</label>
     <div class="select">
-        <div class="display-area">
+        <div class="display-area" (click)="toggleIsShowingOptions()">
             <div class="display-text">
                 {{ !!selectedOption ? selectedOption.name : (placeholder ?? label) }}
             </div>
         </div>
 
-        <div class="select-dropdown" >
+        <div class="select-dropdown" [class.block]="isShowingOptions">
             <div *ngFor="let option of options" class="select-dropdown-item p-3" (click)="selectOption(option)">
                 {{ option.name }}
             </div>
--- a/src/app/shared/components/select/select.component.html
+++ b/src/app/shared/components/select/select.component.html
@@ -7,7 +7,7 @@
             </div>
         </div>
 
-        <div class="select-dropdown" [class.block]="isShowingOptions">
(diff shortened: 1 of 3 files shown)


k-primeholding-dim (approved):

src/app/shared/components/select/select.component.html
```diff
@@ -1,13 +1,13 @@
...
     <div class="select">
-        <div class="display-area">
+        <div class="display-area" (click)="toggleIsShowingOptions()">
             <div class="display-text">
                 {{ !!selectedOption ? selectedOption.name : (placeholder ?? label) }}
             </div>
         </div>
 
-        <div class="select-dropdown" >
+        <div class="select-dropdown" [class.block]="isShowingOptions">
```
KvelaGorrrrnio: Instead of toggling a class, you can use `*ngIf="isShowingOptions"`. That way, `display: none;` can be removed from the `.select-dropdown` selector.

src/app/shared/components/select/select.component.scss
```diff
@@ -18,10 +18,6 @@ label {
   border: 1px solid var(--gray-color);
 }
 
-.select:hover .select-dropdown {
-  display: block;
-}
-
 .select-dropdown {
```
KvelaGorrrrnio, line 22: Just some minor visual quirks.
- If we use `clip-path: inset(0 -15px -15px -15px);`, we avoid having the box shadow drop over the input field itself.
- By adding the same border and border-radius to the dropdown as the select itself, it will have a more consistent feel.
- By adding `cursor: pointer;` to the `select`, it indicates to the user that you should click the input field to interact with it.

KvelaGorrrrnio (approved):
Looks good :)

src/app/shared/components/select/select.component.scss
vivanov90: All css fix are applied

src/app/shared/components/select/select.component.html
vivanov90: Done

vivanov90:
@KvelaGorrrrnio I remove HostListener because I think in that case we don't need it, because any next click must to close the dropdown. In addition @LindvedKrvang find another bug with if we have 2 or more multiSelect or selects they can be open in same time, because we `stopPropagation` the event. If you have a time you can check the new solution.

KvelaGorrrrnio (approved):

PR #161, opened 2024-04-09
SOF-1861 - move notification styles to css

vivanov90:
- Remove notification styles from TS file, add class
- Add notification styles to scss file

### DIFF
--- a/src/app/shared/components/notification-card/notification-card.component.scss
+++ b/src/app/shared/components/notification-card/notification-card.component.scss
@@ -24,7 +24,6 @@
 
 .text {
   overflow-wrap: break-word;
-  width: 100%;
 }
 
 sofie-icon-button {
--- a/src/app/shared/components/notification-card/notification-card.component.scss
+++ b/src/app/shared/components/notification-card/notification-card.component.scss
@@ -1,24 +1,30 @@
 :host {
   --gradient-degree: 90deg;
   --status-color-width: 5%;
+  --status-code-color: var(--info-color);
   padding: 10px;
   border-radius: 8px;
+  border-left: 20px solid var(--status-code-color);
   box-sizing: border-box;
   box-shadow: 1px 1px 2px var(--black-color-with-alpha);
   display: flex;
-  color: var(--black-color);
-}
-
-.notification-card {
-  display: flex;
   width: 100%;
   justify-content: space-between;
+  color: var(--black-color);
+  background-color: var(--white-color);
+
+  &.bad {
+    --status-code-color: var(--danger-color);
+  }
+
+  &.warning {
+    --status-code-color: var(--attention-color);
+  }
 }
 
 .text {
-  padding-left: var(--status-color-width);
   overflow-wrap: break-word;
-  width: calc(100% - var(--status-color-width))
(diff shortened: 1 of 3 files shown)


RasmusAlbrektsen (approved):

src/app/shared/components/notification-card/notification-card.component.scss
```diff
@@ -1,24 +1,30 @@
...
+  &.warning {
+    --status-code-color: var(--attention-color);
+  }
 }
 
 .text {
-  padding-left: var(--status-color-width);
   overflow-wrap: break-word;
-  width: calc(100% - var(--status-color-width))
+  width: 100%;
```
KvelaGorrrrnio: Is this necessary? The div is already blocking, so it should be able to take up the available width.

src/app/shared/components/notification-card/notification-card.component.ts
```diff
@@ -32,27 +23,14 @@ export class NotificationCardComponent implements OnInit {
...
-      case StatusCode.BAD:
-        return ERROR_COLOR
-      case StatusCode.WARNING:
-        return WARNING_COLOR
-      case StatusCode.GOOD:
-      case StatusCode.UNKNOWN:
-        return INFO_COLOR
-    }
+  private setClassName(): void {
+    this.elementRef.nativeElement.classList.add(this.notification.statusCode.toLocaleLowerCase())
```
KvelaGorrrrnio: This is only called once upon init. Instead of modifying the native element, a host binding be used, which will give a reactive binding. Another plus is that it will remove the need for the constructor parameter (and the explicit constructor) and `ngOnInit`.

E.g.:
```typescript
class NotificationCard {
  @HostBinding('class')
  protected get statusCodeClassName(): string {
    return this.notification.statusCode.toLocaleLowerCase()
  }
}
```

KvelaGorrrrnio (commented):
Nice work :rocket:

k-primeholding-dim (approved):

src/app/shared/components/notification-card/notification-card.component.ts
vivanov90: Done

src/app/shared/components/notification-card/notification-card.component.scss
vivanov90: Fix

src/app/shared/components/notification-card/notification-card.component.ts
```diff
@@ -29,30 +20,14 @@ export class NotificationCardComponent implements OnInit {
   public readonly Icon = Icon
   public readonly IconSize = IconSize
 
-  constructor(private readonly elementRef: ElementRef) {}
+  constructor() {}
```
KvelaGorrrrnio: The `constructor` can be removed. If it is not removed, then it should have a access modifier :)

KvelaGorrrrnio (approved):

src/app/shared/components/notification-card/notification-card.component.ts
vivanov90: Done

PR #162, opened 2024-04-10
SOF-1880 - add save and create new button and logic

vivanov90:
- add save and create new button
- increase z-index to notifications to be visible above dialog windows


src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -88,13 +88,14 @@ export class EditKeyboardMappingComponent implements OnInit {
...
 
-  public save(): void {
+  public save(isSaveAndCreateNew?: boolean): void {
     const keyboardMapping: KeyboardMapping = {
       action: {} as Tv2PartAction,
       actionTrigger: {
         ...this.keyboardMapping?.actionTrigger,
         ...this.actionTriggerForm.value,
       },
+      isSaveAndCreateNew,
```
KvelaGorrrrnio: This newly introduced attribute has nothing to do with the KeyboardMapping type. If we want to overload the `onSave` output, then we should use composition, such that the type becomes something like
`{ shouldSaveAndCreateNew: boolean, keyboardMapping: KeyboardMapping }`.

An alternative is to have two separate outputs for the two type of actions. E.g. `onSave` and `onSaveAndCreateNew`.

src/app/shared/components/notification-popup-container/notification-popup-container.component.scss
```diff
@@ -9,5 +9,5 @@
   flex-direction: column;
   gap: 8px;
 
-  z-index: 300;
+  z-index: 1001;
```
KvelaGorrrrnio, line 12: What was covering it?

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -48,6 +48,7 @@
 <sofie-button-group>
   <sofie-button (click)="cancel()" i18n>global.close.button</sofie-button>
   <sofie-spacer></sofie-spacer>
+  <sofie-button *ngIf="!keyboardMapping" (click)="save(true)" [type]="'STANDARD'" [isDisabled]="!actionTriggerForm.valid" [tooltip]="saveButtonErrorsTooltip" i18n>common.save-and-create-new</sofie-button>
```
KvelaGorrrrnio: Since we use a static value for `type`, we can remove the dynamic attribute binding so it becomes `type="STANDARD"`.

src/app/shared/components/notification-popup-container/notification-popup-container.component.scss
vivanov90: Dialog window 
Screenshot 2024-04-11 101632

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: @KvelaGorrrrnio event separate the outputs I don't see the way to separate the callback to afterClosed `method `:

 ```
 public createConfirmDialog(
    title: string,
    message: string,
    okButtonText: string,
    onOk: () => void,
    theme: DialogColorScheme = DialogColorScheme.DARK,
    severity: DialogSeverity = DialogSeverity.INFO
  ): void {
    this.open(ConfirmationDialogComponent, {
      data: {
        title: title,
        message: message,
        buttonText: {
          ok: okButtonText,
          cancel: $localize`dialog.cancel-button:Cancel`,
        },
        theme,
        severity,
      },
    })
      .afterClosed()
      .subscribe(result => {
        if (result) {
          onOk()
        }
      })
  }
```
I will change it to:
`{ shouldSaveAndCreateNew: boolean, keyboardMapping: KeyboardMapping }`

Comment (vivanov90) [reply src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts]: Done

File: src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html (hunk 3 above)
Comment (vivanov90) [reply src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html]: Fix

File: src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts [hunk 4]
```diff
@@ -126,26 +131,29 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
   }
 
   public openCreateKeyboardMapping(): void {
-    this.dialogService.openSidebarDialog(EditKeyboardMappingDialogComponent, (createdKeyboardMapping?: KeyboardMapping) => {
-      if (!createdKeyboardMapping) {
+    this.dialogService.openSidebarDialog(EditKeyboardMappingDialogComponent, (createKeyboardMapping?: EditKeyboardMapping) => {
+      if (!createKeyboardMapping || !createKeyboardMapping.keyboardMapping) {
```
KvelaGorrrrnio: `createKeyboardMapping` sounds like a function that can be called. `EditKeyboardMapping` has the same ring to it. I suggest that `EditKeyboardMapping` is renamed to `EditKeyboardMappingResponse` and the parameter is called `createdKeyboardMappingResponse`.

And since the check is to see if an `EditKeyboardMapping` was provided `|| !createKeyboardMapping.keyboardMapping` can be removed since `.keyboardMapping` is not optional in the  `EditKeyboardMapping` type. This applies in the `openEditKeyboardMapping` method as well.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -16,6 +16,11 @@ import { FormatKeyboardKeysPipe } from '../../../../shared/pipes/format-keyboard
 import { EditKeyboardMappingDialogComponent } from '../edit-keyboard-mapping-dialog/edit-keyboard-mapping-dialog.component'
 import { DialogColorScheme, DialogSeverity } from 'src/app/shared/components/confirmation-dialog/confirmation-dialog.component'
 
+export interface EditKeyboardMapping {
+  keyboardMapping: KeyboardMapping
+  shouldSaveAndCreateNew?: boolean
```
KvelaGorrrrnio: After thinking a bit, a thought popped up: Save happens in all cases. The difference that this models is whether the form should be kept open or being closed. So an alternative could be to call this `shouldClose`.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -16,6 +16,11 @@ import { FormatKeyboardKeysPipe } from '../../../../shared/pipes/format-keyboard
...
+}
```
KvelaGorrrrnio: Since this is a value that the `EditKeyboardMappingComponent` emits, it should be defined in there.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
KvelaGorrrrnio: It might indicate that the dialog framework is not flexible enough for what we want :) But for now this is fine.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
KvelaGorrrrnio, line 1: The latest changes breaks the edit functionality of the edit keyboard mapping dialog. The reason being that the `KeyboardMapping` passed to the `openEditKeyboardMapping` is treated as the `EditKeyboardMapping` type. The table component casts the entity to any when used in the template, which is why it hasn't been caught be the compiler.
 Instead of passing in the `EditKeyboardMapping` type, I suggest that we change the type of the openSidebarDialog slightly to break the binding between input and output parameters.

The `openSidebarDialog` method type could be:
```typescript
  public openSidebarDialog<T extends { new (arg: D, ...args: any): object }, D = undefined, R = undefined>(component: T, onClose: (result?: R) => void, data?: D): void
```

Here an additional type parameter has been added, to model the result type.

KvelaGorrrrnio (changes requested):

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vivanov90: Done

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vivanov90: Done

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vivanov90: Done

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vivanov90: Done

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -126,23 +127,26 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
   }
 
   public openCreateKeyboardMapping(): void {
-    this.dialogService.openSidebarDialog(EditKeyboardMappingDialogComponent, (createdKeyboardMapping?: KeyboardMapping) => {
-      if (!createdKeyboardMapping) {
+    this.dialogService.openSidebarDialog(EditKeyboardMappingDialogComponent, (createdKeyboardMappingResponse?: EditKeyboardMappingResponse) => {
+      if (!createdKeyboardMappingResponse || !createdKeyboardMappingResponse.keyboardMapping) {
```
KvelaGorrrrnio: `keyboardMapping` mapping is a required field on `EditKeyboardMappingResponse`. Thus it is not necessary to check that it is set. Hereby the expression can be changed to `!createdKeyboardMappingResponse`.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -126,23 +127,26 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
...
     })
   }
 
   public openEditKeyboardMapping(keyboardMapping: KeyboardMapping): void {
     this.dialogService.openSidebarDialog(
       EditKeyboardMappingDialogComponent,
-      (updatedKeyboardMapping?: KeyboardMapping) => {
-        if (!updatedKeyboardMapping) {
+      (updatedKeyboardMapping?: EditKeyboardMappingResponse) => {
+        if (!updatedKeyboardMapping || !updatedKeyboardMapping.keyboardMapping) {
```
KvelaGorrrrnio: `keyboardMapping` mapping is a required field on `EditKeyboardMappingResponse`. Thus it is not necessary to check that it is set. Hereby the expression can be changed to `!updatedKeyboardMappingResponse`.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -126,23 +127,26 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
...
+      }
     })
   }
 
   public openEditKeyboardMapping(keyboardMapping: KeyboardMapping): void {
     this.dialogService.openSidebarDialog(
       EditKeyboardMappingDialogComponent,
-      (updatedKeyboardMapping?: KeyboardMapping) => {
-        if (!updatedKeyboardMapping) {
+      (updatedKeyboardMapping?: EditKeyboardMappingResponse) => {
```
KvelaGorrrrnio: To have the parameter name match the type and be consistent with how similar parameters are named, `Response` should be suffixed, such that it becomes `updatedKeyboardMappingResponse`.

KvelaGorrrrnio (commented):
The branch needs to be updated from the base branch :)

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -88,16 +93,19 @@ export class EditKeyboardMappingComponent implements OnInit {
     return this.selectedAction.argument.type
   }
 
-  public save(): void {
-    const keyboardMapping: KeyboardMapping = {
-      action: {} as Tv2PartAction,
-      actionTrigger: {
-        ...this.keyboardMapping?.actionTrigger,
-        ...this.actionTriggerForm.value,
+  public save(shouldClose: boolean): void {
```
LindvedKrvang: `shouldClose` is a boolean flag which is a bad practice.
Boolean flags indicates that the method has more than one responsibility. If it didn't it wouldn't need the flag.
Instead you should create methods that explicitly states what happens. That could look something like:
```
// Don't do this:
public buildTimeline(withBlueprints: boolen): Timeline {
  ...
}

// Do this instead
public buildTimelineWithBlueprints(): Timeline {
  ...
}

public buildTimelineWithoutBlueprints(): Timeline {
  ...
}
```

src/locale/messages.da.xlf
```diff
@@ -468,6 +468,10 @@
         <target state="new">Er du sikker på at du vil afprøve denne Rundown?
         Det vil deaktivere Rundown</target>
       </trans-unit>
+      <trans-unit id="9099550788779670007" datatype="html">
+        <source>common.save-and-create-new</source>
+        <target state="new">Save and create new</target>
```
LindvedKrvang: Danish value: `Gem og opret ny` :)

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vivanov90: Fixed

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vivanov90: Fixed

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vivanov90: Fixed

src/locale/messages.da.xlf
vivanov90: Fixed

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vivanov90: Fixed

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -48,7 +48,11 @@
 <sofie-button-group>
   <sofie-button (click)="cancel()" i18n>global.close.button</sofie-button>
   <sofie-spacer></sofie-spacer>
-  <sofie-button (click)="save()" [type]="'PRIMARY'" [isDisabled]="!actionTriggerForm.valid" [tooltip]="invalidControlsErrorMessage" i18n>common.save</sofie-button>
+
+  <sofie-button *ngIf="!keyboardMapping" (click)="saveAndRemainedOpen()" type="STANDARD" [isDisabled]="!actionTriggerForm.valid" [tooltip]="invalidControlsErrorMessage" i18n
```
KvelaGorrrrnio: Minor thing, `Remained` in `saveAndRemainedOpen` should be `Remain` :)

KvelaGorrrrnio (approved):
Nice work! 💪

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
vivanov90: Fixed

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #166, opened 2024-04-12
SOF-1875 - bug fix, allows vertical scroll

vivanov90:


src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
```diff
@@ -1,3 +1,14 @@
+:host {
+  height: 100vh;
```
KvelaGorrrrnio: I would appreciate if we did not make assumptions based on the window size. Instead the size of the component should be based on the parent element, so this would become `100%`.

vivanov90: Fixed

src/app/rundown-overview/components/rundown-overview/rundown-overview.component.scss
KvelaGorrrrnio, line 1: Now that horizontal scrolling is introduced, it would be nice to have the scrollbar corner to match the scrollbar thumbs with the global scrollbar-styling:
```scss
  & *::-webkit-scrollbar-corner {
    background-color: var(--scrollbar-thumb-color);
  }
```

KvelaGorrrrnio, line 1: Is it correctly observed that the responsibility of scrolling is placed on no specific component, but targets the area that the `sofie-page` element occupies?

An observation: With the new scrolling added, the page elements exhibit some visual quirks, like in the screenshot below.
Skærmbillede 2024-04-16 kl  16 12 36

A solution is to place the responsibility (or at least page sizing) on the `PageComponent`.
One way to do this is to wrap the template in a container, e.g. `<main><!-- old page component template --></main>` and give the `main` element a minimum width of `min-content` to allow for the content to be present. As an extension, the responsibility of scrolling could be placed on the PageComponent. But let us have a chat on that :D

vivanov90: @KvelaGorrrrnio I rework the structure of the overflow and now the scrolls come from page component. In addition that we do fix the horizontal scroll without adding any new components. You can check it, is that structure is better.

LindvedKrvang (approved):
Looks good :)

Minor thing: It would be nice if you could target the PR to `staging` so our staging build also have these changes.

RasmusAlbrektsen (approved):

k-primeholding-dim (approved):

PR #174, opened 2024-04-19
SOF-1906 - Fix Importing Invalid Action In Keyboard Mapping

RasmusAlbrektsen:
We always expect an action in the keyboard mappings settings page. but and imported action that does not exists in the current system, is not handled. After talking with Jacob he told me he still wanted it to go through but with a warning notifcation and an icon showing that something is wrong. 

I haven't been able to find any issues with the fact that we have a non-existing action here, e.g. it doesn't show up on the producer shelf keyboard.


src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -227,6 +240,14 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
     )
   }
 
+  public getKeyboardMappingActionContentType(keyboardMapping: KeyboardMapping): Tv2ActionContentType {
+    return keyboardMapping.action ? keyboardMapping.action.metadata.contentType : Tv2ActionContentType.UNKNOWN
+  }
+
+  public getKeyboardMappingActionName(keyboardMapping: KeyboardMapping): string {
+    return keyboardMapping.action ? keyboardMapping.action.name : ''
+  }
```
LindvedKrvang: You should be able to use the double coalescing operator here instead of the ternary:
image

LindvedKrvang (approved):

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -227,6 +240,14 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
     )
   }
 
+  public getKeyboardMappingActionContentType(keyboardMapping: KeyboardMapping): Tv2ActionContentType {
+    return keyboardMapping.action ? keyboardMapping.action.metadata.contentType : Tv2ActionContentType.UNKNOWN
+  }
```
KvelaGorrrrnio: From reading the code this is counterintuitive, since it goes against the types of the system. `action` is a required, and always truthy, field on the `KeyboardMapping` type, but is used a potential falsy/optional value here.

I would like to have a chat about an alternative solution :)

KvelaGorrrrnio (changes requested):

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.html
```diff
@@ -27,7 +27,7 @@
                     <sofie-rounded-box>{{ getShortcutName(keyboardMapping) }}</sofie-rounded-box>
                 </td>
                 <td>
-                    <sofie-rounded-box class="{{ keyboardMapping.action.metadata.contentType | actionColorClass }}">{{ keyboardMapping.action.name }}</sofie-rounded-box>
+                    <sofie-rounded-box *ngIf="keyboardMapping.action else invalidField"  class="{{ getKeyboardMappingActionContentType(keyboardMapping) | actionColorClass }}">{{ getKeyboardMappingActionName(keyboardMapping) }} </sofie-rounded-box>
```
vivanov90: Seems like a lot of code for a single line to me

RasmusAlbrektsen: This is no longer the case after changing the approach :-)

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
RasmusAlbrektsen: This is no longer part of the solution and has been removed.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
RasmusAlbrektsen: This is no longer used since the approach has been changed.

RasmusAlbrektsen:
The approach for importing invalid actions has been changed to simply ignore them, anything related to showing the keyboard mappings without a valid action has therefore been removed.

KvelaGorrrrnio (approved):

LindvedKrvang (approved):

PR #175, opened 2024-04-19
SOF-1887 - Dont Require Name for Rundowns

RasmusAlbrektsen:
The zod validation no longer requires a name for the rundown, in the edgecase where we have a "Ghost rundown", which was attained in Staging, we should be able to delete it just like in Core.

### DIFF
--- a/src/app/core/parsers/zod-entity-parser.service.ts
+++ b/src/app/core/parsers/zod-entity-parser.service.ts
@@ -112,7 +112,7 @@ export class ZodEntityParser implements EntityParser {
 
   private readonly basicRundownParser = zod.object({
     id: zod.string().min(1),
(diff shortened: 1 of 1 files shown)


LindvedKrvang (approved):
Have you tried creating a Rundown that doesn't have a name and see that we can still load it?

KvelaGorrrrnio (approved):

k-primeholding-dim (approved):

RasmusAlbrektsen:
> Have you tried creating a Rundown that doesn't have a name and see that we can still load it?

I tested by inserting a rundown in the database with no name. I haven't been able to recreate the scenario with the iNews gateway, however.

vivanov90 (approved):

PR #177, opened 2024-04-23
Update staging from develop

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-server/pull/129


KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):
Just to make sure we all understand. This includes the static buttons configuration and will require us to do the fix in Post Deploy Config when deploying to our staging environment.

PR #182, opened 2024-04-24
SOF-1904- Show Style Variant Ingest Update

RasmusAlbrektsen:
The Show Style Variant in the rundown header now changes upon rundown update events to follow ingest.

### DIFF
--- a/src/app/core/services/show-style-variant-state.service.ts
+++ b/src/app/core/services/show-style-variant-state.service.ts
@@ -3,13 +3,40 @@ import { BehaviorSubject, lastValueFrom, Observable } from 'rxjs'
 import { ShowStyleVariant } from '../models/show-style-variant'
 import { ShowStyleVariantService } from '../abstractions/show-style-variant.service'
 import { EventSubscription } from '../../event-system/abstractions/event-observer.service'
+import { RundownEventObserver } from './rundown-event-observer.service'
+import { RundownUpdatedEvent } from '../models/rundown-event'
+import {Logger} from "../abstractions/logger.service";
 
 @Injectable()
 export class ShowStyleVariantStateService implements OnDestroy {
   private readonly showStyleVariantSubjects: Map<string, BehaviorSubject<ShowStyleVariant>> = new Map()
   private readonly eventSubscriptions: EventSubscription[]
+  private readonly logger: Logger
 
-  constructor(private readonly showStyleVariantService: ShowStyleVariantService) {}
+  constructor(
+    private readonly showStyleVariantService: ShowStyleVariantService,
+    private readonly rundownEventObserver: RundownEventObserver,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('ShowStyleVariantStateService')
+    this.subscribeToRundownEvents()
+  }
+
+  private subscribeToRundownEvents(): void {
+    this.rundownEventObserver.subscribeToRundownUpdates(this.updateShowStyleVariantFromEvent.bind(this))
+  }
+
+  private updateShowStyleVariantFromEvent(event: RundownUpdatedEvent): void {
+    this.getCleanShowStyleVariantSubject(event.rundownId)
+      .then(newVariant => {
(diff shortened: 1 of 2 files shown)


src/app/core/services/show-style-variant-state.service.ts
```diff
@@ -3,13 +3,40 @@ import { BehaviorSubject, lastValueFrom, Observable } from 'rxjs'
...
+    private readonly showStyleVariantService: ShowStyleVariantService,
+    private readonly rundownEventObserver: RundownEventObserver,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('ShowStyleVariantStateService')
+    this.subscribeToRundownEvents()
+  }
+
+  private subscribeToRundownEvents(): void {
+    this.rundownEventObserver.subscribeToRundownUpdates(this.updateShowStyleVariantFromEvent.bind(this))
```
KvelaGorrrrnio: This should be assigned to the `eventSubscriptions`, such that we can unsubscribe on destroy.

KvelaGorrrrnio (approved):

src/app/core/services/show-style-variant-state.service.ts
LindvedKrvang: This is not going to be part of this PR, but it would be nice if we could refactor away from returning `EventSupbscription` so that we could simply do `.pipe(takeUntil(this.destroySubject))`

src/app/core/services/show-style-variant-state.service.ts
```diff
@@ -3,13 +3,40 @@ import { BehaviorSubject, lastValueFrom, Observable } from 'rxjs'
...
+  }
+
+  private updateShowStyleVariantFromEvent(event: RundownUpdatedEvent): void {
```
LindvedKrvang: We should really stop using the `fromEvent` pattern that has spread throughout the frontend...
 
The "from event" references something outside the scope of this method and the only thing that you actually need from `RundownUpdatedEvent` is the `rundownId`. And since you have "bound" the method declaration to something outside of the scope of the method, you limit the flexibility in where it could be used. Right now, it can _only_ be used when you listen for RundownEvents, which is quite inflexible and goes against the OpenClose principle.

By making the signature something like `updateShowStyleVariantForRundown(rundownId: string)` you have opened this method up to be used by a plethora of places and you can now extend your program to use this method somewhere else, without having to first modify it (the basis of the OpenClose principle).

src/app/core/services/show-style-variant-state.service.ts
```diff
@@ -3,13 +3,40 @@ import { BehaviorSubject, lastValueFrom, Observable } from 'rxjs'
...
+  constructor(
+    private readonly showStyleVariantService: ShowStyleVariantService,
+    private readonly rundownEventObserver: RundownEventObserver,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('ShowStyleVariantStateService')
+    this.subscribeToRundownEvents()
+  }
+
+  private subscribeToRundownEvents(): void {
```
LindvedKrvang: It seems a little overkill to make entire method that says it subscibes to multiple RundownEvents and then it only subscribes to one.

With the current state of this file, it would be fine to have the one-liner, that is the body of this method, in the constructor.

LindvedKrvang (approved):
The approve is based on the assumption that you unsubscribe to the RundownEvent, as Anders mentioned :)

src/app/core/services/show-style-variant-state.service.ts
RasmusAlbrektsen: For now it has been added to `eventSubscriptions`.

src/app/core/services/show-style-variant-state.service.ts
RasmusAlbrektsen: I see you point, and i have updated the method to your suggestion (even your naming though that was accidental).

src/app/core/services/show-style-variant-state.service.ts
RasmusAlbrektsen: It has been removed and now happens in the constructor!

PR #183, opened 2024-04-25
SOF-1905 Dialog should appear in the middle of one screen when using multiple monitors

LindvedKrvang:
When stretched vertically on multiple monitors we now show dialogs approximately in the middle of the top monitor

### DIFF
--- a/src/app/shared/services/dialog.service.ts
+++ b/src/app/shared/services/dialog.service.ts
@@ -1,8 +1,10 @@
 import { ComponentType } from '@angular/cdk/overlay'
 import { Injectable } from '@angular/core'
-import { MatDialog, MatDialogConfig, MatDialogRef } from '@angular/material/dialog'
+import { DialogPosition, MatDialog, MatDialogConfig, MatDialogRef } from '@angular/material/dialog'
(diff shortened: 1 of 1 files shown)


RasmusAlbrektsen (approved):

src/app/shared/services/dialog.service.ts
```diff
@@ -12,9 +14,26 @@ export class DialogService {
     component: ComponentType<StronglyTypedDialog<DialogData, DialogResult>>,
     config?: MatDialogConfig<DialogData>
   ): MatDialogRef<StronglyTypedDialog<DialogData, DialogResult>, DialogResult> {
+    if (!config) {
+      config = {}
+    }
```
KvelaGorrrrnio: Just a suggestion for an alternative syntax with nullish coalescing assignment: `config ??= {}`.
This could probably be set directly for the `config` parameter, such that the parameter becomes `config: MatDialogConfig<DialogData> = {}`.

KvelaGorrrrnio (approved):

src/app/shared/services/dialog.service.ts
LindvedKrvang: I have updated it to your last suggestion.

PR #187, opened 2024-04-26
SOF-1922/Handle Replaced Piece Events

RasmusAlbrektsen:
This depends on: https://github.com/tv2/sofie-server/pull/138

With the newly created `PieceReplaced` event we can correctly handle the intention of the backend when building our DVE pieces. 

Another issue that was made apparent was how we lose our original planned DVE piece when we go back and set it as next after inserting inputs, this is, however, not due to these PR's but also exists on staging in it self. I will look further into this as a separate task.


LindvedKrvang (approved):

src/app/core/services/models/part-entity.service.ts
```diff
@@ -102,4 +102,15 @@ export class PartEntityService {
...
+
+  public replacePiece(part: Part, replacedPieceId: string, newPiece: Piece): Part {
+    const replacedPieceIndex: number = part.pieces.findIndex(piece => piece.id === replacedPieceId)
+    if (replacedPieceIndex >= 0) {
+      part.pieces.splice(replacedPieceIndex, 1)
+    }
+    return {
+      ...part,
+      pieces: [...part.pieces, newPiece],
+    }
```
KvelaGorrrrnio: `Array.splice` mutates the referenced object, which we should avoid. An alternative to this is to use `Array.toSpliced` or use `Array.filter`. E.g.:

```typescript
return {
  ..part,
  pieces: [
    ...part.pieces.filter(piece => piece.id !== replacedPiece.id),
    newPiece
  ]
}
```
or
```typescript
const replacedPieceIndex: number = part.pieces.findIndex(piece => piece.id === replacedPiece.id)
if (replacedPieceIndex < 0) {
  return part
}
return {
  ...part,
  pieces: [...part.pieces.toSpliced(replacedPieceIndex, 1), newPiece],
}
```

src/app/core/services/rundown-state.service.ts
```diff
@@ -364,6 +366,15 @@ export class RundownStateService implements OnDestroy {
     rundownSubject.next(rundownWithPiece)
   }
 
+  private replacePieceFromEvent(event: RundownPieceReplacedEvent): void {
+    const rundownSubject = this.getRundownSubject(event.rundownId)
+    if (!rundownSubject || !rundownSubject.value) {
```
KvelaGorrrrnio, line 371: Suggestion: Optional chaining can be used to check the existence of the value `!rundownSubject?.value`.

KvelaGorrrrnio (approved):

src/app/core/services/models/part-entity.service.ts
RasmusAlbrektsen: A very good point and an elegant solution. I made use of filtering! Thanks!

src/app/core/services/rundown-state.service.ts
RasmusAlbrektsen: I personally just like how this reads so long as this is all that there is at least. If you really want it changed i of course can do that as well as the other places we use it @KvelaGorrrrnio!

PR #188, opened 2024-04-26
SOF-1917 Now listens for ACTIONS_UPDATED event for updating Actions

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-server/pull/139

### DIFF
--- a/src/app/shared/services/action-state.service.ts
+++ b/src/app/shared/services/action-state.service.ts
@@ -5,10 +5,10 @@ import { Injectable } from '@angular/core'
 import { ConnectionStatusObserver } from '../../core/services/connection-status-observer.service'
 import { Logger } from '../../core/abstractions/logger.service'
 import { EventSubscription } from '../../event-system/abstractions/event-observer.service'
-import { RundownEventObserver } from '../../core/services/rundown-event-observer.service'
-import { RundownActivatedEvent, RundownDeactivatedEvent, RundownRehearseEvent } from '../../core/models/rundown-event'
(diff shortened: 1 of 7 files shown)


RasmusAlbrektsen (approved):

src/app/shared/services/action-state.service.ts
```diff
@@ -74,16 +72,16 @@ export class ActionStateService {
...
-  }
+  private updateActions(actions: Action[], rundownId?: string): void {
+    const id: string = rundownId ?? SYSTEM_ACTIONS_ID
 
-  private onRundownRehearse(event: RundownRehearseEvent): void {
-    this.resetActionsSubject(event.rundownId)
-  }
+    const actionsSubject: BehaviorSubject<Action[]> | undefined = this.actionsSubjects.get(id)
+    if (!actionsSubject) {
+      this.actionsSubjects.set(id, new BehaviorSubject(actions))
```
KvelaGorrrrnio, line 80: If there is no action subject for the id, it means that there are no subscribers for that id. Thus we shouldn't store the actions.

KvelaGorrrrnio (approved):

src/app/shared/services/action-state.service.ts
LindvedKrvang: If we get an ActionUpdatedEvent that means that there are Actions for that Rundown. By storing them immediatly we don't have to fetch them from the backend once a Rundown subscribes to the them,

PR #190, opened 2024-04-30
Sof 1924/introduce content placeholder action

LindvedKrvang:
The ActionStateService now creates a GraphicsContentPlaceholderAction that can be mapped to keyboard bindings.
When finding resolving the keyboard bindings for the PlaceholderActions it will go search for the current Action to bind it to.

### DIFF
--- a/src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
+++ b/src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
@@ -16,7 +16,7 @@ import { StyledKeyBinding } from '../../keyboard/value-objects/styled-key-bindin
 import { ActionService } from '../../shared/abstractions/action.service'
 import { KeyboardTriggerData } from 'src/app/shared/models/keyboard-trigger-data'
 import { RundownMode } from '../../core/enums/rundown-mode'
-import { PlaceholderActionType } from '../../shared/models/action-type'
+import { PlaceholderActionScope, PlaceholderActionType } from '../../shared/models/action-type'
 import { Segment } from '../../core/models/segment'
 
 const CAMERA_COLOR: string = 'var(--tv2-camera-color)'
@@ -106,18 +106,30 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
   }
 
   private findActionForContentPlaceholderAction(placeholderAction: Tv2ContentPlaceholderAction, actionTrigger: ActionTrigger<KeyboardTriggerData>): Tv2Action | undefined {
-    const onAirSegment: Segment | undefined = this.rundown?.segments.find(segment => segment.isOnAir)
-    if (!onAirSegment) {
+    const filterCallback: (action: Tv2Action) => boolean = this.getActionFilterCallbackForPlaceholderAction(placeholderAction)
+
+    const actionsForSegmentToSearch: Tv2Action[] = this.actions.filter(filterCallback).sort((a, b) => a.rank - b.rank)
+    const indexToSelect: number = actionTrigger.data.actionArguments as number // We know this is a number because of Tv2ContentPlaceholderAction.
+    if (indexToSelect > actionsForSegmentToSearch.length) {
       return
     }
-    const actionsForOnAirSegment: Tv2Action[] = this.actions
-      .filter(action => Math.floor(action.rank) === onAirSegment.rank && placeholderAction.metadata.allowedContentTypes.includes(action.metadata.contentType))
-      .sort((a, b) => a.rank - b.rank)
-    const nthToSelect: number = actionTrigger.data.actionArguments as number // We know this is a number because of Tv2ContentPlaceholderAction.
-    if (nthToSelect > actionsForOnAirSegment.length) {
-      return
+    return actionsForSegmentToSearch[indexToSelect - 1] // indexToSelect is one-indexed, so we need to subtract one to get the correct index.
+  }
+
+  private getActionFilterCallbackForPlaceholderAction(placeholderAction: Tv2ContentPlaceholderAction): (action: Tv2Action) => boolean {
+    switch (placeholderAction.metadata.scope) {
+      case PlaceholderActionScope.ON_AIR_SEGMENT: {
+        const onAirSegment: Segment | undefined = this.rundown?.segments.find(segment => segment.isOnAir)
+        if (!onAirSegment) {
+          break
+        }
+        return action => Math.floor(action.rank) === onAirSegment.rank && placeholderAction.metadata.allowedContentTypes.includes(action.metadata.contentType)
+      }
+      default: {
+        this.logger.warn(`PlaceholderScope ${placeholderAction.metadata.scope} is not supported.`)
+      }
     }
-    return actionsForOnAirSegment[nthToSelect - 1] // nthToSelect is one-indexed, so we need to subtract one to get the correct index.
+    return () => false
   }
 
   private updateActionTriggerKeybindings(): void {
--- a/src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
+++ b/src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
@@ -10,12 +10,14 @@ import { RundownStateService } from '../../core/services/rundown-state.service'
 import { Rundown } from '../../core/models/rundown'
 import { Tv2ActionParser } from '../../shared/abstractions/tv2-action-parser.service'
 import { SystemKeyBindingFactory } from '../factories/system-key-binding-factory.service'
-import { Tv2Action, Tv2ActionContentType } from '../../shared/models/tv2-action'
+import { Tv2Action, Tv2ActionContentType, Tv2ContentPlaceholderAction } from '../../shared/models/tv2-action'
 import { Logger } from '../../core/abstractions/logger.service'
 import { StyledKeyBinding } from '../../keyboard/value-objects/styled-key-binding'
 import { ActionService } from '../../shared/abstractions/action.service'
 import { KeyboardTriggerData } from 'src/app/shared/models/keyboard-trigger-data'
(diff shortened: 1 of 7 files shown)


src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -89,14 +91,35 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
...
       }
-      this.actionTriggersWithAction.set(actionTrigger, actionForTrigger)
+      if (action.type === PlaceholderActionType.CONTENT) {
+        action = this.findActionForContentPlaceholderAction(action as Tv2ContentPlaceholderAction, actionTrigger)
+      }
+      if (!action) {
+        return
+      }
+      this.actionTriggersWithAction.set(actionTrigger, action)
     })
```
KvelaGorrrrnio: Instead of having the logic for determining when to resolve the action and the logic for updating in the same method in the same method, they could be separated, which would lead to an immutable flow for making the decisions. E.g.
```typescript
private getResolvedAction(actionTrigger: ActionTrigger<KeyboardTriggerData>): Tv2Action | undefined {
  const action: Tv2Action | undefined = this.actions.find(action => action.id === actionTrigger.actionId)
  if (action?.type !== PlaceholderActionType.CONTENT) {
    return
  }
  return this.findActionForContentPlaceholderAction(action as Tv2ContentPlaceholderAction, actionTrigger)
}
```

Suggestion: `actionTriggersWithAction` is updated via side-effects in the forEach. An alternative solution could be  to use reduce to generate the entries for a new Map.
```typescript
this.actionTriggersWithAction = New Map(
  this.actionTriggers.reduce((entries, actionTrigger) => {
    const action: Tv2Action | undefined = this.getActionForActionTrigger(actionTrigger)
    return action ? [...entries, [actionTrigger, action]] : entries
  }, [])
)
```

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -89,14 +91,35 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
...
 
+  private findActionForContentPlaceholderAction(placeholderAction: Tv2ContentPlaceholderAction, actionTrigger: ActionTrigger<KeyboardTriggerData>): Tv2Action | undefined {
+    const onAirSegment: Segment | undefined = this.rundown?.segments.find(segment => segment.isOnAir)
+    if (!onAirSegment) {
+      return
+    }
+    const actionsForOnAirSegment: Tv2Action[] = this.actions
+      .filter(action => Math.floor(action.rank) === onAirSegment.rank && placeholderAction.metadata.allowedContentTypes.includes(action.metadata.contentType))
+      .sort((a, b) => a.rank - b.rank)
+    const nthToSelect: number = actionTrigger.data.actionArguments as number // We know this is a number because of Tv2ContentPlaceholderAction.
```
KvelaGorrrrnio: The action argument name is `indexToSelect`. It would make sense to use that name here as well.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -89,14 +91,35 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
...
     })
   }
 
+  private findActionForContentPlaceholderAction(placeholderAction: Tv2ContentPlaceholderAction, actionTrigger: ActionTrigger<KeyboardTriggerData>): Tv2Action | undefined {
+    const onAirSegment: Segment | undefined = this.rundown?.segments.find(segment => segment.isOnAir)
+    if (!onAirSegment) {
+      return
+    }
+    const actionsForOnAirSegment: Tv2Action[] = this.actions
+      .filter(action => Math.floor(action.rank) === onAirSegment.rank && placeholderAction.metadata.allowedContentTypes.includes(action.metadata.contentType))
```
KvelaGorrrrnio: Finding the actions related to the on air segment should be based on the `scope` attribute for the placeholder.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
LindvedKrvang: I went with `nthToSelect` because index is useally zero-indexed, so it was a way to beware that this wasn't zero-indexed. I can change it to `indexToSelect` if preferred.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
LindvedKrvang: I'm not sure I follow your first sentence. The logic for determing when to resolve is not in this method. When calling `mapActionTriggersToActions` it has _already_ been decided that now we should "resolve" them and this method has nothing to do with that logic.

For your `.reduce` suggestion: did you try it? I couldn't get it to compile.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
LindvedKrvang: Updated. Introduced `getActionFilterCallbackForPlaceholderAction()`

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -89,14 +91,47 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
...
 
     this.actionTriggers.forEach(actionTrigger => {
-      const actionForTrigger: Tv2Action | undefined = this.actions.find(action => action.id === actionTrigger.actionId)
-      if (!actionForTrigger) {
+      let action: Tv2Action | undefined = this.actions.find(action => action.id === actionTrigger.actionId)
+      if (!action) {
         return
       }
-      this.actionTriggersWithAction.set(actionTrigger, actionForTrigger)
+      if (action.type === PlaceholderActionType.CONTENT) {
```
RasmusAlbrektsen: Would it be an issue to instead say: 
```
if (action?.type === PlaceholderActionType.CONTENT) 
````
Then we can get rid of the first action check.

File: src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts (hunk 1 above)
Comment (KvelaGorrrrnio) [reply src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts]: The code from line 92 to 103, determines the action and whether it should be resolved based on the action type. The change from `const actionForTrigger` to `let actionForTrigger` is an indicator that some conditional logic is added to the method. By moving the computation to a separate method, the decision making that the introduced logic is responsible for is now separated from this method and we can keep it as an immutable computation.

Sorry, the reduce was not tested – my bad. I have added code that compiles below. It is just a suggestion for an immutable data change flow.
```typescript
  private actionTriggersWithAction: Map<ActionTrigger<KeyboardTriggerData>, Tv2Action> = new Map()

   // ...

    this.actionTriggersWithAction = new Map(
      this.actionTriggers.reduce((entries, actionTrigger) => {
        const action: Tv2Action | undefined = this.getResolvedAction(actionTrigger)
        return action ? [...entries, [actionTrigger, action]] : entries
      }, [] as [ActionTrigger<KeyboardTriggerData>, Tv2Action][])
    )
```

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -89,14 +91,47 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
...
+    switch (placeholderAction.metadata.scope) {
+      case PlaceholderActionScope.ON_AIR_SEGMENT: {
+        const onAirSegment: Segment | undefined = this.rundown?.segments.find(segment => segment.isOnAir)
+        if (!onAirSegment) {
+          break
+        }
+        return action => Math.floor(action.rank) === onAirSegment.rank && placeholderAction.metadata.allowedContentTypes.includes(action.metadata.contentType)
+      }
+      default: {
+        this.logger.warn(`PlaceholderScope ${placeholderAction.metadata.scope} is not supported.`)
```
RasmusAlbrektsen: Should `PlaceholderScope` be two words (with "scope" instead) or is it in reference to the `PlaceholderActionScope`?

RasmusAlbrektsen (approved):

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
KvelaGorrrrnio: I missed the point of it being a 1-based index. I don't think that _nth_ exhibit that differently than _index_. So the added comment is fine enough for me :)

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
LindvedKrvang: I have extracted the logic for finding the Action to it's own separate method.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
LindvedKrvang: I guees not. It's gone with some of the changes from Anders suggestions.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
LindvedKrvang: It is in reference to `PlaceholderActionScope`, but I think there is a sound argument in having it as two words for the log message.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
LindvedKrvang: Too late. It's been changed to `indexToSelect` ;)

KvelaGorrrrnio (approved):

PR #193, opened 2024-05-01
SOF 1909/Show Segment numbers

RasmusAlbrektsen:
This PR handles the `referenceTag` put forth from Sofie Server in PR: https://github.com/tv2/sofie-server/pull/143. It looks like this:
image

### DIFF
--- a/src/app/rundown-view/components/segment/segment.component.html
+++ b/src/app/rundown-view/components/segment/segment.component.html
@@ -4,14 +4,9 @@
   [class.c-segment--has-remote-piece]="hasRemotePiece"
 >
   <div [cdkContextMenuTriggerFor]="contextMenu" class="c-segment__header">
-    <div class="c-segment__header-toolbar"
-      [class.c-segment__header-toolbar--unsynced]="segment.isUnsynced"
-    >
+    <div class="c-segment__header-toolbar">
       <span class="segment-reference-tag">
-        <span *ngIf="!segment.isUnsynced else unsynced">{{segment.referenceTag}}</span>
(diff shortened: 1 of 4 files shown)


LindvedKrvang (approved):

src/app/rundown-view/components/segment/segment.component.html
```diff
@@ -7,8 +7,11 @@
     <div class="c-segment__header-toolbar"
       [class.c-segment__header-toolbar--unsynced]="segment.isUnsynced"
     >
-      <span class="segment-identifier">
-        <ng-container *ngIf="segment.isUnsynced">UNSYNCED</ng-container>
+      <span class="segment-reference-tag">
+        <span *ngIf="!segment.isUnsynced else unsynced">{{segment.referenceTag}}</span>
```
KvelaGorrrrnio: The `span` can be a `ng-container` to avoid having the extra element.

KvelaGorrrrnio: Shouldn't the reference tag and unsynced information be available at the same time if both are present?

RasmusAlbrektsen:
This is now the look of the unsynced segments:
image

src/app/rundown-view/components/segment/segment.component.html
RasmusAlbrektsen: It is with the new design!

RasmusAlbrektsen: No longer there after the changes.

KvelaGorrrrnio (approved):

PR #194, opened 2024-05-01
SOF-1899 Show thumbnail for VideoClipActions

LindvedKrvang:
Now shows thumbnails for VideoClipActions in ActionPanels

### DIFF
--- a/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
+++ b/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
@@ -1,2 +1,3 @@
-<div #title class="pt-1 ps-2 pe-2">{{ action.name }}</div>
-<img *ngIf="isVideoClipAction() && videoClipThumbnailUrl" [ngSrc]="videoClipThumbnailUrl" [width]="getThumbnailWidth()" [height]="getThumbnailHeight()" [alt]="action.name">
+<span class="pt-1 ps-2 pe-2">{{ action.name }}</span>
+<div #titleContainer class="titleContainer"></div>
+<img *ngIf="isVideoClipAction() && videoClipThumbnailUrl" [ngSrc]="videoClipThumbnailUrl" [width]="getThumbnailWidth()" [height]="getThumbnailHeight()" [alt]="action.name" top>
--- a/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
+++ b/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
@@ -1,1 +1,2 @@
-{{ action.name }}
+<div #title class="pt-1 ps-2 pe-2">{{ action.name }}</div>
+<img *ngIf="isVideoClipAction() && videoClipThumbnailUrl" [ngSrc]="videoClipThumbnailUrl" [width]="getThumbnailWidth()" [height]="getThumbnailHeight()" [alt]="action.name">
--- a/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
+++ b/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
@@ -1,3 +1,2 @@
-<span class="pt-1 ps-2 pe-2">{{ action.name }}</span>
-<div #titleContainer class="titleContainer"></div>
+<div #title class="title pt-1 ps-2 pe-2">{{ action.name }}</div>
 <img *ngIf="isVideoClipAction() && videoClipThumbnailUrl" [ngSrc]="videoClipThumbnailUrl" [width]="getThumbnailWidth()" [height]="getThumbnailHeight()" [alt]="action.name" top>
--- a/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss
+++ b/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss
@@ -1,7 +1,6 @@
 :host {
   display: inline-block;
   background-color: var(--dark-gray-color);
-  padding: 14px;
 
   width: 130px;
   height: 100px;
--- a/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss
+++ b/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss
@@ -1,4 +1,5 @@
 :host {
+  position: relative;
   display: inline-block;
   background-color: var(--dark-gray-color);
 
@@ -17,6 +18,11 @@
   cursor: pointer;
   transition: opacity 0.1s;
   overflow: hidden;
+
+  span {
+    position: absolute;
+    z-index: 1;
+  }
 
   &.camera {
     background-color: var(--tv2-camera-color);
@@ -50,5 +56,9 @@
 
   &:active {
     opacity: 0.7;
+  }
+
+  .titleContainer {
+    height: 25px;
   }
 }
--- a/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss
+++ b/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss
@@ -1,5 +1,4 @@
 :host {
-  position: relative;
   display: inline-block;
   background-color: var(--dark-gray-color);
 
@@ -19,8 +18,9 @@
   transition: opacity 0.1s;
   overflow: hidden;
 
-  span {
-    position: absolute;
+  .title {
+    position: relative;
+    height: 25px;
     z-index: 1;
   }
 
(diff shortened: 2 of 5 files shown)


src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
```diff
@@ -1,16 +1,60 @@
-import { ChangeDetectionStrategy, Component, HostBinding, Input } from '@angular/core'
-import { Tv2Action } from '../../../shared/models/tv2-action'
+import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ElementRef, HostBinding, Input, OnInit, ViewChild } from '@angular/core'
+import { Tv2Action, Tv2ActionContentType, Tv2VideoClipAction } from '../../../shared/models/tv2-action'
+import { ConfigurationService } from '../../../shared/services/configuration.service'
+import { StudioConfiguration } from '../../../shared/models/studio-configuration'
+
+const ASPECT_RATIO: number = 1.77777777
```
KvelaGorrrrnio: To make it easier for others to know what the ratio is, the expression could be changed to the ratio fraction `const ASPECT_RATIO: number = 16 / 9`.

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
```diff
@@ -1 +1,2 @@
-{{ action.name }}
+<div #title class="pt-1 ps-2 pe-2">{{ action.name }}</div>
+<img *ngIf="isVideoClipAction() && videoClipThumbnailUrl" [ngSrc]="videoClipThumbnailUrl" [width]="getThumbnailWidth()" [height]="getThumbnailHeight()" [alt]="action.name">
```
KvelaGorrrrnio: Couldn't the condition be reduced to `videoClipThumbnailUrl` since it is only set, when it is a video clip action?

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
```diff
@@ -1,16 +1,60 @@
...
 })
-export class Tv2ActionCardComponent {
+export class Tv2ActionCardComponent implements OnInit {
   @Input()
   public action: Tv2Action
 
+  @ViewChild('title')
+  public titleRef: ElementRef
+
+  public videoClipThumbnailUrl: string = ''
```
KvelaGorrrrnio, line 22: Wouldn't `thumbnailUrl` suffice?

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
```diff
@@ -1,16 +1,60 @@
...
+    this.changeDetectorRef.detectChanges()
+  }
+
+  public getThumbnailHeight(): number {
+    return this.elementRef.nativeElement.offsetHeight - this.titleRef.nativeElement.offsetHeight
+  }
+
+  public getThumbnailWidth(): number {
+    return this.getThumbnailHeight() * ASPECT_RATIO
+  }
```
KvelaGorrrrnio, line 56: An alternative to programmatically computing the size when rendering is to use the CSS attribute `aspect-ratio: 16 / 9;` for the image element.

RasmusAlbrektsen (commented):
Looks good, but when the action names are a bit longer it can be difficult to see the thumbnail. Maybe we just let the title go over the thumbnail?
image

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
LindvedKrvang: Yes, it could.
I do however like the explicitness that comes with `isVideoClipAction()` that easily shows that it most be a VideoClipAction. Only using `videoClipThumbnail` would make it implicit that it must be a VideoClipAction.

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
LindvedKrvang: I considered it, but since this component shows other Actions than just VideoClipActions and since `videoClipThumbnailUrl` is only used for VideoClips, I decided to include `videoClip` in the name so you wouldn't think that it would also work for a GraphicsAction.

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
LindvedKrvang: I see your point, but in this case that won't do since I'am using `ngSrc` to render the image. `ngSrc` requires that we set the `width` and `height` attributes on the element:
https://angular.io/guide/image-directive
https://angular.io/api/common/NgOptimizedImage

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
LindvedKrvang: Ahh, that's where 1.77777 comes from. I just took what the console told me to use :p 

It's been updated.

LindvedKrvang:
> Looks good, but when the action names are a bit longer it can be difficult to see the thumbnail. Maybe we just let the title go over the thumbnail? image

Good catch. I did not test with longer file names.
I will come up with a solution :)

LindvedKrvang:
VideoClip Actions with long names no longer affect the size of the thumbnail
image

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
```diff
@@ -1 +1,3 @@
-{{ action.name }}
+<span class="pt-1 ps-2 pe-2">{{ action.name }}</span>
+<div #titleContainer class="titleContainer"></div>
+<img *ngIf="isVideoClipAction() && videoClipThumbnailUrl" [ngSrc]="videoClipThumbnailUrl" [width]="getThumbnailWidth()" [height]="getThumbnailHeight()" [alt]="action.name" top>
```
LindvedKrvang, line 2: The `top` roguishly sneaked its way in. It has been removed in https://github.com/tv2/ng-sofie/pull/195

RasmusAlbrektsen (approved):

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
KvelaGorrrrnio: Personally, I think that the explicitness here can lead to confusion about when the `videoClipThumbnailUrl` is available and adds unnecessary tokens that a reader needs to mentally digest.

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss
```diff
@@ -52,4 +57,8 @@
   &:active {
     opacity: 0.7;
   }
+
+  .titleContainer {
```
KvelaGorrrrnio: We use kebab-case for CSS classes: `title-container`.

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html
```diff
@@ -1 +1,3 @@
-{{ action.name }}
+<span class="pt-1 ps-2 pe-2">{{ action.name }}</span>
+<div #titleContainer class="titleContainer"></div>
```
KvelaGorrrrnio: Suggestion: The newly introduced element can be avoided by making the host element display as inline-flex and use z-index of 1 with a visible vertical overflow to let the text overflow over the image.
So the HTML would be
```html
<div #titleContainer class="title pt-1 ps-2 pe-2">{{ action.name }}</div>
<img *ngIf="isVideoClipAction() && videoClipThumbnailUrl" [ngSrc]="videoClipThumbnailUrl" [width]="getThumbnailWidth()" [height]="getThumbnailHeight()" [alt]="action.name" top>
```
and the SCSS
```scss
:host {
  display: inline-flex;
  flex-direction: column;

  .title {
    height: 25px;
    overflow-y: visible;
    z-index: 1;
  }
}

**Review by KvelaGorrrrnio — APPROVED** [review]

File: src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss (hunk 6 above)
Comment (LindvedKrvang) [reply src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.scss]: Good catch.
This has been changed to just `title` after implementing your other comment.

File: src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html (hunk 7 above)
Comment (LindvedKrvang) [reply src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.html]: The newly introduced element has been removed again.
However, I changed `overflow-y: visible` with `position: relative` since `z-index` wouldn't be applied without it:
```
.title {
  position: relative;
  height: 25px;
  z-index: 1;
}
```

Sidenote: I might be wrong here, but `flex-direction` only works if you are using `display: flex`?

PR #195, opened 2024-05-01
SOF-1900 VideoClipActions in the ActionPanels now show Hoverscrub

LindvedKrvang:
Note: Dependent on https://github.com/tv2/ng-sofie/pull/194
Update: All dependent PRs has been merged.

VideoClipActions in ActionPanels now show Hoversrub.

I move the VideoClipHoverscrubComponent to the Shared module since it is now used in more than one module.


RasmusAlbrektsen:
This seems to break with the issue i mention in: https://github.com/tv2/ng-sofie/pull/194#pullrequestreview-2033507814
This should show around 46 seconds at the edge but it can't extend that far in the scrub, you can't see it but i'm holding the cursor at the very right edge of the thumbnail and if i move any more the scrub dissapears. 
image

image

Again we should just let the title go over the thumbnail and i think everything will be fine.

LindvedKrvang:
> This seems to break with the issue i mention in: [#194 (review)](https://github.com/tv2/ng-sofie/pull/194#pullrequestreview-2033507814) This should show around 46 seconds at the edge but it can't extend that far in the scrub, you can't see it but i'm holding the cursor at the very right edge of the thumbnail and if i move any more the scrub dissapears. image
> 
> image
> 
> Again we should just let the title go over the thumbnail and i think everything will be fine.

This has been solved with the fix in SOF-1899

RasmusAlbrektsen (approved):
Looks good! Works perfectly for me as well!

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
```diff
@@ -12,28 +16,39 @@ const THUMBNAIL_URL_PART: string = '/media/thumbnail/'
...
-      })
+      this.configurationService
+        .getStudioConfiguration()
+        .pipe(takeUntil(this.unsubscribeSubject))
+        .subscribe(studioConfiguration => {
+          this.studioConfiguration = studioConfiguration
+          this.updateVideoClipThumbnailUrl()
+        })
+
+      this.subscribeToMedia()
```
KvelaGorrrrnio, line 51: It would make sense to extract the studio configuration subscription to a separate method, to keep it consistent with how it is done for the media subscription.

And a note: The check for whether it is a video clip action is performed twice fo the media subscription.

KvelaGorrrrnio (approved):

PR #196, opened 2024-05-01
SOF-1901 Show Media unavailble in ActionPanels

LindvedKrvang:
Note: Is dependent on https://github.com/tv2/ng-sofie/pull/195
VideoClipActions in ActionPanels now shows if media is unavailable

### DIFF
--- a/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
+++ b/src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
@@ -89,11 +89,7 @@ export class Tv2ActionCardComponent implements OnInit, OnDestroy {
   }
 
   @HostBinding('class.missing-source')
-  public get showMissingMediaSource(): boolean {
-    return this.shouldShowMediaUnavailable()
-  }
-
-  private shouldShowMediaUnavailable(): boolean {
+  public get isMediaMissing(): boolean {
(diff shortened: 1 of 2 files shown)


RasmusAlbrektsen (approved):

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
```diff
@@ -86,6 +88,15 @@ export class Tv2ActionCardComponent implements OnInit, OnDestroy {
...
   }
 
+  @HostBinding('class.missing-source')
+  public get showMissingMediaSource(): boolean {
+    return this.shouldShowMediaUnavailable()
+  }
+
+  private shouldShowMediaUnavailable(): boolean {
+    return this.isVideoClipAction() && !this.media
+  }
```
KvelaGorrrrnio: `showMissingMediaSource` sounds like an action to perform. It would be nice if the name exhibited that it is a boolean query. A suggestion could be to merge these two methods into the getter that the host binding uses.

KvelaGorrrrnio (approved):

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
LindvedKrvang: Updated

PR #201, opened 2024-05-14
Sof 1979/show invalid segments as invalid

LindvedKrvang:
Note: This PR is dependent on https://github.com/tv2/sofie-blueprints-inews/pull/243 and https://github.com/tv2/sofie-server/pull/152
Update: All dependent PRs have been merged.

Introduce `invalidity` on Segments. Segments that are invalid are not able to be taken. They are also shown differently to indicate that they are invalid.

Also in this PR:
Renamed the CSS classes in `segment.component.scss` to not follow the BEMIT convention.

### DIFF
--- a/src/app/core/models/invalidity.ts
+++ b/src/app/core/models/invalidity.ts
@@ -1,0 +1,3 @@
+export interface Invalidity {
+  reason: string
+}
--- a/src/app/core/models/invalidity.ts
+++ b/src/app/core/models/invalidity.ts
@@ -1,3 +1,3 @@
 export interface Invalidity {
-  reason: string
+  readonly reason: string
 }
--- a/src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
+++ b/src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
@@ -1,0 +1,19 @@
+<div class="invalid-segment ms-4 me-4 mt-1 mb-2">
+    <div class="header">
+        <div class="header-toolbar">
+          <span class="reference-tag">
+            {{segment.referenceTag}}
+          </span>
+        </div>
+        <div class="invalid p-2 pt-1 pb-1">
+            <sofie-icon-button [icon]="Icon.TRIANGLE_EXCLAMATION" [iconSize]="IconSize.M"></sofie-icon-button>
+            <span class="ms-1">INVALID</span>
+        </div>
+        <div class="title m-2">
+            {{ segment.name }}
+        </div>
+        <div class="text m-2">
+            {{ segment.invalidity?.reason }}
+        </div>
+    </div>
+</div>
--- a/src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
+++ b/src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
@@ -2,18 +2,18 @@
     <div class="header">
         <div class="header-toolbar">
           <span class="reference-tag">
-            {{segment.referenceTag}}
+            {{ segment.referenceTag }}
           </span>
         </div>
         <div class="invalid p-2 pt-1 pb-1">
-            <sofie-icon-button [icon]="Icon.TRIANGLE_EXCLAMATION" [iconSize]="IconSize.M"></sofie-icon-button>
+            <sofie-icon-button [icon]="Icon.TRIANGLE_EXCLAMATION" [iconSize]="IconSize.M" [disabled]="true"></sofie-icon-button>
             <span class="ms-1">INVALID</span>
+            <div class="text">
+                {{ segment.invalidity?.reason }}
+            </div>
         </div>
         <div class="title m-2">
             {{ segment.name }}
-        </div>
-        <div class="text m-2">
-            {{ segment.invalidity?.reason }}
         </div>
     </div>
 </div>
--- a/src/app/rundown-view/components/rundown/rundown.component.html
+++ b/src/app/rundown-view/components/rundown/rundown.component.html
@@ -1,20 +1,26 @@
 <div class="segments">
     <ng-container #segmentList *ngFor="let segment of rundown?.segments; trackBy: trackSegment">
-        <sofie-segment
-                class="segment"
-                [attr.data-segment-id]="segment.id"
-                *ngIf="!segment.isHidden"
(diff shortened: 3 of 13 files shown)


RasmusAlbrektsen (approved):
Nice! 
And thanks for removing BEMIT, that is very nice :-D

src/app/core/models/invalidity.ts
```diff
@@ -0,0 +1,3 @@
+export interface Invalidity {
+  reason: string
```
KvelaGorrrrnio: The reason should not change, thus it can be marked as readonly.

src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="invalid-segment ms-4 me-4 mt-1 mb-2">
+    <div class="header">
+        <div class="header-toolbar">
+          <span class="reference-tag">
+            {{segment.referenceTag}}
+          </span>
+        </div>
+        <div class="invalid p-2 pt-1 pb-1">
+            <sofie-icon-button [icon]="Icon.TRIANGLE_EXCLAMATION" [iconSize]="IconSize.M"></sofie-icon-button>
```
KvelaGorrrrnio: A minor thing. Because this is an icon button, the cursor indicates that one can interact with the element – which is not the case.

src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
```diff
@@ -0,0 +1,19 @@
...
+            <span class="ms-1">INVALID</span>
+        </div>
+        <div class="title m-2">
+            {{ segment.name }}
+        </div>
+        <div class="text m-2">
+            {{ segment.invalidity?.reason }}
+        </div>
```
KvelaGorrrrnio: Since the invalidity reason is coupled to the fact that the segment is invalid, I suggest that these are in close proximity and  are visually fused together to create a more cohesive user experience. I have 2 suggestions. Either the reason can be moved up right under the `∆ INVALID` text like
image
or it can be in a tooltip like
image
I think I lean towards the first option.

src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="invalid-segment ms-4 me-4 mt-1 mb-2">
+    <div class="header">
+        <div class="header-toolbar">
+          <span class="reference-tag">
+            {{segment.referenceTag}}
```
KvelaGorrrrnio: Would be nice with spacing in the brackets :)

src/app/rundown-view/components/segment/segment.component.scss
```diff
@@ -1,48 +1,47 @@
...
     background-color: var(--tv2-remote-color);
   }
 
-  &--set-as-next &__header {
+  &.is-next .header {
     background-color: var(--next-color);
   }
 
-  &--on-air &__header {
+  &.on-air .header {
```
KvelaGorrrrnio, line 24: To match the `is-next` syntax, shouldn't this be `is-on-air`?

src/app/rundown-view/components/rundown/rundown.component.html
```diff
@@ -1,20 +1,26 @@
...
+        <sofie-invalid-segment *ngIf="segment.invalidity; else miniShelfTemplate" [segment]="segment">
+        </sofie-invalid-segment>
+        <ng-template #miniShelfTemplate>
+            <sofie-mini-shelf
+                    *ngIf="segment.metadata?.miniShelfVideoClipFile; else segmentTemplate"
+                    [segment]="segment"
+                    [videoClipAction]="miniShelfSegmentActionMappings[segment.id]"
+            ></sofie-mini-shelf>
+        </ng-template>
+        <ng-template #segmentTemplate>
```
KvelaGorrrrnio, line 12: It would be nice with a newline between the different parts and templates, to have a visual separation of grouped content.

src/app/rundown-view/components/segment/segment.component.scss
LindvedKrvang: I would say no. CSS classes describes what things are. A Segment can be OnAir.
If we should change something I think I would change `is-next` to `next`.

src/app/rundown-view/components/rundown/rundown.component.html
LindvedKrvang: You have that in the indentation?

src/app/core/models/invalidity.ts
LindvedKrvang: Updated

src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
LindvedKrvang: Updated the `IconButton` to if you give it `disabled=true` the cursor won't change when hovering the icon.

src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
LindvedKrvang: Updated it to:
image

src/app/rundown-view/components/invalid-segment/invalid-segment.component.html
LindvedKrvang: Updated. Also updated the place I copied this from

src/app/rundown-view/components/segment/segment.component.scss
KvelaGorrrrnio: Sounds like a plan :+1:

src/app/rundown-view/components/rundown/rundown.component.html
KvelaGorrrrnio: `sofie-invalid-segment` and the two `ng-template`s are on the same level, so they can visually melt a bit together. Adding newlines between mutually exclusive "templates" would make it easier (and faster) to distinguish them from one another.

KvelaGorrrrnio (approved):
Nice work! There are still 2 open threads from last batch.

PR #203, opened 2024-05-15
SOF-1981 InsertIntoSplitScreenActions shows the color of what they are inserting

LindvedKrvang:
Note: This is dependent on https://github.com/tv2/sofie-server/pull/154
All dependent PRs have been merged.

Also in this PR: Deleted unused file `tv2-action-group-service` and interfaces just used by that file


RasmusAlbrektsen (approved):

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -192,6 +200,18 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
...
 
+  private getBackgroundColorForSplitScreen(action: Tv2Action): string {
+    if (!this.isActionInsertIntoSplitScreenAction(action)) {
+      return SPLIT_SCREEN_COLOR
+    }
+    const topColor: string = this.getActionContentTypeBackgroundColor(action.metadata.insertedContentType)
+    return `linear-gradient(to bottom, ${topColor} 50%, var(--tv2-split-screen-lower) 50%)`
+  }
+
+  private isActionInsertIntoSplitScreenAction(action: Tv2Action): action is Tv2InsertIntoSplitScreenAction {
```
KvelaGorrrrnio: The first `Action` seems redundant. Suggestion: `isInsertIntoSplitScreenAction`.

src/app/shared/models/tv2-action.ts
```diff
@@ -35,23 +35,10 @@ export interface Tv2VideoClipAction extends Tv2PartAction {
...
-    remoteNumber: string
-  }
-}
-
-export interface Tv2TransitionAction extends Tv2PartAction {
-  metadata: {
-    contentType: Tv2ActionContentType.TRANSITION
+    contentType: Tv2ActionContentType.SPLIT_SCREEN
+    insertedContentType: Tv2ActionContentType
   }
```
KvelaGorrrrnio, line 42: Why are the camera, remote and transition action types removed?

LindvedKrvang: Because they where not used.
They where only used by `tv2-action-group.service` which also was not used (that file has also been deleted).
We should not have unused/dead code in our repository.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts

KvelaGorrrrnio (approved):

PR #204, opened 2024-05-15
SOF-1928 Replaced planned Pieces disappear

LindvedKrvang:
Replaced planned Pieces are kept and being reinserted into the Part on reset

### DIFF
--- a/src/app/core/models/part.ts
+++ b/src/app/core/models/part.ts
@@ -16,4 +16,5 @@ export interface Part {
   readonly isPlanned: boolean
   readonly isUntimed: boolean
   readonly metadata?: unknown
+  readonly replacedPieces: Piece[]
 }
--- a/src/app/core/models/part.ts
+++ b/src/app/core/models/part.ts
@@ -16,5 +16,5 @@ export interface Part {
   readonly isPlanned: boolean
   readonly isUntimed: boolean
   readonly metadata?: unknown
-  readonly replacedPieces: Piece[]
+  readonly replacedPieces: readonly Piece[]
 }
--- a/src/app/core/services/models/part-entity.service.ts
+++ b/src/app/core/services/models/part-entity.service.ts
@@ -41,11 +41,13 @@ export class PartEntityService {
   }
 
   public reset(part: Part): Part {
+    const pieces: Piece[] = [...part.pieces.filter(piece => piece.isPlanned), ...(part.replacedPieces ?? [])]
+    part.replacedPieces.splice(0, part.replacedPieces.length)
     return {
       ...part,
       playedDuration: 0,
       executedAt: 0,
-      pieces: part.pieces.filter(piece => piece.isPlanned),
+      pieces,
     }
   }
 
@@ -104,6 +106,10 @@ export class PartEntityService {
   }
 
   public replacePiece(part: Part, replacedPieceId: string, newPiece: Piece): Part {
+    const pieceToBeReplaced: Piece | undefined = part.pieces.find(piece => piece.id === replacedPieceId)
+    if (pieceToBeReplaced && pieceToBeReplaced.isPlanned) {
+      part.replacedPieces.push(pieceToBeReplaced)
+    }
     return {
       ...part,
       pieces: [...part.pieces.filter(piece => piece.id !== replacedPieceId), newPiece],
--- a/src/app/core/services/models/part-entity.service.ts
+++ b/src/app/core/services/models/part-entity.service.ts
@@ -41,13 +41,13 @@ export class PartEntityService {
   }
 
   public reset(part: Part): Part {
-    const pieces: Piece[] = [...part.pieces.filter(piece => piece.isPlanned), ...(part.replacedPieces ?? [])]
-    part.replacedPieces.splice(0, part.replacedPieces.length)
+    const pieces: Piece[] = [...part.pieces, ...part.replacedPieces].filter(piece => piece.isPlanned)
     return {
       ...part,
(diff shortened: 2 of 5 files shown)


RasmusAlbrektsen (approved):

src/app/core/models/part.ts
```diff
@@ -16,4 +16,5 @@ export interface Part {
   readonly isPlanned: boolean
   readonly isUntimed: boolean
   readonly metadata?: unknown
+  readonly replacedPieces: Piece[]
```
KvelaGorrrrnio: Since this should only hold planned pieces, it would make sense that is called `replacedPlannedPieces`.
It removes the uncertainty of what it holds.

KvelaGorrrrnio: To disallow referential updates of immutable state, I suggest that the Array type is made readonly as well, so either `Readonly<Piece[]> or `ReadonlyArray<Piece>`.

src/app/core/services/models/part-entity.service.ts
```diff
@@ -41,11 +41,13 @@ export class PartEntityService {
   }
 
   public reset(part: Part): Part {
+    const pieces: Piece[] = [...part.pieces.filter(piece => piece.isPlanned), ...(part.replacedPieces ?? [])]
```
KvelaGorrrrnio: `Part.replacedPieces` is not optional, so the nullish coalescing should be removed.
The number of replaced pieces should not be tremendous, so one could improve the readability by moving the filter out of the array instantiation, e.g.:
```typescript
 [...part.pieces, ...part.replacedPieces].filter(piece => piece.isPlanned)
```

src/app/core/services/models/part-entity.service.ts
```diff
@@ -41,11 +41,13 @@ export class PartEntityService {
...
+    part.replacedPieces.splice(0, part.replacedPieces.length)
     return {
       ...part,
       playedDuration: 0,
       executedAt: 0,
-      pieces: part.pieces.filter(piece => piece.isPlanned),
+      pieces,
     }
```
KvelaGorrrrnio: We should not impose side-effects on input parameters. Instead, an empty array can be instantiated on the new part instance like so:
```typescript
return {
  ...part,
  playedDuration: 0,
  executedAt: 0,
  pieces,
  replacedPieces: [],
}
```

src/app/core/services/models/part-entity.service.ts
```diff
@@ -104,6 +106,10 @@ export class PartEntityService {
   }
 
   public replacePiece(part: Part, replacedPieceId: string, newPiece: Piece): Part {
+    const pieceToBeReplaced: Piece | undefined = part.pieces.find(piece => piece.id === replacedPieceId)
+    if (pieceToBeReplaced && pieceToBeReplaced.isPlanned) {
+      part.replacedPieces.push(pieceToBeReplaced)
+    }
     return {
       ...part,
       pieces: [...part.pieces.filter(piece => piece.id !== replacedPieceId), newPiece],
```
KvelaGorrrrnio: We should not mutate input parameters. Instead, a new replaced pieces array should be created on the new part object.
```typescript
    return {
      ...part,
      pieces: [...part.pieces.filter(piece => piece.id !== replacedPieceId), newPiece],
      replacedPieces: pieceToReplace?.isPlanned ? [...part.replacedPieces, pieceToReplace] : part.replacedPieces,
    }
```

src/app/core/models/part.ts
LindvedKrvang: I'm hesitant to call it `replacedPlannedPieces`. Mainly because right now it is `part-entity.service` that only puts Planned Pieces in it. But `part-entity.service` is a different file. So if we call the attribute `replacedPlannedPieces` then we are naming an attribute in one file after how _another file_ uses it. 

There is nothing in this interface that can force it to be planned Pieces, so it's entirely possible (and legal) for some other file to put an unplanned Piece in this array.

Now, if this was a class that had full control of what goes into the array then I would be all for it to call it `replacedPlannedPieces`, but since we have taken the functional approach here we don't have that same guarentee.

LindvedKrvang: Update to `readonly Piece[]` since `ReadonlyArray<T>` is not allowed by our linting rules
image

src/app/core/services/models/part-entity.service.ts
LindvedKrvang: Updated.

Sidenote: I must admit, the more I work with functional programming, the less fan I am becoming of it. All the filtering and reinstantiation is a lot more cumbersome to work with than need be.

src/app/core/services/models/part-entity.service.ts
LindvedKrvang: Updated

src/app/core/services/models/part-entity.service.ts
LindvedKrvang: Updated

src/app/core/models/part.ts
KvelaGorrrrnio: Changing the type for this to only contain planned pieces is simple (either by extension or type intersection), in which case it is guaranteed. If it is allowed to put an unplanned piece in this, then the code prior to [this change](https://github.com/tv2/ng-sofie/pull/204#discussion_r1604424074) would have had a hidden dependency.

I am fine with keeping it as is.

KvelaGorrrrnio (approved):

KvelaGorrrrnio (approved):

PR #205, opened 2024-05-16
Sof 1982/set color in keyboard mapping configuration

LindvedKrvang:
Note: Is dependent on https://github.com/tv2/ng-sofie/pull/203
All dependent PRs have been merged.

Introduce ColorPickerComponent.

Now possible to select an "override color" when creating an KeyboardMapping. For now, the selected color does nothing (it will be implemented by SOF-1983).

### DIFF
--- a/src/app/shared/components/color-picker/color-picker.component.html
+++ b/src/app/shared/components/color-picker/color-picker.component.html
@@ -4,9 +4,10 @@
     <sofie-icon-button *ngIf="helpText" [tooltip]="helpText" [icon]="Icon.CIRCLE_QUESTION" [iconSize]="IconSize.M"></sofie-icon-button>
   </span>
     <div class="color-picker" ngx-colors-trigger [(ngModel)]="value" [palette]="palette" [colorPickerControls]="'no-alpha'" [format]="'hex'" (input)="emitChanges()" (click)="markAsTouched()">
-        <div class="color-box" [style]="{ 'background': getColor }">
+        <span *ngIf="!!value" class="color-box" [style]="{ 'background': getColor }"></span>
+        <span class="text">
             {{ !!value ? value : placeholder ?? label }}
-        </div>
+        </span>
         <sofie-icon-button *ngIf="value" [icon]="Icon.X_MARK" [iconSize]="IconSize.L" (click)="clearValue($event)"></sofie-icon-button>
     </div>
     <sofie-error-message *ngIf="errorMessage">{{ errorMessage }}</sofie-error-message>
--- a/src/app/shared/components/color-picker/color-picker.component.html
+++ b/src/app/shared/components/color-picker/color-picker.component.html
@@ -4,11 +4,11 @@
     <sofie-icon-button *ngIf="helpText" [tooltip]="helpText" [icon]="Icon.CIRCLE_QUESTION" [iconSize]="IconSize.M"></sofie-icon-button>
   </span>
     <div class="color-picker" ngx-colors-trigger [(ngModel)]="value" [palette]="palette" [colorPickerControls]="'no-alpha'" [format]="'hex'" (input)="emitChanges()" (click)="markAsTouched()">
-        <span *ngIf="!!value" class="color-box" [style]="{ 'background': getColor }"></span>
+        <span *ngIf="!!value" class="color-box" [style]="{ 'background': getColor() }"></span>
         <span class="text">
             {{ !!value ? value : placeholder ?? label }}
         </span>
-        <sofie-icon-button *ngIf="value" [icon]="Icon.X_MARK" [iconSize]="IconSize.L" (click)="clearValue($event)"></sofie-icon-button>
+        <sofie-icon-button *ngIf="value" [icon]="Icon.X_MARK" [iconSize]="IconSize.L" (click)="clearValue(); $event.stopPropagation()"></sofie-icon-button>
     </div>
     <sofie-error-message *ngIf="errorMessage">{{ errorMessage }}</sofie-error-message>
 </div>
--- a/src/app/shared/components/color-picker/color-picker.component.html
+++ b/src/app/shared/components/color-picker/color-picker.component.html
@@ -1,14 +1,14 @@
-<div class="container pt-2 pb-2">
-  <span *ngIf="label || helpText" class="mb-1">
+<div class="pt-2 pb-2">
+  <div *ngIf="label || helpText" class="label-container mb-1">
     <label *ngIf="label">{{ label }}<sofie-required-mark *ngIf="isRequired"></sofie-required-mark></label>
     <sofie-icon-button *ngIf="helpText" [tooltip]="helpText" [icon]="Icon.CIRCLE_QUESTION" [iconSize]="IconSize.M"></sofie-icon-button>
-  </span>
+  </div>
     <div class="color-picker" ngx-colors-trigger [(ngModel)]="value" [palette]="palette" [colorPickerControls]="'no-alpha'" [format]="'hex'" (input)="emitChanges()" (click)="markAsTouched()">
-        <span *ngIf="!!value" class="color-box" [style]="{ 'background': getColor() }"></span>
-        <span class="text">
-            {{ !!value ? value : placeholder ?? label }}
+        <span *ngIf="value" class="color-box" [style.background]="getColor()"></span>
+        <span class="input-text">
+            {{ value || placeholder ?? label }}
         </span>
         <sofie-icon-button *ngIf="value" [icon]="Icon.X_MARK" [iconSize]="IconSize.L" (click)="clearValue(); $event.stopPropagation()"></sofie-icon-button>
     </div>
-    <sofie-error-message *ngIf="errorMessage">{{ errorMessage }}</sofie-error-message>
+<sofie-error-message *ngIf="errorMessage">{{ errorMessage }}</sofie-error-message>
 </div>
--- a/src/app/shared/components/color-picker/color-picker.component.html
+++ b/src/app/shared/components/color-picker/color-picker.component.html
@@ -1,0 +1,13 @@
+<div class="container pt-2 pb-2">
+  <span *ngIf="label || helpText" class="mb-1">
+    <label *ngIf="label">{{ label }}<sofie-required-mark *ngIf="isRequired"></sofie-required-mark></label>
+    <sofie-icon-button *ngIf="helpText" [tooltip]="helpText" [icon]="Icon.CIRCLE_QUESTION" [iconSize]="IconSize.M"></sofie-icon-button>
+  </span>
+    <div class="color-picker" ngx-colors-trigger [(ngModel)]="value" [palette]="palette" [colorPickerControls]="'no-alpha'" [format]="'hex'" (input)="emitChanges()" (click)="markAsTouched()">
+        <div class="color-box" [style]="{ 'background': getColor }">
+            {{ !!value ? value : placeholder ?? label }}
+        </div>
+        <sofie-icon-button *ngIf="value" [icon]="Icon.X_MARK" [iconSize]="IconSize.L" (click)="clearValue($event)"></sofie-icon-button>
+    </div>
+    <sofie-error-message *ngIf="errorMessage">{{ errorMessage }}</sofie-error-message>
+</div>
--- a/src/app/shared/components/color-picker/color-picker.component.scss
+++ b/src/app/shared/components/color-picker/color-picker.component.scss
@@ -17,6 +17,8 @@
 
 .color-picker {
   position: relative;
+  display: flex;
+  flex-direction: row;
   width: 100%;
   padding: 0.3rem 0.3rem;
   font-weight: lighter;
@@ -35,6 +37,13 @@
 
 .color-box {
   padding: 0.2rem 0.5rem;
+  width: 50px;
+  height: 22px;
+  border-radius: 4px;
+}
+
+.text {
+  padding: 0.2rem;
   font-weight: lighter;
   font-size: 0.85rem;
   font-family: Roboto, Arial, sans-serif;
--- a/src/app/shared/components/color-picker/color-picker.component.scss
+++ b/src/app/shared/components/color-picker/color-picker.component.scss
@@ -1,0 +1,55 @@
+:host {
+  &.ng-invalid {
+    &.ng-touched {
+      input {
+        border-color: var(--danger-color);
+      }
+    }
+  }
+
+  &.ng-valid,
+  &.ng-untouched {
+    sofie-error-message {
+      display: none;
+    }
+  }
+}
+
+.color-picker {
+  position: relative;
+  width: 100%;
+  padding: 0.3rem 0.3rem;
+  font-weight: lighter;
+  font-size: 0.85rem;
+  font-family: Roboto, Arial, sans-serif;
+  border-radius: 4px;
+  cursor: pointer;
+  border: 1px solid var(--gray-color);
+
+  sofie-icon-button {
+    position: absolute;
+    right: 10px;
+    top: 8px;
+  }
+}
+
+.color-box {
+  padding: 0.2rem 0.5rem;
+  font-weight: lighter;
+  font-size: 0.85rem;
+  font-family: Roboto, Arial, sans-serif;
+}
+
+.container {
+  display: flex;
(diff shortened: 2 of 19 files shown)


RasmusAlbrektsen (approved):
This looks very good!

src/app/shared/models/keyboard-trigger-data.ts
```diff
@@ -5,6 +5,7 @@ export interface KeyboardTriggerData {
   keys: Keys
   actionArguments?: string | number
   label: string
+  overrideColor?: string
```
KvelaGorrrrnio: Is the optionality needed? An empty string could represent no selected.

src/locale/messages.da.xlf
```diff
@@ -488,14 +486,22 @@
...
+        <source>action-triggers.override-color.label</source>
+        <target state="new">Overskriv standard farve</target>
+      </trans-unit>
+      <trans-unit id="2313705565869821168" datatype="html">
+        <source>color-picker.choose-color.placeholder</source>
+        <target state="new">Vælg en farve</target>
+      </trans-unit>
+      <trans-unit id="4439615397476866844" datatype="html">
+        <source>action-triggers.override-color.help-text</source>
+        <target state="new">Vælg en ny farve hvis du gerne overskrive standard farven for genvejen.</target>
```
KvelaGorrrrnio:
```suggestion
        <target state="new">Vælg en ny farve hvis du gerne vil overskrive standard farven for genvejen.</target>
```

src/app/shared/components/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  protected value?: string
+
+  private isTouched: boolean = false
+
+  private onChangeCallback: (value?: string) => void
+  private onTouchedCallback: () => void
+
+  constructor() {}
+
+  protected get getColor(): string {
```
KvelaGorrrrnio: Either this should not be a getter method or it should be called the attribute name `get color()` or `getColor()`.

src/app/shared/components/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  protected markAsTouched(): void {
+    if (!this.onTouchedCallback || this.isTouched) {
+      return
+    }
+    this.onTouchedCallback()
+    this.isTouched = true
+  }
+
+  protected clearValue(event: Event): void {
+    event.stopPropagation()
```
KvelaGorrrrnio: Stopping the event propagation is not a responsibility of clearing the value. Instead it can be put directly in the click handler in the template as `(click)="clearValue(); $event.stopPropagation()"`.

src/app/shared/components/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  ]
+
+  @Input() public label: string
+  @Input() public placeholder?: string
+
+  @Input() public helpText: string
+  @Input() public isRequired: boolean
+  @Input() public errorMessage?: string
+
+  @Output() public valueChange: EventEmitter<string> = new EventEmitter()
```
KvelaGorrrrnio, line 49: Just a suggestion: Instead of value it could be color, so `colorChange`.
Feel free to disregard if it is to keep consistency with other input fields.

src/app/shared/components/color-picker/color-picker.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="container pt-2 pb-2">
+  <span *ngIf="label || helpText" class="mb-1">
+    <label *ngIf="label">{{ label }}<sofie-required-mark *ngIf="isRequired"></sofie-required-mark></label>
+    <sofie-icon-button *ngIf="helpText" [tooltip]="helpText" [icon]="Icon.CIRCLE_QUESTION" [iconSize]="IconSize.M"></sofie-icon-button>
+  </span>
+    <div class="color-picker" ngx-colors-trigger [(ngModel)]="value" [palette]="palette" [colorPickerControls]="'no-alpha'" [format]="'hex'" (input)="emitChanges()" (click)="markAsTouched()">
+        <span *ngIf="!!value" class="color-box" [style]="{ 'background': getColor }"></span>
```
KvelaGorrrrnio: instead of setting the background as an object, you can use the attribute chaining syntax:
`[style.background]="getColor"`

KvelaGorrrrnio: The `!!` is redundant.

src/app/shared/components/color-picker/color-picker.component.html
```diff
@@ -0,0 +1,14 @@
...
+        <span class="text">
+            {{ !!value ? value : placeholder ?? label }}
```
KvelaGorrrrnio: The `!!` is redundant. To simplify the expression, you can use the or operator as it has been done on line 2:
```typescript
value || placeholder ?? label
```

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -0,0 +1,64 @@
...
+  padding: 0.2rem;
+  font-weight: lighter;
+  font-size: 0.85rem;
+  font-family: Roboto, Arial, sans-serif;
+}
+
+.container {
+  display: flex;
+  flex-direction: column;
+}
```
KvelaGorrrrnio: If the span on line 2 in the template is changed to a div, then this can be removed.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -0,0 +1,64 @@
...
+
+span {
+  display: flex;
+  justify-content: space-between;
+}
```
KvelaGorrrrnio: I assume that this is for spacing the help icon from the label. If that assumption is correct, then this is a too broad selector. Instead create a class for that element. E.g. `.label-container`.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -0,0 +1,64 @@
...
+    top: 8px;
+  }
+}
+
+.color-box {
+  padding: 0.2rem 0.5rem;
+  width: 50px;
+  height: 22px;
+  border-radius: 4px;
+}
```
KvelaGorrrrnio: Setting both padding along with height and width seems unnecessary.
The mix of px and rem also makes it scale not ideally if the font-size changes.
I would suggest to set the height with rem and set the aspect ratio to something like 16/9, e.g.:
```scss
.color-box {
  height: 1.5rem;
  aspect-ratio: 16 / 9;
  border-radius: 4px;
}
```
The border radius is kept as pixels since, other border radii uses pixels as well.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -0,0 +1,64 @@
...
+
+.text {
```
KvelaGorrrrnio: `text` is a bit ambiguous, since we have multiple kinds of text in the component. A suggestion could be `input-text`.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -0,0 +1,64 @@
...
+  padding: 0.2rem;
+  font-weight: lighter;
+  font-size: 0.85rem;
+  font-family: Roboto, Arial, sans-serif;
```
KvelaGorrrrnio: This is the same font family as specified for the entire document. It should not be necessary. If the current setup doesn't apply the font family correctly globally, then it should be fixed there.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -0,0 +1,64 @@
...
+      display: none;
+    }
+  }
+}
+
+.color-picker {
+  position: relative;
+  display: flex;
+  flex-direction: row;
+  width: 100%;
```
KvelaGorrrrnio: Is this necessary?

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -0,0 +1,64 @@
...
+  padding: 0.3rem 0.3rem;
+  font-weight: lighter;
+  font-size: 0.85rem;
+  font-family: Roboto, Arial, sans-serif;
+  border-radius: 4px;
+  cursor: pointer;
+  border: 1px solid var(--gray-color);
```
KvelaGorrrrnio: It would be nice if related properties were grouped (border-radius and border).

src/app/shared/components/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  protected value?: string
+
+  private isTouched: boolean = false
+
+  private onChangeCallback: (value?: string) => void
+  private onTouchedCallback: () => void
+
+  constructor() {}
```
KvelaGorrrrnio, line 58: If it should be there, then an access modifier would be nice :)

src/app/shared/components/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,93 @@
...
+    '#265753',
+    '#370020',
+    '#1769ff',
+    '#370020',
+    '#016492',
+    '#003c58',
+    '#00ac17',
+    '#ef6c00',
+  ]
+
```
KvelaGorrrrnio, line 41: Out of curiosity, where do these values come from? :)

KvelaGorrrrnio (commented):
If you open the color-picker in a sidepanel and press Escape, then the sidepanel is closed, but the color-picker is still there:
Skærmbillede 2024-05-17 kl  10 41 23

src/locale/messages.en-US.xlf
```diff
@@ -486,14 +486,22 @@
...
-        <source>keyboard-mapping.settings-page.invalid-field</source>
-        <target state="new">Invalid data, needs to be edited.</target>
-      </trans-unit>
       <trans-unit id="5805882241277564116" datatype="html">
         <source>keyboard-mapping-settings-page-import-keyboard-mappings.actions-missing</source>
         <target state="new">One or more keyboard mappings contained invalid actions and were not imported!</target>
       </trans-unit>
+      <trans-unit id="8483503986037571867" datatype="html">
+        <source>action-triggers.override-color.label</source>
+        <target state="new">Override default color</target>
```
KvelaGorrrrnio: Since what _default color_ refers to is not specified anywhere, I suggest that we change it to `Override action color` or name it something like `Key color`.

src/app/shared/components/color-picker/color-picker.component.ts
LindvedKrvang: `value` was used mainly to keep consistent with the `ControlValueAccessor` interface required for it to be used in reactive forms.

src/app/shared/models/keyboard-trigger-data.ts
LindvedKrvang: Made it non-optional

src/locale/messages.da.xlf
LindvedKrvang: Updated

src/app/shared/components/color-picker/color-picker.component.ts
LindvedKrvang: Good catch.
It is no longer at getter method.

src/app/shared/components/color-picker/color-picker.component.ts
LindvedKrvang: Updated

src/app/shared/components/color-picker/color-picker.component.ts
LindvedKrvang: We only put access level modifier on our constructors if they aren't public.
We don't have a single instance where we have put `public` in front of our constuctors.

src/app/shared/components/color-picker/color-picker.component.ts
LindvedKrvang: A selection from `_color-vars.scss`.
I am very much interested in different suggestions for how to decide the palette.

src/app/shared/components/color-picker/color-picker.component.html
LindvedKrvang: Updated

LindvedKrvang: Updated

src/app/shared/components/color-picker/color-picker.component.html
LindvedKrvang: Updated

src/app/shared/components/color-picker/color-picker.component.scss
LindvedKrvang: Updated

src/app/shared/components/color-picker/color-picker.component.scss
LindvedKrvang: Updated

src/app/shared/components/color-picker/color-picker.component.scss
LindvedKrvang: Updated

src/app/shared/components/color-picker/color-picker.component.scss
LindvedKrvang: Updated

src/app/shared/components/color-picker/color-picker.component.scss
LindvedKrvang: Deleted

src/app/shared/components/color-picker/color-picker.component.scss
LindvedKrvang: Apparrently not. It has been removed

src/app/shared/components/color-picker/color-picker.component.scss
LindvedKrvang: Updated

src/locale/messages.en-US.xlf
LindvedKrvang: Update to override Key color

src/app/shared/components/color-picker/color-picker.component.ts
KvelaGorrrrnio: At some point we talked about making them explicit to ensure that a deliberate choice was made.
Is that something we want or don't care about?

src/app/shared/components/color-picker/color-picker.component.ts
KvelaGorrrrnio: One could use css variables, but it would surprise me if that was supported in the color picker. So no good solutions at the moment – was just curious :D

src/locale/messages.en-US.xlf
KvelaGorrrrnio: Is it intentional to keep `default` in this translation?

src/locale/messages.da.xlf
```diff
@@ -488,14 +486,22 @@
...
+        <source>action-triggers.override-color.label</source>
+        <target state="new">Overskriv standard farve</target>
+      </trans-unit>
+      <trans-unit id="2313705565869821168" datatype="html">
+        <source>color-picker.choose-color.placeholder</source>
+        <target state="new">Vælg en farve</target>
+      </trans-unit>
+      <trans-unit id="4439615397476866844" datatype="html">
+        <source>action-triggers.override-color.help-text</source>
+        <target state="new">Vælg en ny farve hvis du gerne vil overskrive standard farven for genvejen.</target>
```
KvelaGorrrrnio: The danish translation does not match the english in terms of `default color` vs `key color`.

KvelaGorrrrnio (approved):

src/app/shared/components/color-picker/color-picker.component.ts
LindvedKrvang: I don't recall talking about doing it for constructors.
We do it for all methods, which we want to continue doing.

src/locale/messages.en-US.xlf
LindvedKrvang: No, it has been updated to be `key`.

src/locale/messages.da.xlf
LindvedKrvang: Updated to be `Overskriv genvejs farven` and `... hvis du gerne vil overskrive farven for genvejen`.

PR #209, opened 2024-05-31
On air segments have sticks to viewport boundaries.

KvelaGorrrrnio:
Adds sticky segment component that is used by the rundown component to display an on air segment. The sticky segment is set with boundaries at the top and bottom of the viewport, sticking 1px outside of the viewport in order to detect when the segment sticks to the viewport boundary.
image

### DIFF
--- a/src/app/rundown-view/components/sticky-segment/sticky-segment.component.ts
+++ b/src/app/rundown-view/components/sticky-segment/sticky-segment.component.ts
@@ -27,9 +27,7 @@ export class StickySegmentComponent implements AfterViewInit, OnDestroy {
 
   private readonly intersectionObserver: IntersectionObserver
 
-  constructor(
-    private readonly hostElement: ElementRef,
-  ) {
+  constructor(private readonly hostElement: ElementRef) {
     this.intersectionObserver = this.createViewportIntersectionObserver()
   }
 
--- a/src/app/rundown-view/components/sticky-segment/sticky-segment.component.ts
+++ b/src/app/rundown-view/components/sticky-segment/sticky-segment.component.ts
@@ -1,0 +1,52 @@
+import { AfterViewInit, Component, ElementRef, Input, OnDestroy } from '@angular/core'
+import { Segment } from '../../../core/models/segment'
+
+@Component({
+  selector: 'sofie-sticky-segment',
+  templateUrl: './sticky-segment.component.html',
+  styleUrls: ['./sticky-segment.component.scss'],
+})
+export class StickySegmentComponent implements AfterViewInit, OnDestroy {
+  @Input()
+  public segment: Segment
+
+  @Input()
+  public isRundownActiveOrRehearsal: boolean
+
+  @Input()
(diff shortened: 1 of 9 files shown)


src/app/rundown-view/components/sticky-segment/sticky-segment.component.ts
```diff
@@ -0,0 +1,50 @@
+import { AfterViewInit, Component, ElementRef, Input, OnDestroy } from '@angular/core'
+import { Segment } from '../../../core/models/segment'
+
+@Component({
+  selector: 'sofie-sticky-segment',
+  templateUrl: './sticky-segment.component.html',
+  styleUrls: ['./sticky-segment.component.scss'],
+})
+export class StickySegmentComponent implements AfterViewInit, OnDestroy {
```
LindvedKrvang, line 9: Does this really need to be a new component?
Is basically just the SegmentComponent with a CSS class added to it.

LindvedKrvang:
It bugs me a little that the background of the Sticky Segment doesn't go "all the way to the end". Not sure if we should do anything about it
image

src/app/rundown-view/components/sticky-segment/sticky-segment.component.ts
KvelaGorrrrnio: It keeps the original segment component cleaner and makes it easier to extend the functionality of the sticky segment without altering the original segment component – e.g. if we want to add a button for scrolling it into view or something else.

KvelaGorrrrnio:
> It bugs me a little that the background of the Sticky Segment doesn't go "all the way to the end". Not sure if we should do anything about it image

Initially I had the same thought, but making it extend to the right edge of the viewport would mean that the user can't use the scrollbar due to the sticky segment overlaying the scrollbar.

src/app/rundown-view/components/sticky-segment/sticky-segment.component.ts
LindvedKrvang: My biggest issue is that almost all the input variables (which there are quite a few of) is just parse-through...

LindvedKrvang:
> > It bugs me a little that the background of the Sticky Segment doesn't go "all the way to the end". Not sure if we should do anything about it image
> 
> Initially I had the same thought, but making it extend to the right edge of the viewport would mean that the user can't use the scrollbar due to the sticky segment overlaying the scrollbar.

Lets see if we get any comments about it

LindvedKrvang (approved):

src/app/rundown-view/components/sticky-segment/sticky-segment.component.ts
RasmusAlbrektsen: I see the point Lindved is making about everything being passed through, but i think it makes sense to have some general pattern we implement to avoid growing components further, especially if we can argue that this makes for a new "version" of a component.

RasmusAlbrektsen (approved):
Looks good and works great!

PR #213, opened 2024-06-11
Update to Angular 16

MadsThy:
This updates the Sofie Web client from Angular 14 to Angular 16. It is impossible to proceed further in updates, as Angular 16 is the last version to support Node 16. If we want to update further, we need our infrastructure to support Node 18+.

Besides the update of Angular, a small change was made to Prettier, so it works with VS Code formatting.

The main brunt of the change is that Material design has been updated, which means we have to use legacy components now. It doesn't change anything, but we should consider moving to the "real" material components in the future when there is time to update this.

**Additional**
To fix the linting errors that occur after the upgrade, @typescript-eslint/recommended was implemented in .eslintRC. This means many errors started showing up because we weren't following the recommended ESlint rules in the first place. They have now been added to the list of rules and turned into warnings instead of errors so that we can proceed. It should probably be investigated in the future though if we want to keep them as a warning or turn them off (or potentially fix the linting errors instead)


LindvedKrvang (approved):
Nice work :)

aardhuis (approved):
Nicely done :)

.prettierrc.json
```diff
@@ -7,5 +7,6 @@
   "arrowParens": "avoid",
   "trailingComma": "es5",
   "bracketSameLine": true,
-  "printWidth": 200
+  "printWidth": 200,
+  "endOfLine": "auto"
```
KvelaGorrrrnio: Any reason this is `auto` and not `lf`? :)
Having the same eol character for all files would be nice.

KvelaGorrrrnio (approved):

.prettierrc.json
MadsThy: My bad! My brain reverted back to last time I was working on a Windows based setup :) I agree that it should be LF! Changing now :)

PR #214, opened 2024-06-11
Improve GitHub check flow

MadsThy:
This PR (in my mind) improves the flow of how we use the checking steps in our validation.

It takes the steps we already have and moves them into a matrix setup. This is then used to create a step for each of the actions, making it more visible in our checking what is going on, and also utilizing the yarn caching as per the documentation from GitHub.


LindvedKrvang (commented):
I have no objections of putting our checks in a matrix.
Before I approve I just need to ask whether these changes are why there is the `Unchanged files with check annotations` section on this PR? If yes, then I am more hesitant since that would get real annoying, real quick
image

MadsThy:
@LindvedKrvang That is a consequence of me using the "Angular update" branch as a base when I made this change 😄 Those "check annotations" will persist in the Repo after the Angular update PR is merged, since it now "correctly" implements the recommended checks from "@typescript-eslint/recommended".

If you don't want that noise in this PR, I can remake the PR with the develop branch as a base 😄 then it will be clean

.github/workflows/check.yml
```diff
@@ -2,37 +2,27 @@ name: 'Check'
...
 on:
   pull_request:
-jobs:
 
-  test:
-    runs-on: 'ubuntu-latest'
-    name: 'Checking code and dependencies'
+jobs:
+  build-test-lint-validate:
+    runs-on: ubuntu-latest
```
KvelaGorrrrnio: Minor thing: It would be nice if the use of single-quotes for string values is consistent. Personally I am in favor of having them wrapped in single-quotes :)

KvelaGorrrrnio (approved):
Nice work :)

.github/workflows/check.yml
MadsThy: Done :)

MadsThy:
> I have no objections of putting our checks in a matrix. Before I approve I just need to ask whether these changes are why there is the `Unchanged files with check annotations` section on this PR? If yes, then I am more hesitant since that would get real annoying, real quick image

Should be fixed now :)

MadsThy:
Closing this PR due to noise

PR #216, opened 2024-06-13
SOF-2016: Show part invalidity

KvelaGorrrrnio:
Depends on: https://github.com/tv2/sofie-server/pull/168.

The PR adds visual indication of invalid parts with an overlayed attention-colored grid, along with tooltips describing the reason for the invalidity.

Skærmbillede 2024-06-13 kl  09 38 06

The PR also introduces the `AttentionBannerComponent`, which removes the redundant styling and setup we have for showing a banner with warning icon, a label and a description.
By doing this, the color of the _UNSYNCED_ banner on the segment header is changed to use the attention color, like  the invalid segment banner.

The container showing overlays for parts have been changed to use the full height, such that the visual impact for part overlays is greater.

Skærmbillede 2024-06-13 kl  09 41 30

### DIFF
--- a/src/app/core/models/part.ts
+++ b/src/app/core/models/part.ts
@@ -1,5 +1,6 @@
 import { Piece } from './piece'
-import { AutoNext } from '../models/auto-next'
+import { AutoNext } from './auto-next'
+import { Invalidity } from './invalidity'
 
 export interface Part {
   readonly id: string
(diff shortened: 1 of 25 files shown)


LindvedKrvang (changes requested):
The code looks good.
The reason for the request for changes is missing functionality. I am still able to right click an invalid Part and set it as next. That should not be possible. (It should also not be autoselected as next if the Part just before it is Taken)

LindvedKrvang (approved):
Allright, you are making the it impossible to take an invalid Part in a different PR, so you get an approve anyway ;)

RasmusAlbrektsen (approved):

PR #217, opened 2024-06-14
SOF-2016: Make invalid parts unreachable

KvelaGorrrrnio:
Updates rundown navigation to skip invalid parts, when setting the next cursor.
Context menus are only shown for valid parts.


RasmusAlbrektsen (approved):

MadsThy (approved):
LGTM

PR #219, opened 2024-06-17
SOF-2025 - Spanning until segment end

MadsThy:


aardhuis:
I marvel at its perfection!

aardhuis (approved):

src/app/core/models/piece.ts
```diff
@@ -8,6 +8,7 @@ export interface Piece {
   readonly start: number
   readonly duration?: number
   readonly isPlanned: boolean
+  readonly isSpanning?: boolean
```
KvelaGorrrrnio: Since this property is a computed attribute that is specific to the frontend, I would suggest that it does not live on the Piece interface.
I would argue that it is unnecessary, since it seems to only be used, when the piece is seen in the context of a part, that is not the owner of the piece. For that use case, the lifespan and partId can be used to derive the same information.

src/app/rundown-view/components/scrollable-timeline/scrollable-timeline.component.ts
```diff
@@ -52,6 +59,24 @@ export class ScrollableTimelineComponent {
...
+  private getIDPositions(parts: readonly Part[]): { partID: number; pieceID: number } {
+    let pieceID = -1
+    return {
+      partID: parts.findIndex(part => {
+        pieceID = part.pieces.findIndex(piece => piece.lifespan === PieceLifespan.SPANNING_UNTIL_SEGMENT_END)
+        return pieceID !== -1
+      }),
+      pieceID,
+    }
+  }
```
KvelaGorrrrnio: A convention we have in the team is to use camelCase for variables and attributes. I know that it is common that `Id` can also be written as `ID`, but to keep it consistent with the rest of the codebase, I suggest that `Id` is used over `ID`.

A comment on the use of the term id in this context. Both parts and pieces have a id, which is UUID string, but what is used here is the index in relation to the parent entity.

src/app/rundown-view/components/scrollable-timeline/scrollable-timeline.component.ts
KvelaGorrrrnio, line 1: Note that this should also be applied for the timeline component `FollowPlayheadTimelineComponent`.

src/app/rundown-view/components/scrollable-timeline/scrollable-timeline.component.ts
KvelaGorrrrnio: This solution only takes into account the first found spanning piece, but it should take into account all spanning pieces in a segment; and also take into account that a spanning piece can be ended by another spanning piece on the same layer (`Piece.layer`).
For example if I have a spanning graphics piece on the the first part and another spanning graphics piece on the second part, then the third part should show the spanning piece created in the second part; and the second part should not show the first spanning piece. (There might be an edge-case where the spanning graphics piece in the second part have a start offset greater than 0). It should probably be discussed with our PM.

One thing that could be important to note here is whether the `SPANNING_UNTIL_RUNDOWN_END` pieces should be displayed in the current segment or not. That is a question for our PM as well :D

From the known parts of the desired behavior, I would suggest to use a reducer to iterate over all parts and add the spanning pieces to the next parts (and next iteration). I have tried to outline it in the code below. Note that it can be split up into smaller methods to make it a bit more manageable :) Disclaimer: The code has not been tested, so treat it as pseudo-code.

```typescript
  private getAugmentedParts(): Part[] {
    return this.segment.parts.reduce(
      ({ parts, spanningPieces }: { parts: Part[], spanningPieces: Piece[] }, part: Part) => {
        // Filter out spanning pieces overwritten by pieces in the current part.
        const currentPieceLayers: string[] = part.pieces.map(piece => piece.layer)
        const spanningPiecesToAddToPart: Piece[] = spanningPieces.filter(piece => !currentPieceLayers.includes(piece.layer))

        // Update the spanning pieces based on the spanning pieces of the current part.
        const currentSpanningPieces: Piece[] = this.getSpanningUntilEndOfSegmentPieces(part)
        const currentSpanningPieceLayers: string[] = spanningPieces.map(piece => piece.layer)
        const newSpanningPieces: Piece[] = spanningPieces
          .filter(piece => !currentSpanningPieceLayers.includes(piece.layer))
          .concat(currentSpanningPieces)

        // Add the filtered spanning pieces to the current part and pass the new spanning pieces to the next iteration.
        return {
          parts: [...parts, { ...part, pieces: [...part.pieces, ...spanningPiecesToAddToPart] }],
          spanningPieces: newSpanningPieces
        }
      },
      { parts: [], spanningPieces: [] }
    ).parts
  }

  private getSpanningUntilEndOfSegmentPieces(part: Part): Piece[] {
    return part.pieces.filter(piece => piece.lifespan === PieceLifespan.SPANNING_UNTIL_SEGMENT_END)
  }
```

src/app/rundown-view/components/scrollable-timeline/scrollable-timeline.component.ts
```diff
@@ -52,6 +59,24 @@ export class ScrollableTimelineComponent {
     )
   }
 
+  private copySpanningElement(parts: readonly Part[], partID: number, pieceID: number): void {
+    const pieceToCopy = parts[partID].pieces[pieceID]
+    for (let i = partID + 1; i < parts.length; i++) {
+      parts[i].pieces.push({ ...pieceToCopy, isSpanning: true })
```
KvelaGorrrrnio: We try to keep the state objects immutable, to keep the computations on them deterministic. I don't know why `Part.pieces` is not marked as readonly, but i should be to discourage mutations.
The parts and piece information can be used to calculate timing information for the show. If we apply visual mutations on the state objects, then it can lead to incorrect timings.

In cases where we want to augment the data to fit a certain visualization or frontend functionality, a new object should be created, e.g. in this case:

```typescript
const pieceToCopy: Piece = { ...parts[partIndex].pieces[partIndex], isSpanning: true }
return parts.map(part => ({
  ...part
  pieces: [...part.pieces, pieceToCopy]
})
```
Note that I return the newly created parts here to avoid side effects from the method :)

RasmusAlbrektsen:
Remember that this is a bugfix that should be targeted towards the Staging branch.

RasmusAlbrektsen:
The tooltips for spanned pieces (not the original) are all used for the piece tooltips. This results in them having a tooltip including the information for all the trailing pieces, as can be seen in the image i've attached. Whether this is something that is handed down incorrectly or that we somehow group a lot of pieces on each part is the question. 
image

src/app/rundown-view/components/scrollable-timeline/scrollable-timeline.component.ts
MadsThy: Noted :) Changing to index

src/app/core/models/piece.ts
MadsThy: Removed from piece interface

MadsThy:
Closing this PR as it was branched out from the wrong branch to begin with :) Opening new one here:

https://github.com/tv2/sofie-web-client/pull/220

PR #220, opened 2024-06-18
SOF-2025 - Bug: Span piece until segment end

MadsThy:
Redid the branch so it could be based off of staging instead, since this is where it will be merged to

This bugfix augments a piece in a segment before it is sent down the component tree, thus adding a "isSpanned" boolean to the object.

As it is visible in the screenshot, if a segment has more than 1 "span until segment end" piece, it will stop once it reaches the next one in the segment, and allow the next piece to be the one that is spanned.

One of the important things to notice is that the label in the piece is not shown when it is spanned.

image

### DIFF
--- a/src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
+++ b/src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
@@ -1,11 +1,13 @@
 import { ChangeDetectionStrategy, Component, Input, OnChanges, SimpleChange, SimpleChanges } from '@angular/core'
-import { Part } from '../../../core/models/part'
-import { Segment } from '../../../core/models/segment'
-import { PartEntityService } from '../../../core/services/models/part-entity.service'
 import { RundownService } from '../../../core/abstractions/rundown.service'
-import { Tv2OutputLayer } from '../../../core/models/tv2-output-layer'
 import { Tv2PieceType } from '../../../core/enums/tv2-piece-type'
+import { Part } from '../../../core/models/part'
+import { Piece } from '../../../core/models/piece'
+import { PieceLifespan } from '../../../core/models/piece-lifespan'
+import { Segment } from '../../../core/models/segment'
+import { Tv2OutputLayer } from '../../../core/models/tv2-output-layer'
 import { Tv2PieceMetadata } from '../../../core/models/tv2-piece'
+import { PartEntityService } from '../../../core/services/models/part-entity.service'
 
 const PRE_PLAYHEAD_INSET_IN_PIXELS: number = 40
 const POST_PLAYHEAD_INSET_IN_PIXELS: number = 200
@@ -17,11 +19,18 @@ const POST_PLAYHEAD_INSET_IN_PIXELS: number = 200
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class FollowPlayheadTimelineComponent implements OnChanges {
-  @Input()
-  public time: number
+  public get segment(): Segment {
+    return this._segment
+  }
+
+  @Input() public set segment(segment: Segment) {
+    this._segment = { ...segment, parts: this.checkForSpanningElements(segment.parts) }
+  }
+
+  private _segment: Segment
 
   @Input()
-  public segment: Segment
+  public time: number
 
   @Input()
   public outputLayers: Tv2OutputLayer[]
@@ -75,6 +84,19 @@ export class FollowPlayheadTimelineComponent implements OnChanges {
       const segment: Segment = segmentChange.currentValue
       this.shouldShowPartCountdown = segment.parts.some(part => this.doesPartContainVideoClipOrVoiceOver(part))
     }
+  }
+
+  private checkForSpanningElements(parts: readonly Part[]): Part[] {
+    return parts.reduce((result: Part[], currentPart: Part, index: number) => {
+      const previousPart = result[index - 1]
+      const prevPieceWithSpanElement = previousPart?.pieces.findIndex(piece => piece.lifespan === PieceLifespan.SPANNING_UNTIL_SEGMENT_END)
+      const currPieceHasSpanElement = currentPart.pieces.some(piece => piece.lifespan === PieceLifespan.SPANNING_UNTIL_SEGMENT_END)
+
+      if (prevPieceWithSpanElement >= 0 && !currPieceHasSpanElement) {
+        currentPart.pieces.push({ ...previousPart.pieces[prevPieceWithSpanElement], isSpanned: true } as Piece)
+      }
+      return [...result, currentPart]
+    }, [])
   }
 
   private getPreviousParts(onAirPartIndex: number): Part[] {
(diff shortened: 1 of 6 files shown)


RasmusAlbrektsen (approved):

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
```diff
@@ -77,6 +86,19 @@ export class FollowPlayheadTimelineComponent implements OnChanges {
     }
   }
 
+  private checkForSpanningElements(parts: readonly Part[]): Part[] {
+    return parts.reduce((result: Part[], currentPart: Part, index: number) => {
```
LindvedKrvang, line 90: What's the reason for using `reduce` and not `map` here?

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
```diff
@@ -77,6 +86,19 @@ export class FollowPlayheadTimelineComponent implements OnChanges {
     }
   }
 
+  private checkForSpanningElements(parts: readonly Part[]): Part[] {
```
LindvedKrvang, line 89: There is a mismatch between the name of the function and the return type. 

`checkFor` indicates more that it should return a boolean confirming if it has a spanning element or not. Not that it should return an array of Parts.

`elements` is quite vague. What elements are it checking for? Rundowns? Segments? Parts? Since the return type is `Part[]` it indicates the elements in question might also be Parts.

The name doesn't really indicate what happens if the check is true. You modify Pieces in this function, but that is not indicated by the name at all.

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.html
```diff
@@ -5,19 +5,12 @@
   [class.media-unavailable]="isMediaUnavailable()"
   sofieTooltip
   [sofieTooltipTemplate]="pieceTooltip"
-  (tooltipMetadataChange)="updatePositionInVideo($event)"
->
+  (tooltipMetadataChange)="updatePositionInVideo($event)">
   <div class="label" [style.margin-left.px]="labelOffsetInPixels">
-    <span #labelTextElement>{{ piece.name }}</span>
+    <span *ngIf="piece.isSpanned !== true" #labelTextElement>{{ piece.name }}</span>
```
LindvedKrvang, line 10: Could be written as `*ngIf="!piece.isSpanned`

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
```diff
@@ -1,13 +1,16 @@
...
+import { Tv2AudioMode } from '../../../core/enums/tv2-audio-mode'
+import { Piece } from '../../../core/models/piece'
+import { Tv2Piece } from '../../../core/models/tv2-piece'
 import { TooltipMetadata } from '../../../shared/directives/tooltip.directive'
+import { Media } from '../../../shared/services/media'
+import { MediaStateService } from '../../../shared/services/media-state.service'
 
 const LABEL_TEXT_INSET_IN_PIXELS: number = 14
+type AugmentedPiece = Piece & {
+  isSpanned?: boolean
```
LindvedKrvang, line 12: We have two types of "spanning" Pieces. They can span the entire Segment, or they can span the entire Rundown. This doesn't indicate which it is and could be either.
I suggest calling it `isSpanningSegment` since that is the value it is storing.

src/app/rundown-view/components/scrollable-timeline/scrollable-timeline.component.ts
```diff
@@ -52,6 +54,19 @@ export class ScrollableTimelineComponent {
     )
   }
 
+  private checkForSpanningElements(parts: readonly Part[]): Part[] {
```
LindvedKrvang, line 57: Same comment for this as the other place you have this method.

Also, you already have this method... 
We should avoid duplicated code as much as possible.

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
```diff
@@ -77,6 +86,19 @@ export class FollowPlayheadTimelineComponent implements OnChanges {
...
+    return parts.reduce((result: Part[], currentPart: Part, index: number) => {
+      const previousPart = result[index - 1]
+      const prevPieceWithSpanElement = previousPart?.pieces.findIndex(piece => piece.lifespan === PieceLifespan.SPANNING_UNTIL_SEGMENT_END)
+      const currPieceHasSpanElement = currentPart.pieces.some(piece => piece.lifespan === PieceLifespan.SPANNING_UNTIL_SEGMENT_END)
+
+      if (prevPieceWithSpanElement >= 0 && !currPieceHasSpanElement) {
+        currentPart.pieces.push({ ...previousPart.pieces[prevPieceWithSpanElement], isSpanned: true } as Piece)
```
LindvedKrvang, line 96: You should avoid the `as` keyword as much as possible in production code. You are basically telling Typescript to avoid using types.

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.html
MadsThy: It could be, I just personally thought that it was more telling that I was expecting to check for the true statement :) After having looked at !piece.isSpanned for the whole period I was developing, I ended up choosing the semantic "if piece isSpanned is not true" instead of "if not piece isSpanned".

I could argue for both sides, but will leave it up to the team to decide about this :)

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
MadsThy: This was mainly due to the input from @KvelaGorrrrnio and the nature of reduce giving me "better" built-in access to the previous parts, instead of having to use the array reference in map. Do you have a preference? The same could be achieved using map, just with slight modifications

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
MadsThy: Agreed :) Do you have a preference for how to send down augmented items through the component tree? I was not allowed to augment the original Piece type 😄

src/app/rundown-view/components/offsetable-piece/offsetable-piece.component.ts
MadsThy: Agreed :) This was mainly due to the low amount of domain knowledge, so thank you for the clarification. Will create a new PR with this change in.

src/app/rundown-view/components/scrollable-timeline/scrollable-timeline.component.ts
MadsThy: Agreed :) Would you suggest i create a new service for this, or put it in one of these existing ones?

- PartEntityService,
- RundownService

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
MadsThy: Opting for findAndInsertSpanningParts then :) That acceptable?

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
MadsThy: I will include a change to the function using `map` instead :) See if you like that better

PR #227, opened 2024-06-28
Sof 1919 use virtual keyboard on settings screen

aardhuis:
There was a tight coupling between the virtual keyboard and the rundown when it came from the output from the keyboard, so I emitted an output from the virtual keyboard and subscribed to that in the component.


src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
```diff
@@ -20,6 +20,10 @@ export class VirtualKeyboardComponent implements OnChanges {
   @Input()
   public keystrokes: string[]
 
+  @Output() public keyPressed = new EventEmitter<string[]>()
+
+  // keypressed.emit(somearray)
```
MadsThy:
```suggestion
```

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -1,12 +1,47 @@
...
+        errorMessage="action-triggers.shortcut.error">
       </sofie-keyboard-input>
-      <sofie-keyboard-input i18n-label label="action-triggers.physical-mapping.label" [formControlName]="'mappedToKeys'" i18n-helpText helpText="action-triggers.physical-mapping-help.tooltip">
+      <sofie-button-group>
+        <sofie-button #toggleButton (click)="toggleKeyboardVisibility()" i18n>
+          edit-keyboard-mapping.toggle-virtual-keyboard.button
+        </sofie-button>
+        <sofie-spacer></sofie-spacer>
+      </sofie-button-group>
+      <div  class="wrapper">
```
MadsThy:
```suggestion
      <div class="wrapper">
```

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
```diff
@@ -6,3 +6,38 @@
...
+
+.floating-keyboard {
+  position: absolute;
+  top: 250px;
+  left: 25%;
+  width: 74%;
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+  flex: 1 1 100%;
```
MadsThy:
```suggestion
```
Er denne nødvendig? :)

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
```diff
@@ -6,3 +6,38 @@
...
+  min-width: 0;
+  z-index: 100;
+}
+
+.sofie-virtual-keyboard {
+  width: auto;
+  height: auto;
```
MadsThy:
```suggestion
```
Dem her burde du kunne klare dig uden :) Width og height burde gerne stå på auto som standard

src/app/settings/keyboard-mapping-settings/keyboard-mapping-settings.module.ts
```diff
@@ -10,8 +11,8 @@ const routes: Routes = [{ path: '', component: KeyboardMappingSettingsPageCompon
 
 @NgModule({
   declarations: [KeyboardMappingSettingsPageComponent, EditKeyboardMappingDialogComponent, EditKeyboardMappingComponent],
-  imports: [SharedModule, RouterModule.forChild(routes), ProducerShelfModule],
-  providers: [],
+  imports: [SharedModule, KeyboardModule, RouterModule.forChild(routes), ProducerShelfModule],
+  providers: [], //{ provide: KeyboardConfigurationService, useClass: KeyboardMappingSettingsKeyboardConfiguration }
```
MadsThy:
```suggestion
  providers: [],
```

MadsThy (approved):

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -1,12 +1,47 @@
...
+      <sofie-keyboard-input
+        [isRequired]="true"
+        i18n-label
+        label="action-triggers.shortcut.label"
+        [formControlName]="'keys'"
+        i18n-errorMessage
+        errorMessage="action-triggers.shortcut.error">
       </sofie-keyboard-input>
-      <sofie-keyboard-input i18n-label label="action-triggers.physical-mapping.label" [formControlName]="'mappedToKeys'" i18n-helpText helpText="action-triggers.physical-mapping-help.tooltip">
+      <sofie-button-group>
```
LindvedKrvang: You only have a single button, so you shouldn't be needing a button-group

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -1,12 +1,47 @@
...
+        <sofie-button #toggleButton (click)="toggleKeyboardVisibility()" i18n>
+          edit-keyboard-mapping.toggle-virtual-keyboard.button
+        </sofie-button>
+        <sofie-spacer></sofie-spacer>
+      </sofie-button-group>
+      <div class="wrapper">
+        <div *ngIf="isKeyboardVisible" class="floating-keyboard">
+          <sofie-virtual-keyboard class="sofie-virtual-keyboard" (keyPressed)="buildKeyCombo($event)" [keyBindings]="keyBindings" [keystrokes]="keystrokes"></sofie-virtual-keyboard>
+        </div>
+      </div>
```
LindvedKrvang: The UI could be looking a bit nicer:
image
There is now this large "gap" between your toggle button and the next input, which doesn't really make sense. It would be nice to if we could have the same distance in the gap between inputs. It gives it a more clean and "neat-looking" feeling.

Your toggle button also looks out of place.
It would be nice if it was icon (maybe the questionmark icon the others have - or better yet, a keyboard icon) that was placed above the input to the right (like with "Override key color").
When hovered, you should display "Toggle virtual keyboard" in a tooltip and when clicked, you should open the keyboard. Remember, this feature is to be used in an edgecase, so it shouldn't be taking up a significant portion of the UI.

Also, it would be nice that when you click outside of the virtual keyboard that it will close itself again. That' going to be even more necessary when the toggle isn't this big button, but a small icon instead.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
```diff
@@ -6,3 +6,35 @@
 .selected-action {
   font-weight: lighter;
 }
+
+.wrapper {
+  display: inline-flex;
+  width: 100%;
+  padding: 20px 10px 20px 20px;
```
LindvedKrvang: We have some CSS classs to use for getting margins and paddings (they were introduced at a later staged so unfortunately there is still places that gets their margin/padding from other CSS classes).

Instead of setting the padding in here, you should use the classes `m-1, mt-2, p-3, pb-4` etc.
So in your HTML it would look something like this:
```
<div class="p-3 pt-2">
```

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -1,12 +1,47 @@
...
       </sofie-keyboard-input>
-      <sofie-keyboard-input i18n-label label="action-triggers.physical-mapping.label" [formControlName]="'mappedToKeys'" i18n-helpText helpText="action-triggers.physical-mapping-help.tooltip">
+      <sofie-button-group>
+        <sofie-button #toggleButton (click)="toggleKeyboardVisibility()" i18n>
+          edit-keyboard-mapping.toggle-virtual-keyboard.button
+        </sofie-button>
+        <sofie-spacer></sofie-spacer>
+      </sofie-button-group>
+      <div class="wrapper">
+        <div *ngIf="isKeyboardVisible" class="floating-keyboard">
```
LindvedKrvang: You don't need the `wrapper` the div.
It creates a large "gap" in the UI that makes the UI look less structed because there is a random gap in the middle of it.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -65,6 +75,16 @@ export class EditKeyboardMappingComponent implements OnInit {
     }
   }
 
+  public buildKeyCombo($event: string[]): void {
```
LindvedKrvang: Methods shouldn't know anything about what is happening outside of its own scope. That means that method-, parameter- and variable-names shouldn't refer to things outside of the method.

You have named your parameter `$event`, but `buildKeyCombo` doesn't know that it is being used in a flow where there is an event. Also, there is nothing stopping me from writing:
```
public someMethodThatIsNotTriggeredByAnEvent(): void {
  const keys: string[] = ['KeyA']
  this.buildKeyCombo(keys)
}
```
You have named your parameter after _what you intend to parse to your method_ instead of _what is the parameter_. The parameter is an array consisting of keys, so it should simply be called `keys`. That it's currently being called by an event is not information that the method is privy to.

Sidenote: I would expoect that somethind calleed `buildX` returns something (just as I expect it for `createX`), but you return void.
Your method isn't really building anything. It's updating the "keys form" with new keys. So a possible name for it could be "updateKeys()" or something similar.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -65,6 +75,16 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+    var currentValue: string[] = this.actionTriggerDataForm.get('keys')?.value ?? []
+
+    $event.forEach(item => {
+      if (item !== undefined && !currentValue.includes(item)) {
```
LindvedKrvang: When will `item` ever be undefined? 
`$event` is a `string[]`, so it's not possible for `$event` to contain an undefined value, hence this check is redundant.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
```diff
@@ -6,3 +6,35 @@
...
+  width: 74%;
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+  min-width: 0;
+  z-index: 100;
+}
+
+.sofie-virtual-keyboard {
+  display: flex;
```
RasmusAlbrektsen, line 32: You can add `color: var(--white-color);` here, as it would greatly improve the appearance since we are not in a "toned down" setting color-wise.

src/app/keyboard/components/virtual-keyboard/virtual-keyboard.component.ts
```diff
@@ -20,6 +20,9 @@ export class VirtualKeyboardComponent implements OnChanges {
   @Input()
   public keystrokes: string[]
 
+  @Output() public keyPressed = new EventEmitter<string[]>()
+
+
```
RasmusAlbrektsen: Extra newline here!

RasmusAlbrektsen (commented):
I would like to suggest something i believe would be a lot prettier than pressing a button to toggle a keyboard. I'd rather we have an icon within the actual field we use the keyboard for. Such an icon can be found in our icon package, an example is this one: https://fontawesome.com/icons/keyboard?f=classic&s=regular

Furthermore, if possible, it would be cool to give it a little "tail" from where it originates. Like in the image below where i quickly drew out an example with everything im talking about:

 
settings keyboard example

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
aardhuis: Concerns addressed and committed to git (but not yet pushed)

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
aardhuis: Concerns addressed and committed to git (but not yet pushed)

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
aardhuis: A tail? What manner of sorcery is this?

My CSS skills aren't that 1337 but I can see what I can do!

RasmusAlbrektsen: Very much in the lack of a better word, i hope the picture shows what i mean! Haha :-D

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
aardhuis: Personal Obsidian "code standard section" updated with this information.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
aardhuis: You mean something like this right?

https://stackoverflow.com/questions/30299093/speech-bubble-with-arrow

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
aardhuis: The way I see it I have two choices here:
1) Put a font-awesome faKeyboard icon behind the label
2) Put a font-awesome faKeyboard icon at the tail end of the input field

If I opt for route 2, you probably want me to create a specialized component that has all that jazz baked into it, amiright, but option one is just adding a font awesome icon behind the label?

aardhuis: I found Sofie-icon-button and will use that

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
aardhuis: (Looking through the codebase we might need some cleanup if we are to be in compliance with this rule)

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
aardhuis: Okay @LindvedKrvang. I have fixed and pushed this (but not everything you commented on). At the moment, since this functionality is a one-off occurrence, I used CSS to position it instead of tweaking the sofie-keyboard-input.component.ts file. Reason: One-off fix and not something that will be used elsewhere.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
aardhuis: @RasmusAlbrektsen I have added the font awesome keyboard icon. Working on the fin, but the icon can be tested now,

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
aardhuis: Wrapper nuked in latest push

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
aardhuis: Ref: Close when clicking outside: 
Currently It is nice to be able to click in the shortcuts textarea and reset it, use a combination if keyboard and virtual keyboard, etc. If we implement a click out that will close the dialog if we attempt to "X out" the current shortcut, etc.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
aardhuis: The color: var(--white-color); was SWEEEET

aardhuis:
Alright bois, your wishes have been added. How do you like them apples?

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -1,12 +1,57 @@
 <sofie-row class="mb-3">
   <form [formGroup]="actionTriggerForm" class="left-container">
-    <div [formGroupName]="'data'">
-      <sofie-text-input i18n-label label="global.label" [formControlName]="'label'"></sofie-text-input>
-      <sofie-keyboard-input [isRequired]="true" i18n-label label="action-triggers.shortcut.label" [formControlName]="'keys'" i18n-errorMessage errorMessage="action-triggers.shortcut.error">
+    <div [formGroupName]="'data'" style="position: relative;">
```
RasmusAlbrektsen: Here and on line 14 as well you use inline styling which we really try to avoid, unless it can be done by our utility classes in: https://github.com/tv2/sofie-web-client/blob/staging/src/scss/_sofie-utility-classes.scss

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -30,6 +32,17 @@ export class EditKeyboardMappingComponent implements OnInit {
...
   @Output() public onCancel: EventEmitter<void> = new EventEmitter()
 
+  public isKeyboardVisible = false
+  protected readonly Icon = Icon
+  protected readonly IconSize = IconSize
+
+  public keystrokes: string[] = []
+  public keyBindings: KeyBinding[] = []
+
+  public toggleKeyboardVisibility(): void {
```
RasmusAlbrektsen: This method can be moved down to the other methods rather than be here.

RasmusAlbrektsen (approved):
Looks good! :-D 
Just a couple of comments but i approve once they're done!

PR #235, opened 2024-08-28
fix: Changes Out of Focus position based on screen and window height

RasmusAlbrektsen:
After testing in staging the label showed weird behaviour, i made a quick fix to take into account the screen size and window height by using the approach we have for dialogs.


LindvedKrvang (approved):

src/app/rundown-view/components/unfocused/unfocused.component.ts
```diff
@@ -19,4 +19,12 @@ export class UnfocusedComponent {
...
     this.isUnfocused = false
   }
+
+  public getCenterPositionForLabel(): string {
+    const isWindowHeightLargerThanCurrentScreenHeight: boolean = window.innerHeight > window.screen.availHeight
+    if (!isWindowHeightLargerThanCurrentScreenHeight) {
+      return `${window.innerHeight / 2}`
+    }
+    return `${window.screen.availHeight / 2}`
+  }
```
MadsThy: Instead of triggering the change detection so much, let's make it use a variable for the margin :D Using function calls will trigger a lot of change detection :)

```suggestion
@HostListener('window:resize')
  public onResize(): void {
    this.marginTop = this.getCenterPositionForLabel()
  }

  public marginTop = 0

  public getCenterPositionForLabel(): number {
    const isWindowHeightLargerThanCurrentScreenHeight: boolean = window.innerHeight > window.screen.availHeight
    if (!isWindowHeightLargerThanCurrentScreenHeight) {
      return window.innerHeight / 2
    }
    return window.screen.availHeight / 2
  }
```

src/app/rundown-view/components/unfocused/unfocused.component.html
```diff
@@ -1 +1 @@
-<span class="label">{{ outOfFocusMessage }}</span>
+<span class="label" [style.margin-top.px]="getCenterPositionForLabel()">{{ outOfFocusMessage }}</span>
```
MadsThy: See other comment for explanation for this change 😄 

```suggestion
<span class="label" [style.margin-top.px]="marginTop">{{ outOfFocusMessage }}</span>
```

MadsThy (approved):

src/app/rundown-view/components/unfocused/unfocused.component.ts
LindvedKrvang: I second this suggestion :)
Only small change: We want to keep our instance variables (`marginTop`) above our functions and constructor.

KvelaGorrrrnio (approved):

src/app/rundown-view/components/unfocused/unfocused.component.ts
RasmusAlbrektsen: It has been implemented! Thanks for the great suggestion :-)

src/app/rundown-view/components/unfocused/unfocused.component.html
RasmusAlbrektsen: It has been changed :-)

PR #238, opened 2024-08-29
SOF-2163: Multiple next cursors

KvelaGorrrrnio:
When a part is marked as next during ingest events, the frontend receives a PART_CREATED event, but did not update the segment and part next-markings.
This edge case is now handled, by checking if the part is marked as next and then treating it as it was inserted as next.


MadsThy (approved):

src/app/core/services/models/rundown-entity.service.ts
```diff
@@ -163,6 +163,10 @@ export class RundownEntityService {
   }
 
   public insertPartInSegment(rundown: Rundown, part: Part): Rundown {
+    if (part.isNext) {
+      return this.insertPartAsNext(rundown, part)
```
LindvedKrvang, line 167: This inserts the Part into the Segment and on line 177 the Part is  inserted into the Segment again, meaning the Part will be inserted twice???

LindvedKrvang: Nevermind, I just noticed the `return` :p

LindvedKrvang (approved):

PR #239, opened 2024-08-30
SOF-2146: Rebrand to Alba

KvelaGorrrrnio:
Removes Sofie logo and replaces `Sofie` with `Alba` in the UI.
Also includes fix for inconsistent page title retrieval for the logo text in the rundown overview.

No logo for Alba has been added yet (besides the text `Alba`).

Renaming the components and services from Sofie to Alba is not done in here.
The suggestion is to do this after first release.


src/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="utf-8">
-  <title>Sofie</title>
+  <title>Alba</title>
   <base href="/">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="icon" type="image/x-icon" href="assets/sofie-logo.svg">
```
MadsThy: Im guessing it's fine we leave the favicon in right? :)

MadsThy (approved):

src/index.html
KvelaGorrrrnio: Nice catch :) I have removed it and added a TODO to readd it when we have an icon for it. There is no reason to keep it, since it is pointing to a deleted asset.

RasmusAlbrektsen (approved):

PR #240, opened 2024-08-30
SOF-2188: Prevent default browser key behavior

KvelaGorrrrnio:
Removes check of whether observed key events should prevent default behavior. Instead all key events that targets the observed event area have their default behavior prevented.
Note: This can give issues if text input fields are added to the observed evet area.


RasmusAlbrektsen (approved):

MadsThy (approved):

LindvedKrvang (approved):
Deleted code, what's not to like? :D

PR #244, opened 2024-09-04
SOF-2197 No longer reset the next Part if it is unplanned when receiving an ingest event

LindvedKrvang:
Everytime we got an ingest event we ran a flow that would remove all unplanned Parts from a Segment only taking into consideration if it was the OnAir Part. 
Now we also take into consideration if the unplanned Part is next.

Note: I have added the method `removeUnplannedUnplayedPartsAndPiecesExceptNextPart` because the method `removeUnplannedUnplayedPartsAndPieces` is used by `insertPartAsNext` and that method _do want_ to remove the next Part if it is unplanned. 
Due to the functional programming approach taking here, this was the best solution I could come up with. Suggestions are welcome :)


MadsThy (approved):

src/app/core/services/models/segment-entity.service.ts
```diff
@@ -95,17 +95,31 @@ export class SegmentEntityService {
...
+      parts: segment.parts.filter(partFilter).map(part => (part.isNext && part.isPlanned ? this.partEntityService.reset(part) : part)),
     }
   }
 
+  private insertElementAtIndex<T>(elements: T[], element: T, index: number): T[] {
+    return [...elements.slice(0, index), element, ...elements.slice(index)]
+  }
+
+  public removeUnplannedUnplayedPartsAndPiecesExceptNextPart(segment: Segment): Segment {
+    const partFilter: (part: Part) => boolean = (part: Part) => part.isOnAir || part.isPlanned || part.playedDuration > 0 || this.isPartUnplannedAndNext(part)
```
KvelaGorrrrnio: Is there a scenario where we want to remove a part marked as next, no matter if it is planned or not?
If not, then I suggest that we just check if `part.isNext`. Based on the condition `part.isPlanned`, we also know that it is unplanned in any subsequent or-clauses.

KvelaGorrrrnio (approved):

src/app/core/services/models/segment-entity.service.ts
LindvedKrvang: I don't think there is. 
The functionality still works with your suggested change, so I have updated it.

RasmusAlbrektsen (approved):

KvelaGorrrrnio (approved):

PR #245, opened 2024-09-06
fix: Stop jittery MiniShelf's when missing thumbnail

RasmusAlbrektsen:
I noticed that thumbnails for the MiniShelf are continuously tried to default to the deleted svg when it cannot be retrieved from Media manager. This caused it to jitter in the Rundown View. 

Also: 
I just noticed this whilst doing something else and remembered Jacob talking about it, is it necessary for something this small to be created as a bug task to be reflected sprint, or is it okay to just make a quick fix like this?


src/app/rundown-view/components/mini-shelf/mini-shelf.component.ts
```diff
@@ -123,7 +118,7 @@ export class MiniShelfComponent implements OnInit, OnDestroy, OnChanges {
   }
 
   protected handleMissingImage(event: Event): void {
-    ;(event.target as HTMLImageElement).src = this.fallbackPreviewUrl
+    ;(event.target as HTMLImageElement).alt = 'Thumbnail not found.'
```
MadsThy, line 121: i dont think i've ever seen JS code with a beginning ; 😄 What is that used for?

MadsThy (approved):

MadsThy:
> Also: I just noticed this whilst doing something else and remembered Jacob talking about it, is it necessary for something this small to be created as a bug task to be reflected sprint, or is it okay to just make a quick fix like this?

I am usually of the opinion that if it is a task that is the size of "I can fix this in my coffee break", then you just do it without a task :) But you could make a task and put it in "done" to showcase that you have actually fixed something :)

RasmusAlbrektsen:
PR is closed since Anders found the bug as well and is making a fix with a related task! A good answer to my own question of why this is a good thing to reflect on the board as a task i guess :-D

KvelaGorrrrnio:
#247

PR #252, opened 2024-09-11
SOF-2207: Action panel playhead

KvelaGorrrrnio:
Adds playhead to action panels indicating the show progression in relation to the parts that actions are generated from.

The action panels scroll to follow the playhead when it changes, highlights actions that are generated from the on air part and dims actions from parts prior the the on air part.

The scroll behavior clashed with the scroll behavior when scrolling to a segment in the rundown view. This was caused because scrolling to a segment used scrolling for the entire page, instead of targeting the rundown viewport. This is now fixed by having scrolling to segments target the rundown viewport.

image


src/app/producer-shelf/components/tv2-action-panel/tv2-action-panel.component.ts
```diff
@@ -1,29 +1,102 @@
...
+  }
+
+  private getActionIdNearestToOnAirPart(actionRankForOnAirPart: number | undefined): string | undefined {
+    if (actionRankForOnAirPart === undefined) {
+      return undefined
+    }
+    return this.actions.find(action => action.rank >= actionRankForOnAirPart)?.id
+  }
+
+  private getIsPastLastAction(actionRankForOnAirPart: number | undefined): boolean {
```
LindvedKrvang: You are returning a boolean value so the `get` is technically not needed. `isPastLastAction` gives a better indication that the answer is a boolean.

LindvedKrvang (approved):

src/app/producer-shelf/components/tv2-action-panel/tv2-action-panel.component.ts
KvelaGorrrrnio: The method is used for checking if we are past the last action, such that it can be assigned to the variable `isPastLastAction`. I have renamed it to `updateIsPastLastAction` and moved the assignment into it.

RasmusAlbrektsen (approved):

PR #258, opened 2024-09-23
SOF-2276 Listen for BulkIngestEvent

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-server/pull/230

Now listens for the BulkIngestEvent.
When we received the event, we iterate over all the event and call the methods that would have beeen called if the Events weren't being bulked.


MadsThy (approved):

RasmusAlbrektsen (approved):

src/app/core/models/rundown-event.ts
```diff
@@ -125,3 +125,8 @@ export interface PartUnsyncedEvent extends RundownEvent {
   type: RundownEventType.PART_UNSYNCED
   part: Part
 }
+
+export interface BulkIngestEvent extends RundownEvent {
+  type: RundownEventType.BULK_INGEST_UPDATES
+  ingestEvents: RundownEvent[]
```
KvelaGorrrrnio: If dedicated type is introduced in the backend (`IngestEvent`), then it would be nice if it was used here as well.

src/app/core/parsers/zod-rundown-event-parser.service.ts
```diff
@@ -235,6 +236,13 @@ export class ZodRundownEventParser implements RundownEventParser {
       .transform((part: unknown) => this.entityParser.parsePart(part)),
   })
 
+  private readonly bulkIngestEventParser = zod.object({
+    type: zod.literal(RundownEventType.BULK_INGEST_UPDATES),
+    timestamp: zod.number(),
+    rundownId: zod.string().min(1),
+    ingestEvents: zod.object({}).passthrough().array(),
```
KvelaGorrrrnio: Instead of accepting all objects, we can compile a list of events supported, e.g.
```typescript
{
  ingestEvents: this.partDeletedEventParser
    .or(this.partUpdatedEventParser)
    .or(this.partCreatedEventParser)
    // rest of the parsers
    .array()
}
```
This way we won't have to cast the type in `parseBulkIngestEvent`.

src/app/core/models/rundown-event.ts
LindvedKrvang: Updated

src/app/core/parsers/zod-rundown-event-parser.service.ts
LindvedKrvang: This was the comment I was hoping to receive :D 
It's been updated

KvelaGorrrrnio (approved):

PR #259, opened 2024-09-24
Add sofie settings to settings page

KvelaGorrrrnio:
Adds Sofie Core settings in an iframe as the menu item _Sofie settings_ under the Settings page. While doing this, the Clear configuration cache page is merging into the Sofie settings page.

image

Since there is no ambuiguity between sofie and old sofie, after renaming the product to Alba, oldSofie references are renamed to sofie.


src/app/settings/components/sofie-settings/sofie-settings.component.ts
```diff
@@ -1,22 +1,29 @@
...
+import { DialogService } from '../../../shared/services/dialog.service'
 import { NotificationService } from '../../../shared/services/notification.service'
-import { DialogColorScheme, DialogSeverity } from 'src/app/shared/components/confirmation-dialog/confirmation-dialog.component'
+import { DialogColorScheme, DialogSeverity } from '../../../shared/components/confirmation-dialog/confirmation-dialog.component'
 
 @Component({
-  selector: 'sofie-clear-cache',
-  templateUrl: './clear-cache.component.html',
-  styleUrls: ['./clear-cache.component.scss'],
+  selector: 'sofie-sofie-settings',
```
LindvedKrvang: Maybe call it `sofie-core-settings` instead of the double Sofie? It looks like a typo ;)

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/app/settings/components/sofie-settings/sofie-settings.component.ts
KvelaGorrrrnio: I have deleted the selector since it is only created programmatically :)

PR #262, opened 2024-09-26
SOF-2288 Show zebra stribes for FullscreenGraphicsActions

LindvedKrvang:
Note: Is dependent on https://github.com/tv2/sofie-server/pull/236
Update: All dependent PRs have been merged.

The zebra stribes was heavily bound to only work for VideoClips.
The zebra stribes has been extended to also work for FullscreenGraphicsActions. In order for the stribes for work for the FullscreenGraphicsActions, the Actions will need a `sourceName` (provided in the linked PR above).


RasmusAlbrektsen (approved):

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
```diff
@@ -63,18 +62,45 @@ export class Tv2ActionCardComponent implements OnInit, OnDestroy {
...
+    if (this.isVideoClipAction()) {
+      const videoClipAction: Tv2VideoClipAction = this.action as Tv2VideoClipAction
+      return videoClipAction.metadata.fileName
+    }
+    if (this.isGraphicsAction()) {
+      const graphicsAction: Tv2GraphicsAction = this.action as Tv2GraphicsAction
+      return graphicsAction.metadata.sourceName!
+    }
+    throw new Error("Trying to fetch Media SourceName from an Action that doesn't have Media")
+  }
```
KvelaGorrrrnio: Just a thought:
If the `isGraphicsAction` and `isVideoClipAction` are changed to taking an action, then the signature can be changed to the signature below, so we don't have to make temporary variables that uses type assertions.
```typescript
class Tv2ActionCardComponent {
  private isVideoClipAction(action: Tv2Action): action is Tv2VideoClipAction { /* implementation */ }
  private isGraphicsAction(action: Tv2Action): action is Tv2GraphicsAction { /* implementation */ }

  private getMediaSourceName(): string {
    if (this.isVideoClipAction(this.action)) {
      return this.action.metadata.fileName
    }
    if (this.isGraphicsAction(this.action) && this.action.metadata.sourceName) {
      return this.action.metadata.sourceName
    }
    throw new Error("Trying to fetch media source name from an action that doesn't have media.")
  }
}
```
This way, it can also easily be extracted to a separate class, so the classification of actions can be used other places as well :)

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
```diff
@@ -63,18 +62,45 @@ export class Tv2ActionCardComponent implements OnInit, OnDestroy {
...
+
+  private getMediaSourceName(): string {
+    if (this.isVideoClipAction()) {
+      const videoClipAction: Tv2VideoClipAction = this.action as Tv2VideoClipAction
+      return videoClipAction.metadata.fileName
+    }
+    if (this.isGraphicsAction()) {
+      const graphicsAction: Tv2GraphicsAction = this.action as Tv2GraphicsAction
+      return graphicsAction.metadata.sourceName!
```
KvelaGorrrrnio: The source name for graphics action metadata is optional. So perhaps this should be part of the condition? Otherwise we might get some funky behavior :)

KvelaGorrrrnio (approved):
🚀

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
LindvedKrvang: I like it. It's been updated :)

src/app/producer-shelf/components/tv2-action-card/tv2-action-card.component.ts
LindvedKrvang: The reason I didn't do it was because of the casting necessary in the condition. With your other suggested changes that becomes a non issue :)

It's been updated.

PR #268, opened 2024-10-01
fix/Allow users to go back to overview when loading

RasmusAlbrektsen:
We've seen issues where we get stuck on the "loading a rundown"-screen, possibly due to excessive amounts of ingest event. To ensure the producer can easily go back and forth when in full screen mode, this button will help as a immediate solution.


src/app/rundown-view/components/rundown-view/rundown-view.component.html
```diff
@@ -1,6 +1,9 @@
 <ng-container *ngIf="isLoadingRundown; else rundownLoaded">
     <div class="removed-rundown">
-        <mat-card-title class="c-loading-rundown">Loading the rundown...</mat-card-title>
+        <mat-card-title class="c-loading-rundown">Loading the rundown...</mat-card-title><br/><br/>
```
MadsThy: Im not entirely a fan of using <br> to add spacing 😄 haha... But if it works it works.

I would suggest using CSS if you can :) add a margin-bottom to the c-loading-rundown class for instance.

MadsThy (approved):

src/app/rundown-view/components/rundown-view/rundown-view.component.html
KvelaGorrrrnio: Agree with @MadsThy  :)
You can add the class `mb-4` (or another fitting value) to achieve it.

KvelaGorrrrnio (approved):

src/app/rundown-view/components/rundown-view/rundown-view.component.html
LindvedKrvang: Atleast don't put it at the end of the line. It's extremely easy to miss it there since you would read the first tag of the line and assume that's the only tag on it.

src/app/rundown-view/components/rundown-view/rundown-view.component.html
```diff
@@ -1,6 +1,9 @@
...
+        <a href="/">
+            <button mat-raised-button>Go back to rundown overview</button>
```
LindvedKrvang: What's the reason for not using a `sofie-button`?

src/app/rundown-view/components/rundown-view/rundown-view.component.html
RasmusAlbrektsen: In all honesty i just copied the existing buttons and changed the name of it. I completely agree and will change it :-)

src/app/rundown-view/components/rundown-view/rundown-view.component.html
RasmusAlbrektsen: I honestly don't know, maybe it's from before we made it? In any case i've changed it to use the `sofie-button`.

src/app/rundown-view/components/rundown-view/rundown-view.component.html
RasmusAlbrektsen: It has been changed.

LindvedKrvang (approved):

PR #270, opened 2024-10-03
refactor: Add title to notifications and use localize for all of them

RasmusAlbrektsen:
This PR makes it so our notifications have titles and message and correctly uses localization.


src/app/rundown-view/services/mini-shelf-cycle.service.ts
```diff
@@ -29,8 +29,8 @@ export class MiniShelfCycleService {
       const nextMiniShelfAction: Action = this.getActionForMiniShelfSegment(nextMiniShelfSegment, rundownActions)
       this.actionService.executeAction(nextMiniShelfAction.id, rundown.id).subscribe()
     } catch (error) {
-      this.logger.data(error).error('Failed cycling forward in mini shelves.')
-      this.notificationService.createErrorNotification('Unable to cycle forward in mini shelves.')
+      this.logger.data(error).error('Failed cycling forwards in mini shelves.')
```
MadsThy, line 32: i think the correct spelling is actually forward and not forwards :D

src/app/rundown-view/services/mini-shelf-cycle.service.ts
```diff
@@ -55,8 +55,8 @@ export class MiniShelfCycleService {
       const nextMiniShelfAction: Action = this.getActionForMiniShelfSegment(nextMiniShelfSegment, rundownActions)
       this.actionService.executeAction(nextMiniShelfAction.id, rundown.id).subscribe()
     } catch (error) {
-      this.logger.data(error).error('Failed cycling backward in mini shelves.')
-      this.notificationService.createErrorNotification('Unable to cycle backward in mini shelves.')
+      this.logger.data(error).error('Failed cycling backwards in mini shelves.')
```
MadsThy, line 58: same with "backward" :)

src/app/shared/components/notification-card/notification-card.component.html
```diff
@@ -1,2 +1,5 @@
-<div class="text">{{ notification.message }}</div>
+<div class="text">
+  <span *ngIf="notification.title" style="display: block; font-weight: 500;">{{ notification.title }}</span>
+  {{ notification.message }}
+</div>
```
KvelaGorrrrnio: Since `Notification.title` is a required attribute, the if statement should probably be moved to a span for the `Notification.message`.

It would also be nice if the styling of the title is moved to the SCSS file :)

src/app/shared/services/notification.service.ts
```diff
@@ -20,37 +20,39 @@ export class NotificationService {
   }
 
   public createNotificationFromStatusMessage(statusMessage: StatusMessage): void {
-    const message: string = `${statusMessage.title}: ${statusMessage.message}` + (statusMessage.statusCode === StatusCode.GOOD ? ` Okay` : '')
+    const message: string = `${statusMessage.message}${statusMessage.statusCode === StatusCode.GOOD ? ` Okay` : ''}`
```
KvelaGorrrrnio: Should we have the ` Okay` suffix. I suggest we remove it and have the backend generate an appropriate message.

Suggestion for the backend code:
```typescript
export class DeviceChangedService implements DataChangeService {
  private convertDeviceToStatusMessage(device: Device): StatusMessage {
    return {
      id: `${DEVICE_STATUS_MESSAGE_PREFIX}${device.id}`,
      statusCode: device.statusCode,
      title: device.name,
      message: this.getDeviceMessage(device)
    }
  }

  private getDeviceMessage(device: Device): string {
    if (device.statusMessage) {
      return device.statusMessage
    }
    if (device.statusCode === StatusCode.GOOD) {
      return 'Device is in a good state.'
    }
    return ''
  }
}
```

KvelaGorrrrnio (approved):

MadsThy (approved):

src/app/rundown-view/services/mini-shelf-cycle.service.ts
RasmusAlbrektsen: I think it's okay to use forwards when im describing the verb of cycling.

src/app/shared/services/notification.service.ts
LindvedKrvang: I vote for removing the `Okay` suffix.
It was added when Notifications was primarily used for Device Statusses. In any other context the `Okay` does not make sense.

src/app/shared/services/notification.service.ts
```diff
@@ -20,37 +20,39 @@ export class NotificationService {
...
     const notification: Notification = {
       id: statusMessage.id,
-      message,
+      title: statusMessage.title,
+      message: message,
```
LindvedKrvang: Minor thing: Since the attribute and the variable has the same name you could simply write `message` and it would be assigned.

But it comes down to preferences ;)

LindvedKrvang (approved):
Much needed refactor :D

src/app/shared/services/notification.service.ts
RasmusAlbrektsen: I changed it to just be `message` :-)

src/app/shared/services/notification.service.ts
RasmusAlbrektsen: I have removed it, however, do you want a PR for the backend now since it is gone?

src/app/shared/components/notification-card/notification-card.component.html
RasmusAlbrektsen: Changed it and moved the styling :-)

src/app/shared/services/notification.service.ts
LindvedKrvang: For DeviceStatusses it would be nice :)

RasmusAlbrektsen: Created it: https://github.com/tv2/sofie-server/pull/250

PR #274, opened 2024-10-15
Minor fixes

KvelaGorrrrnio:
- Allows mouse events through unsynced overlay to enable use of tooltips for underlying pieces.
- Removes unsynced parts from GUI when doing an INSERT_PART_AS_ON_AIR action.
- Ignores unplanned parts when navigating to a new segment, since the unplanned part will no longer exist while the next cursor is generated in the backend.


RasmusAlbrektsen (approved):

LindvedKrvang (approved):
We should consider whether we want to remove the logic of finding the next Part from the frontend and let the backend handle it. (I vote yes, but it should be done on development)

KvelaGorrrrnio:
> We should consider whether we want to remove the logic of finding the next Part from the frontend and let the backend handle it. (I vote yes, but it should be done on development)

I am a bit skeptical about that - but let us have a chat in the team about it :)

PR #276, opened 2024-10-24
SOF-2307: Embed font faces

KvelaGorrrrnio:
Embeds font-faces for Roboto and Material Icons into the source code, instead of requesting them from the internet when running in production (and in development).

The PR also:
- Makes it so that long words like `sommerfugl` breaks to a new line in the virtual keyboard keys.
- The padding on virtual keyboard keys are decreased to allow more content.
- Adds spacing under the title in the static buttons panel equivalent to the other action panels.

image
image


LindvedKrvang (approved):

MadsThy (approved):
LGTM

PR #278, opened 2024-10-25
fix: Use piece duration for piece duration in tooltips.

KvelaGorrrrnio:
The duration shown in tooltips for a piece was the part duration, which would show the wrong time for e.g. transition effects, since they don't have to span the entire part.
The PR also adds a lower limit of 0 for piece display duration to prevent a visual bug where a past piece is displayed as if it has no duration.


src/app/rundown-view/services/tv2-piece-tooltip-content-field.service.ts
```diff
@@ -31,19 +31,19 @@ const TOOLTIP_TIME_FORMAT: string = 'HH?:mm:ss'
...
           { label: NAME_OF_CLIP_LABEL, value: tv2Piece.name },
-          { label: DURATION_LABEL, value: '' + this.timerPipe.transform(partDuration, TOOLTIP_TIME_FORMAT) ?? UNKNOWN_DATA },
+          { label: DURATION_LABEL, value: '' + this.timerPipe.transform(pieceDurationInMs, TOOLTIP_TIME_FORMAT) ?? UNKNOWN_DATA },
           { label: AVAILABLE_FOR_PLAYOUT_LABEL, value: media ? YES_LABEL : NO_LABEL },
         ]
       case Tv2PieceType.JINGLE:
         return [
           { label: NAME_OF_CLIP_LABEL, value: tv2Piece.name },
-          { label: DURATION_LABEL, value: '' + this.timerPipe.transform(partDuration, TOOLTIP_TIME_FORMAT) ?? UNKNOWN_DATA },
+          { label: DURATION_LABEL, value: '' + this.timerPipe.transform(pieceDurationInMs, TOOLTIP_TIME_FORMAT) ?? UNKNOWN_DATA },
```
MadsThy: This suggestion is for both this line, and the change above :)

``` suggestion
          { label: DURATION_LABEL, value: `${this.timerPipe.transform(pieceDurationInMs, TOOLTIP_TIME_FORMAT) ?? UNKNOWN_DATA}` },
```

MadsThy (approved):

RasmusAlbrektsen (approved):

src/app/rundown-view/services/tv2-piece-tooltip-content-field.service.ts
KvelaGorrrrnio: Updated :)

PR #282, opened 2024-10-30
SOF-2223 Update ports to call AlbaTv2Server

LindvedKrvang:
Updated Ports so that we know call `AlbaTv2Server` instead of `AlbaServer`.

This changes does require AlbaTv2Server to be running to make the system works.


src/environments/environment.ts
```diff
@@ -7,9 +7,9 @@ import { LogLevel } from '../app/core/abstractions/logger.service'
 
 export const environment: Environment = {
   production: false,
-  apiBaseUrl: 'http://localhost:3005/api',
+  apiBaseUrl: 'http://localhost:30010/api',
```
MadsThy:
```suggestion
  apiBaseUrl: 'http://localhost:3010/api',
```

MadsThy (approved):

RasmusAlbrektsen (approved):

KvelaGorrrrnio (approved):
Same as what Mads noticed :)

src/environments/environment.ts
LindvedKrvang: image

LindvedKrvang: It's been updated

PR #283, opened 2024-10-31
SOF-2316: Synchronize time with backend

KvelaGorrrrnio:
_**Note**: The PR is dependent on tv2/sofie-server#270 (and tv2/sofie-tv2-server#45)._
To ensure that the producer has the correct timings, based on the server time, the client now synchronizes time with the backend.

This is done with a simplified NTP implementation, in `NtpTimeService`, where the time spent on  the server to respond is neglected. The approximation provides sufficient precision for the time synchronization.

## Important: `timeService.getTime()` should be used instead of `Date.now()`.

The PR also removes throttling for rundown reset.


RasmusAlbrektsen (approved):

src/app/core/services/ntp-time-service.ts
```diff
@@ -0,0 +1,52 @@
...
+import { NtpEventObserver } from './models/ntp-event-observer'
+import { NtpEvent } from '../models/ntp-event'
+import { ConnectionStatusObserver } from './connection-status-observer.service'
+
+const NTP_REQUEST_INTERVAL_IN_MS: number = 10_000
+const MAX_NTP_TIME_OFFSET_SAMPLE_SIZE: number = 15
+const NTP_TIME_OFFSET_SAMPLE_SIZE: number = 3
+
+@Injectable({
+  providedIn: 'root',
```
LindvedKrvang, line 12: `providedIn: 'root'` makes it a singleton that's available to the entire application.
We have generally tried to avoid that.
Is it deliberate that it is here or is it just from the default template of service creation?

LindvedKrvang (approved):

src/app/core/services/models/ntp-event-observer.ts
```diff
@@ -0,0 +1,38 @@
...
+@Injectable()
+export class NtpEventObserver {
+  private readonly logger: Logger
+
+  constructor(
+    private readonly eventObserver: EventObserver,
+    private readonly ntpEventParser: NtpEventParser,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('NtpEventObserver')
```
MadsThy, line 17:
```suggestion
    this.logger = logger.tag(this.constructor.name)
```

Just a suggestion :) There are multiple places in the project where we could do this.

src/app/core/services/ntp-time-service.ts
```diff
@@ -0,0 +1,52 @@
...
+  private updateTimeOffsetsAndTimeDrift(newTimeOffsetInMs: number): void {
+    this.timeOffsetsInMs = this.timeOffsetsInMs.concat(newTimeOffsetInMs).slice(-MAX_NTP_TIME_OFFSET_SAMPLE_SIZE)
+    this.timeDriftInMs = this.timeOffsetsInMs.reduce((timeDriftSumInMs, timeOffsetInMs) => timeDriftSumInMs + timeOffsetInMs, 0) / Math.max(this.timeOffsetsInMs.length, 1)
+  }
+
+  private requestNtpSamples(sampleSize: number): void {
+    Array.from({ length: sampleSize }).forEach(() => this.ntpEventObserver.emitNtpEvent(Date.now()))
+  }
+
+  public getTime(): number {
```
MadsThy, line 49:
```suggestion
  public getTimeWithDrift(): number {
```

Just a suggestion :)

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.spec.ts
```diff
@@ -5,21 +5,20 @@ import { PartEntityService } from '../../../core/services/models/part-entity.ser
...
-    const mockedPartEntityService: PartEntityService = mock<PartEntityService>()
-    const mockedRundownService: RundownService = mock<RundownService>()
-
     await TestBed.configureTestingModule({
       declarations: [FollowPlayheadTimelineComponent],
       imports: [SharedModule],
       providers: [
-        { provide: PartEntityService, useValue: instance(mockedPartEntityService) },
-        { provide: RundownService, useValue: instance(mockedRundownService) },
+        { provide: PartEntityService, useValue: instance(mock()) },
```
MadsThy, line 19: not looking to mock the instances of the services? :) Just mock() ? Didn't know that was an option 😄

MadsThy (approved):

src/app/core/services/ntp-time-service.ts
KvelaGorrrrnio: It is deliberate. Instantiating multiple copies will work, but will make the use of time inconsistent across the application and increase load on the backend.

src/app/core/services/models/ntp-event-observer.ts
KvelaGorrrrnio: In the frontend we don't use `this.constructor.name` since it is obscured when bundled due to variable collision prevention and size reduction :)

src/app/core/services/ntp-time-service.ts
KvelaGorrrrnio: There is nothing in the abstraction that states anything about time drift or NTP, so I will leave it as is. We could make an implementation that calls Date.now() and it would still be a valid implementation - though it won't fit our needs.

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.spec.ts
KvelaGorrrrnio: The type argument has no real effect in our use cases, since we mostly put them directly where they should be used or assign it to a const that has the correct type.

PR #286, opened 2024-11-05
SystemKeyBindingFactory actions work on the newest version of the rundown.

KvelaGorrrrnio:
This prevents race conditions to affect what version of the rundown to work on, e.g. when setting next or doing a take.


KvelaGorrrrnio:
Closing since this is not the root problem that is being solved here, which is that subject references are overwritten in RundownStateService. Another PR will be opened soon.

PR #289, opened 2024-11-07
fix: Clean up key bindings when leaving rundown view.

KvelaGorrrrnio:
Fixes issue where key bindings were triggered for a rundown previously, but no longer, shown in the rundown view.


RasmusAlbrektsen (approved):

LindvedKrvang (approved):

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -35,7 +35,7 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
   private rundown?: Rundown
   private systemKeyBindings: KeyBinding[] = []
   private actionTriggerKeyBindings: KeyBinding[] = []
-  private readonly keyBindingsSubject: Subject<KeyBinding[]>
+  private keyBindingsSubject: Subject<KeyBinding[]>
```
MadsThy: You are assigning this to be a behaviorSubject further down :D I propose this change instead:

```suggestion
  private keyBindingsSubject = new BehaviorSubject<KeyBinding[]>(this.systemKeyBindings)
```

And then you can delete line 59

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
```diff
@@ -277,13 +280,18 @@ export class ActionTriggerProducerKeyBindingService implements KeyBindingService
   }
 
   public destroy(): void {
+    this.systemKeyBindings = []
     this.keyBindingsSubject.complete()
+    this.keyBindingsSubject = new BehaviorSubject(this.systemKeyBindings)
```
MadsThy, line 284:
```suggestion
    this.keyBindingsSubject.next(this.systemKeyBindings)
```

I don't know what purpose it serves to call complete to then directly update it afterwards 😄 Will this change do?

MadsThy (approved):

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
KvelaGorrrrnio: The purpose is that when we call destroy, we should ensure that updates are no longer propagated to old consumers.

src/app/rundown-view/services/action-trigger-producer-key-binding.service.ts
KvelaGorrrrnio: It is updated to suggestion :)

PR #292, opened 2024-11-25
Pieces inserted by actions can be stopped by using the same action that created it from an action panel.

KvelaGorrrrnio:
_**Note:** Is depending on https://github.com/tv2/sofie-server/pull/287_

The PR introduces logic for stopping pieces inserted on air with actions. The actions used in _STATIC BUTTONS_ have weird behavior and does not mimic was in Sofie, so they do no exhibit the toggle-behavior that the other action panels have.

While implementing this, it was observed that we don't have the necessary information for resetting pieces on set next. A separate PR will be made for fixing this.


RasmusAlbrektsen (approved):

src/app/core/services/rundown-state.service.ts
```diff
@@ -375,6 +377,15 @@ export class RundownStateService implements OnDestroy {
     rundownSubject.next(rundownWithPiece)
   }
 
+  private stopPieceFromEvent(event: RundownPieceStoppedEvent): void {
+    const rundownSubject = this.getRundownSubject(event.rundownId)
+    if (!rundownSubject || !rundownSubject.value) {
```
MadsThy, line 382:
```suggestion
    if (!rundownSubject?.value) {
```
Just a suggestion 😄

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
```diff
@@ -5,12 +5,18 @@
...
+          [actionIdsWithPiecesOnAir]="actionIdsWithPiecesOnAir"
+          (executeAction)="executeOrStopActionBasedOnPiecesOnAir($event)"
         ></sofie-tv2-action-panel>
         <sofie-virtual-keyboard [keyBindings]="keyBindings" [keystrokes]="keystrokes"></sofie-virtual-keyboard>
     </div>
 
     <div class="sidebar">
-        <sofie-tv2-action-panel title="STATIC BUTTONS" [actions]="staticButtonActions" (executeAction)="executeAction($event)" [isScrollable]="false"></sofie-tv2-action-panel>
+        <sofie-tv2-action-panel
+          title="STATIC BUTTONS"
```
MadsThy, line 16:
```suggestion
          [title]="'STATIC BUTTONS'"
```

Just a suggestion to put the input notation on :)

MadsThy (approved):

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
KvelaGorrrrnio: Why should it be dynamic, when the content is static? :)

src/app/core/services/rundown-state.service.ts
KvelaGorrrrnio: Good suggestion :+1: It is updated!

src/app/core/services/models/part-entity.service.spec.ts
```diff
@@ -332,11 +333,15 @@ describe(PartEntityService.name, () => {
...
 
       const result: Part = testee.reset(part)
 
       expect(result.pieces).toEqual(expectedPieces)
     })
   })
 })
+
+function createTestee(params: { pieceEntityService?: PieceEntityService } = {}): PartEntityService {
+  return new PartEntityService(params.pieceEntityService ?? new PieceEntityService())
```
LindvedKrvang, line 346: So this will always instantiate a "real" instance instead of a mock if no `PieceEntityService`?

src/app/core/services/models/piece-entity.service.ts
```diff
@@ -0,0 +1,25 @@
+import { Injectable } from '@angular/core'
+import { Piece } from '../../models/piece'
+
+@Injectable()
+export class PieceEntityService {
```
LindvedKrvang, line 5: All the methods in this class are good candidates for tests :)

src/app/core/services/models/segment-entity.service.spec.ts
```diff
@@ -391,6 +392,10 @@ describe(SegmentEntityService.name, () => {
...
 
 function createTestee(maybePartEntityService?: PartEntityService): SegmentEntityService {
-  const partService: PartEntityService = maybePartEntityService ?? new PartEntityService()
+  const partService: PartEntityService = maybePartEntityService ?? createPartEntityService()
   return new SegmentEntityService(partService)
 }
+
+function createPartEntityService(): PartEntityService {
+  return new PartEntityService(new PieceEntityService())
+}
```
LindvedKrvang, line 401: Is it on purpose that you create real instances instead of mocks?

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
```diff
@@ -89,6 +101,39 @@ export class ProducerShelfComponent implements OnInit, OnDestroy {
       this.shelfConfiguration = event.shelfConfiguration
       this.updateActionPanels()
     })
+
+    setInterval(() => this.updateActionPieceRelations(), CHECK_ON_AIR_PIECES_INTERVAL_IN_MS)
```
LindvedKrvang, line 105: We should probably ensure that we clear this interval again in the `OnDestroy` lifecycle hook

src/app/core/services/models/part-entity.service.spec.ts
KvelaGorrrrnio: Yes :)

src/app/core/services/models/segment-entity.service.spec.ts
KvelaGorrrrnio: It is on purpose :) I'd rather test the logic and outcome of a functionality, than testing the program structure.

src/app/core/services/models/piece-entity.service.ts
KvelaGorrrrnio: Good point! I will create tests for them in a separate PR.

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
KvelaGorrrrnio: Nice catch! I will add it in a separate PR :)

PR #295, opened 2024-12-11
Sof 2071/replace legacy angular material modules

LindvedKrvang:
In Angular v16 the version AngularMaterials that we were using was deprecated.
This PR removes a lot of the AngularMaterial dependencies since we simply don't need them.
We now only depend on the DialogModule of Angular Materiale.


src/app/shared/components/confirmation-dialog/confirmation-dialog.component.scss
```diff
@@ -1,58 +1,49 @@
...
-    margin-top: 12px;
-    margin-bottom: -12px;
-}
 
-.u-row {
-    display: flex;
-    justify-content: space-between;
-    align-items: center;
-}
+    .dialog-content {
```
MadsThy, line 39: i think something is wrong with the indentation in this file :)

src/app/shared/components/confirmation-dialog/confirmation-dialog.component.ts
```diff
@@ -65,4 +72,15 @@ export class ConfirmationDialogComponent extends StronglyTypedDialog<Confirmatio
...
+
+  public confirmDialog(): void {
+    this.dialogRef.close(true)
+  }
+
+  public cancelDialog(): void {
+    this.dialogRef.close(false)
+  }
+
+  protected readonly Icon = Icon
```
MadsThy, line 84: should these not be at the top of the file? :)

MadsThy (approved):

RasmusAlbrektsen (approved):

src/app/shared/components/confirmation-dialog/confirmation-dialog.component.scss
LindvedKrvang: Here is a picture of it. It looks good to me
image

src/app/shared/components/confirmation-dialog/confirmation-dialog.component.ts
LindvedKrvang: They should. It's Webstorm that autoplaced them down there.
I'll move them up in SOF-2364

PR #301, opened 2024-12-16
Sof 2372/able to display modifiers

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-web-client/pull/300 and https://github.com/tv2/sofie-tv2-server/pull/66

Fetches and displays all PanelInputModifiers.
Also listens for CREATED and DELETED events and reflect them.

Also in this PR:
- Moved some cardHeader specific styling away from the static buttons component and into the card component


src/locale/messages.da.xlf
```diff
@@ -642,6 +642,14 @@
         <source>panel-configuration.edit.tooltip</source>
         <target state="new">Rediger Panel</target>
       </trans-unit>
+      <trans-unit id="1144406098650646151" datatype="html">
+        <source>panel-input-modifiers.card-header</source>
+        <target state="new">Modifikators</target>
```
RasmusAlbrektsen: We should probably change this to`Modifikatorer` before all our danish users rebel against us.

RasmusAlbrektsen (approved):
Great work!

src/locale/messages.da.xlf
LindvedKrvang: We currently don't support a way to change the language from the UI, so they would never know :D :D :D 

It has been updated.

src/app/settings/panel-settings/components/panel-input-modifiers/panel-input-modifiers.component.ts
```diff
@@ -0,0 +1,26 @@
...
+    {
+      name: $localize`panel-input-modifier.column-title.name`,
+      sortCallback: (a: PanelInputModifier, b: PanelInputModifier): number => a.name.localeCompare(b.name),
+    },
+  ]
+
+  constructor(private readonly panelInputModifierStateService: PanelInputModifierStateService) {}
+
+  public ngOnInit(): void {
+    this.panelInputModifierStateService.subscribeToPanelInputModifiers(panelInputModifiers => (this.panelInputModifiers = panelInputModifiers))
```
MadsThy, line 24: Does this need to be run in OnInit? :) Otherwise I would suggest putting it in the constructor

src/app/settings/panel-settings/services/panel-input-modifier-state.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  private removePanelInputModifier(panelInputModifierId: string): void {
+    const indexToDelete: number = this.panelInputModifiers.findIndex(modifier => modifier.id === panelInputModifierId)
+    if (indexToDelete === -1) {
+      return
+    }
+    this.panelInputModifiers.splice(indexToDelete, 1)
+    this.updateSubject()
+  }
```
MadsThy, line 38: Just a suggestion :)

```suggestion
  private removePanelInputModifier(panelInputModifierId: string): void {
    this.panelInputModifiers = this.panelInputModifiers.filter(modifier => modifier.id !== panelInputModifierId)
    this.updateSubject()
  }
```

src/app/settings/shelf-settings/components/static-buttons-configuration-card/static-buttons-configuration-card.component.html
```diff
@@ -1,14 +1,12 @@
...
         <sofie-icon-button [icon]="Icon.PEN" [iconSize]="IconSize.M" [tooltip]="'Edit Static Buttons'" (click)="openEditStaticButtons()"></sofie-icon-button>
     </ng-container>
         <div class="static-actions-display">
-            <ng-container *ngIf="selectedStaticActions.length !== 0 else noStaticButtons">
-                <div class="action-container">
-                    <sofie-action-box *ngFor="let action of selectedStaticActions" [action]="action"></sofie-action-box>
-                </div>
-            </ng-container>
+            <div class="action-container" *ngIf="selectedStaticActions.length !== 0 else noStaticButtons">
+                <sofie-action-box *ngFor="let action of selectedStaticActions" [action]="action"></sofie-action-box>
```
MadsThy, line 8:
```suggestion
                <sofie-action-box *ngFor="let action of selectedStaticActions" [action]="action" />
```

Just a suggestion :) Let me quickly find the rule to prefer self-closing, if you want to implement that

MadsThy: I made a PR where the rule is in now :)

MadsThy (approved):

PR #302, opened 2024-12-16
Sof 2373/able to create modifier

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-web-client/pull/301

Able to create a PanelInputModifier.

There is now a "create" button that, when clicked, opens our Sidebar with a form for a PanelInputModifier. 
Filling out the form and clicking save will create a new Modifier.


src/app/settings/panel-settings/components/edit-panel-input-modifier/edit-panel-input-modifier.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public cancel(): void {
+    this.onCancel.emit()
+  }
+
+  private updateSaveButtonTooltip(): void {
+    const invalidControlsLabels: string[] = this.findInvalidControlsLabels()
+    if (invalidControlsLabels.length > 0) {
+      this.saveButtonTooltip = $localize`common.form-required-field.button-tooltip` + ` ${invalidControlsLabels.join(', ')}`
```
RasmusAlbrektsen, line 48: This is very minor and i should have caught it when it came to. But when we use the `common.form-required-field.button-tooltip` we end with a period and then write the fields after. 
image

Maybe we should end with a `:` or write something following the statement to say e.g. (using the image example):
`The fields are: Model, Hostname`

Again, a minor thing, but it just stood out to me.

RasmusAlbrektsen (approved):
Nice work :-)

src/app/settings/panel-settings/components/edit-panel-configuration/edit-panel-configuration.component.ts
```diff
@@ -54,6 +56,9 @@ export class EditPanelConfigurationComponent implements OnInit {
     if (this.panelConfiguration?.model) {
       this.updatePanelLayoutConfigurationOptions(this.panelConfiguration.model)
     }
+
+    this.updateSaveButtonTooltip()
+    this.form.valueChanges.subscribe(() => this.updateSaveButtonTooltip())
   }
 
   public save(): void {
```
MadsThy, line 64: I think you need a check in the save function to make sure it doesn't run if the button is disabled :)

```
  if(!form.valid){
    return
  }
```

src/app/settings/panel-settings/components/edit-panel-input-modifier/edit-panel-input-modifier.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      id: this.formBuilder.control(this.panelInputModifier?.id),
+      name: this.formBuilder.control(this.panelInputModifier?.name, Validators.required),
+    })
+
+    this.updateSaveButtonTooltip()
+    this.form.valueChanges.subscribe(() => this.updateSaveButtonTooltip())
+  }
+
+  public save(): void {
+    this.onSave.emit(this.form.value)
```
MadsThy, line 38:
```suggestion
    if(!form.valid) {
     return
    }
    this.onSave.emit(this.form.value)
```

src/app/settings/panel-settings/components/edit-panel-input-modifier/edit-panel-input-modifier.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  templateUrl: './edit-panel-input-modifier.component.html',
+})
+export class EditPanelInputModifierComponent implements OnInit {
+  @Input()
+  public panelInputModifier?: PanelInputModifier
+
+  @Output()
+  public onSave: EventEmitter<PanelInputModifier> = new EventEmitter()
+  @Output()
+  public onCancel: EventEmitter<void> = new EventEmitter()
```
MadsThy, line 16:
```suggestion
  @Input() public panelInputModifier?: PanelInputModifier

  @Output() public onSave: EventEmitter<PanelInputModifier> = new EventEmitter()
  @Output() public onCancel: EventEmitter<void> = new EventEmitter()
```

MadsThy (approved):

src/app/settings/panel-settings/components/edit-panel-configuration/edit-panel-configuration.component.ts
LindvedKrvang: The button is outside of the `save()` function scope. So the `save()` function should not know anything about a button.
This seems like a minor detail, but it is quite an important distinction (that appliess to all functions).

That being said: The `form` is not part of the button and the `form` is something the `save()` function does know about and it wouldn't be a bad idea for the `save()` function to validate that the `form` is valid before saving :)

src/app/settings/panel-settings/components/edit-panel-input-modifier/edit-panel-input-modifier.component.ts
LindvedKrvang: Good catch. It's been updated. Also updated some english values to danish as they were in the danish translation.

PR #303, opened 2024-12-16
Sof 2374/able to delete modifier

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-web-client/pull/302

Able to delete a Modifier.
If a modifier is being used by a PanelLayoutConfiguration, it is not possible to delete the modifier and the delete-icon is disabled.
Note: In order to achieve this a "shell" of the `InputConfigurations` on PanelLayoutConfiguration has been implemented. It should still be implemented prober later.

Also in this PR:
- SofieIconButton is now actually disabled when the attibute `disabled` is true.


src/app/shared/components/icon-button/icon-button.component.ts
```diff
@@ -22,6 +22,9 @@ export class IconButtonComponent implements OnInit, OnChanges {
   @HostBinding('attr.disabled')
   public disabled: boolean = false
 
+  @Input()
+  public disabledTooltip: string = ''
```
MadsThy, line 26:
```suggestion
  @Input() public disabledTooltip: string = ''
```

Just a suggestion :)

src/app/settings/panel-settings/components/panel-input-modifiers/panel-input-modifiers.component.ts
```diff
@@ -37,4 +49,19 @@ export class PanelInputModifiersComponent implements OnInit {
       this.panelInputModifierStateService.createPanelInputModifier(panelInputModifier)
     })
   }
+
+  public deletePanelInputModifier(panelInputModifier: PanelInputModifier): void {
+    this.dialogService.createConfirmDialog(
```
MadsThy, line 52:
```suggestion
if(isDeleteDisabled(panelInputModifier)) {
  return
}

    this.dialogService.createConfirmDialog(
```

Pseudo code to make it so that the button doesn't react if you press it :)

MadsThy (approved):

src/app/settings/panel-settings/services/panel-input-modifier-state.service.ts
```diff
@@ -4,24 +4,36 @@ import { BehaviorSubject } from 'rxjs'
...
+import { PanelLayoutConfigurationStateService } from './panel-layout-configuration-state.service'
+
+const MODIFIER_DELIMITER: string = ';'
 
 @Injectable()
 export class PanelInputModifierStateService {
   private panelInputModifiers: PanelInputModifier[] = []
   private readonly panelInputModifiersSubject: BehaviorSubject<PanelInputModifier[]> = new BehaviorSubject(this.panelInputModifiers)
 
+  private panelInputModifiersNotAbleToBeDeleted: PanelInputModifier[] = []
```
RasmusAlbrektsen: How about `nonDeletablePanelInputModifiers` or maybe `lockedPanelInputModifiers`?

RasmusAlbrektsen: Same obviously goes for the methods below if you like it :-)

RasmusAlbrektsen (approved):

src/app/settings/panel-settings/components/panel-input-modifiers/panel-input-modifiers.component.ts
LindvedKrvang: This method has nothing to do with the button and has no influence on the button.

Sidenote: The buttons already don't react if they are disabled.

src/app/settings/panel-settings/services/panel-input-modifier-state.service.ts
LindvedKrvang: I like how it becomes much more distungishable from `panelInputModifiers` :)
It's been updated.

PR #312, opened 2025-01-10
Sof 2383/create mkt1a visual representation

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-web-client/pull/311

This PR introduces the visual reprensentation of a MKT1A panel.
image

It works the same as for the MK48 panel. This panel also have a TBar input.


src/app/settings/panel-settings/components/visual-panel-implementations/panel-display-input/panel-display-input.component.scss
RasmusAlbrektsen, line 1: I would reduce the border to 0.1rem in the animation for the displays. It looks cleaner since the background is black anyway, and it will also stop the "jittering" that happens in the MKT1A when you select a display and then select a button (and reverse). 
image

RasmusAlbrektsen (approved):
Good stuff!

vushu (approved):
LGTM

src/app/settings/panel-settings/components/visual-panel-implementations/panel-display-input/panel-display-input.component.scss
LindvedKrvang: No argument here. Especially since it removes the "jumping around" :D 
It's been updated

PR #314, opened 2025-01-11
implemented touch support for dragging shelf and revised mouse implem…

vushu:
Added touch support for dragging shelf and revised the mouse dragging code.


src/app/shared/components/draggable-shelf/draggable-shelf.component.ts
```diff
@@ -14,13 +17,10 @@ interface PositionData {
...
   public isCollapsed: boolean = false
   public heightInPixels: number = 0
-  public verticalDragPositionFromDragHandleBottomInPixels?: number
 
   protected readonly Icon = Icon
   protected readonly IconSize = IconSize
-
-  @ViewChild('dragHandle')
-  private readonly dragHandleElement: ElementRef<HTMLDivElement>
+  private _currentState: DragState = DragState.END_DRAGGING
```
LindvedKrvang: We like to be explicit in our naming of variables, methods etc. Ideally, it should be possible to look at a name and then know what the variable is for. 

`state` is a very broad term and can be used to reference a lot of different things. So the name `currentState` leaves a broad interpretation. What's stopping me from storing the current height of the shelf in the `currentState` variable?

The quickfix here would be to rename the variable to `currentDragState`. But looking at your implementation, it looks like you are simply interested in knowing whether you are dragging or not. So you could suffice with something like: `private isDraggingShelf: boolen = false`.

Minor thing: We don't prefix our variables with underscores.

src/app/shared/components/draggable-shelf/draggable-shelf.component.ts
```diff
@@ -55,24 +55,40 @@ export class DraggableShelfComponent implements OnInit {
...
-    window.addEventListener(
-      'mouseup',
-      () => {
-        this.verticalDragPositionFromDragHandleBottomInPixels = undefined
-        window.removeEventListener('mousemove', onDragMove)
-        this.storePositionData()
-      },
-      { once: true }
-    )
+  }
```
LindvedKrvang: `onMouseDragMove()` and `onTouchDragMove()` has the exact same implementation.
To help future maintenance developers (and reduce the risk of only changing one of the two places when a change is needed) you should extract the implementation into a seperate function:
```
public onMouseDragMove(event: MouseEvent): void {
  this.dragShelf(event.clientY)
}

private dragShelf(positionToDragTo: number): void {
  if (this._currentState !== DragState.START_DRAGGING) {
    return
  }
  this.heightInPixels = Math.max(0, window.innerHeight - positionToDragTo)
}

public onTouchDragMove(event: TouchEvent): void {
  this.dragShelf(event.touches[0].clientY)
}
```

src/app/shared/components/draggable-shelf/draggable-shelf.component.ts
```diff
@@ -1,11 +1,14 @@
...
 
 interface PositionData {
   heightInPixels: number
   isCollapsed: boolean
 }
-
+enum DragState {
+  START_DRAGGING,
+  END_DRAGGING,
+}
```
LindvedKrvang: There is something in the naming here that is bugs me a little bit.

I think it's because the name of the Enum has the word `state` in it, while the entries of the Enum is named as verbs instead of nouns.

The state is either "being dragged" or "not being dragged". If the state is "start dragging", then we are continuously starting to drag, but the start of the drag only happens once. (Same for "end dragging".)

As mentioned in another comment: It looks like all you need here is a boolean to tell you whether you are dragging or not :)

LindvedKrvang (approved):
Not bad :) 

The comments are of the minor nature. I would however, still like for us to have a brief talk about them.

RasmusAlbrektsen (commented):
I agree with the comments made by @LindvedKrvang, and would like to see those fixed before a merge. However, this looks very good! 

I'm refraining from an approve at this moment since i think we should keep this PR up and use it to make a build, try out the deploy process and then test it in the staging environment. :-)

RasmusAlbrektsen (approved):

PR #317, opened 2025-01-17
SOF-2386  Configure Action input

LindvedKrvang:
Able to configure a Button input with a Color and an Action.

In this PR:
- Able to select a color for a regular Button input.
- Able to select an Action for a regular Button input.
- Able to select ActionArguments for an Action requiring arguments.
- Filtrering away InputConfiguration that does not have a "command" when saving, so the backend doesn't throw a 400.
- Introduce types for SYSTEM_ACTIONS being provided from the backend.
- Minor tweaks to the SofieNumberInput to reflect changes as they are updated and not just on focus lost.
- The save button tooltip now displays a more descriptive message when there are invalid inputs.
- Added translations


vushu (approved):
Looks fine to me.

src/app/settings/panel-settings/services/panel-form.service.ts
```diff
@@ -1,13 +1,23 @@
...
 import { PanelLayoutConfiguration } from '../models/panel-layout-configuration'
-import { PanelCommand, PanelCommandType, PanelInputConfiguration, PanelInputType } from '../models/panel-input-configuration'
+import { KeyEvent, PanelCommand, PanelCommandType, PanelInputConfiguration, PanelInputType } from '../models/panel-input-configuration'
+
+interface AbstractControlWithId {
+  id: string
+  control: AbstractControl
+}
+
+const ERROR_MESSAGE_SEPARATOR: string = '; '
```
RasmusAlbrektsen, line 11: Just making sure, is it on purpose that there's a space after the `;` here?

RasmusAlbrektsen (approved):

src/app/settings/panel-settings/services/panel-form.service.ts
LindvedKrvang: It is :)

PR #319, opened 2025-01-22
SOF-2287 Able to configure the text of a Display input

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-web-client/pull/317

When clicking a Display input it is now possible to configure the text of the Display.


vushu (approved):

RasmusAlbrektsen (approved):
Nice! 
Only thing that comes to mind is that i think it would make sense if inputs that only have one possible function, directly open up the possibilities for editing it. E.g. clickling a display would instantly show the text field, rather than have me press Display first and then seeing it.

LindvedKrvang:
> Nice! Only thing that comes to mind is that i think it would make sense if inputs that only have one possible function, directly open up the possibilities for editing it. E.g. clickling a display would instantly show the text field, rather than have me press Display first and then seeing it.

I agree, that would be nice. However, I currently use the click event when clicking that option, to add the formGroup to the form. (And the "go back" to remove the formGroup again".) If we go straight to the option, we would either need something else to manage the added/removal of the formGroups or, update all our validating logic to disregard "unconfigured" formGroups (and we won't be able to simply do `form.value` anymore to extract the configuration)

PR #323, opened 2025-01-24
Fixed shelf stuck problem and simplified further

vushu:
- Fixed the issue with the shelf getting stuck to the mouse,

-  Also fixed another issue where collapsing didn't work.

- I have added a timeout to diffentiate between draggging and double clicking which is used for collapsing the shelf.

I discovered that pointerup, pointermove, pointerdown, also supports touch events, which enabled me to remove the previous touch event logic.


LindvedKrvang (approved):

RasmusAlbrektsen (commented):
I'm not able to make this work when i emulate and use touch events through Chrome, does it work for you? 

Edit: I had messed up my settings, i got it to work once i figured it out :-D Sorry for the confusion!

RasmusAlbrektsen (approved):

PR #326, opened 2025-01-29
Sof 2391/able to add modified behaviour to action

LindvedKrvang:
It is now possible to configure a ButtonInput, ButtonWithDisplayInput and DisplayButton with modified behaviour.
Clicking the newly added button "Add modifier" will reveal appropiate inputs that can be used to configure modified behaviour for the selected PanelInput.
image

It's possible to add as many modified behaviours as one see fits.
Clearing the input configuration will also clear all modified behaviours.
If a modified behaviour has been added, but not configured correctly, then it is not possible to save the PanelLayoutConfiguration.


src/app/settings/panel-settings/services/panel-form.service.ts
```diff
@@ -187,13 +198,84 @@ export class PanelFormService {
...
+        type: PanelCommandType.DISPLAY,
+      },
+      text: '',
+    }
+  }
+
+  public updateModifierFormIdForInputConfiguration(modifierForm: FormGroup, selectedModifierIds: string[], inputConfigurationForm: FormGroup): void {
+    const id: string | undefined = inputConfigurationForm.get(ID_CONTROL_NAME)?.value
+    if (!id) {
+      this.logger.data(inputConfigurationForm).warn("Can't update modifier id. No id found on panelInputConfigurationForm")
```
RasmusAlbrektsen, line 247: I hear faint whispering in the distance, i think it's Anders. He's asking us to put `.` after our log statements.

src/app/settings/panel-settings/services/panel-form.service.ts
```diff
@@ -53,8 +63,9 @@ export class PanelFormService {
   }
 
   private buildInputConfigurationFormGroup(id: string, inputConfiguration: Omit<PanelInputConfiguration, 'id'>): FormGroup {
+    const beginningDigitRegex: RegExp = /^\d+/
```
RasmusAlbrektsen: Not significant, but it just sounds weird to me for some reason, how about `initialDigitRegex` or `startingDigitRegex`.

src/locale/messages.da.xlf
```diff
@@ -826,6 +826,26 @@
...
+        <source>panel-input-configuration-section.select-modifier.placeholder</source>
+        <target state="new">Vælg modifikatorer</target>
+      </trans-unit>
+      <trans-unit id="4085714197007370764" datatype="html">
+        <source><x id="PH" equiv-text="inputConfigurationName"/>panel-input-configuration.invalid.modifier-id</source>
+        <target state="new"><x id="PH" equiv-text="inputConfigurationName"/>: Mangler modifikatorer</target>
+      </trans-unit>
+      <trans-unit id="5915583677884164476" datatype="html">
+        <source>panel-input-configuration.modifier-prefix<x id="PH" equiv-text="type"/> <x id="PH_1" equiv-text="id.split(MODIFIER_ID_DELIMITER)[1]"/></source>
+        <target state="new">Modifikatorer: <x id="PH" equiv-text="type"/> <x id="PH_1" equiv-text="id.split(MODIFIER_ID_DELIMITER)[1]"/></target>
```
RasmusAlbrektsen: This one has `:` after it, but the english version doesn't. Which is the correct version?

RasmusAlbrektsen (approved):
Very good sir!

vushu (approved):

src/app/settings/panel-settings/services/panel-form.service.ts
LindvedKrvang: You shouldn't be listening to whispers from the dark side ;)

src/app/settings/panel-settings/services/panel-form.service.ts
LindvedKrvang: I have renamed it to `firstCharIsDigitRegex`.
I would like it to emphasize that it is checking whether the first character is a digit.

src/locale/messages.da.xlf
LindvedKrvang: I think it is the one without the `:`. 
It's been updated

RasmusAlbrektsen (approved):

PR #334, opened 2025-02-06
Will insert part wrt it's rank.

vushu:


src/app/core/services/models/segment-entity.service.ts
```diff
@@ -88,14 +88,17 @@ export class SegmentEntityService {
     return parts.filter(part => !part.isUnsynced)
   }
 
+  private insertPartSortedByRank(parts: Part[], newPart: Part): Part[] {
+    return [...parts, newPart].sort((a: Part, b: Part) => a.rank - b.rank)
```
LindvedKrvang: Is sorting the entire array necessary? It seems like a potential computation heavy operation, when the array should already be sorted.

src/app/core/services/models/segment-entity.service.ts
```diff
@@ -88,14 +88,17 @@ export class SegmentEntityService {
     return parts.filter(part => !part.isUnsynced)
   }
 
+  private insertPartSortedByRank(parts: Part[], newPart: Part): Part[] {
```
LindvedKrvang: It would be very nice if you would place the order of methods in the order they are used.
Then when you read the file from top to bottom you get a nice consistent flow.
It would look like this:
```
methodA(): void {
  methodB()
}

methodB(): void {
  methodD()
}

methodD(): void {
  // Something
}

methodC(): void {
  // Something else
}
```

LindvedKrvang (approved):

src/app/core/services/models/segment-entity.service.ts
vushu: yes in worst case the time complexity will be O(nˆ2).

Another algorithm we could use to
run through all parts if n.rank < part.rank  and  part.rank > (n+1).rank, we insert the part in between.
this may be a approach since it's already sorted, I will proably take a look at it.

LindvedKrvang: The sort might be fine. (Atleast it is very simple to understand, which is a big plus).

We just need to find the index. But if finding the index makes the code way more complex then I vote for the simple sort (and if we later run into a performance issue, than we can update it)

vushu: agreed, lets keep it simple.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #335, opened 2025-02-10
Sof 2394/able to change layers in visual representation

LindvedKrvang:
When editing a PanelLayoutConfiguration it is now possible to select which Modifier layers that should be displayed in the visual representation.
Selecting a modifier (or multiple) will update the representation to show what has been configured for the chosen modifier(s).

An Input configured to be a "MODIFIER" is always shown regardless of what modifers have been selected.


vushu (approved):
Looks good.

RasmusAlbrektsen (approved):

PR #337, opened 2025-02-13
SOF-2478 - System Actions Visible on Keyboard

RasmusAlbrektsen:
The actions we use when comparing the shortcut actions to available actions did not include the system actions. I included these by gathering them from the state service and including them into the available actions.


LindvedKrvang (approved):

vushu (approved):
image

it works like charm :)

PR #338, opened 2025-02-13
Added toggle recall mode in rundown

vushu:
This is how it looks for now.

image

The recall icon has a tooltip and animates while in recall mode.


src/app/shared/components/mode-activation-bar/mode-activation-bar.component.html
```diff
@@ -0,0 +1 @@
+<div [class]="takeMode.toLowerCase()"><span class="bar-text">{{ getTakeModeText() }}</span></div>
```
LindvedKrvang: Instead of using CSS to hide the bar you could use the `ngIf` directive:
```
// html
<div *ngIf="shouldShowTakeModeBar()">...</div>

// ts
public shouldShowTakeModeBar(): boolean {
  return this.rundown.takeMode === TakeMode.RECALL
}
```

This way you only need one CSS class. (But you would have to animate the bar through "Angulars" animations: https://angular.dev/guide/animations/transition-and-triggers#animate-entering-and-leaving-a-view)

src/app/shared/components/mode-activation-bar/mode-activation-bar.component.ts
```diff
@@ -0,0 +1,19 @@
...
+import { RundownTakeMode } from '../../../core/enums/rundown-take-mode'
+
+@Component({
+  selector: 'sofie-mode-activation-bar',
+  templateUrl: 'mode-activation-bar.component.html',
+  styleUrls: ['./mode-activation-bar.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ModeActivationBarComponent {
+  @Input() public takeMode: RundownTakeMode
```
LindvedKrvang: I would just parse the entire Rundown as an input. 
We are working with objects, so it's a good idea to just pass the objects along. It gives more flexibility of what you want to do/can achieve :)

src/app/shared/components/mode-activation-bar/mode-activation-bar.component.ts
```diff
@@ -0,0 +1,19 @@
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core'
+import { RundownTakeMode } from '../../../core/enums/rundown-take-mode'
+
+@Component({
+  selector: 'sofie-mode-activation-bar',
+  templateUrl: 'mode-activation-bar.component.html',
+  styleUrls: ['./mode-activation-bar.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ModeActivationBarComponent {
```
LindvedKrvang: We already have something called `RundownMode` which is different from `TakeMode`.
Unless this component is supposed to support that as well (which it currently isn't), I would suggest to rename it to `TakeModeActivationBarComponent`. 
 
`Activation` doesn't sit right with me. `Activation` describes the act of activating somthing. The way `TakeMode` has been implemented, we are technically not activating anything. We are changing a state. What happens when we add a third entry to the `TakeMode` enum? What does activation then mean?
I suggest calling this component: `TakeModeStatusBarComponent`.

src/app/shared/components/mode-activation-bar/mode-activation-bar.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  @Input() public takeMode: RundownTakeMode
+
+  public getTakeModeText(): string {
+    if (this.takeMode === RundownTakeMode.RECALL) {
+      return $localize`mode-activation-bar.recall-activated`
+    }
+    return ''
```
LindvedKrvang: If you want, you can use a ternary here :)
```
return this.takeMode === RundownTakeMode.RECALL
  ? $localize`mode-activation-bar.recall-activated`
  : ''
```

src/app/shared/components/mode-activation-bar/mode-activation-bar.component.ts
LindvedKrvang: Sidenote: If you do choose to go away from `activation`, then you need to remember to rename all the places that references "activation".

src/app/core/enums/rundown-take-mode.ts
```diff
@@ -0,0 +1,4 @@
+export enum RundownTakeMode {
```
LindvedKrvang: What's the reason for calling it `RundownTakeMode` and not just `TakeMode`?

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.html
```diff
@@ -0,0 +1,14 @@
+<div sofieTooltip
+  [sofieTooltipTemplate]="rundown.takeMode === RundownTakeMode.STANDARD ? activateTooltip : deactivateTooltip"
+  [class]="rundown.takeMode.toLocaleLowerCase()"
```
LindvedKrvang: I'm not a fan of how tightly coupled this code is to the CSS class name through the use of `toLowerCase`.
If we refactor `RECALL` to `SWAP` then we break the styling because the CSS name is not changed and the compiler can't tell us that this is broken.

You have done it somewhere else where I would like to see it changed as well.

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.scss
```diff
@@ -0,0 +1,40 @@
+:host {
+  div {
+    --recall-icon-code-color: var(--dark-gray-color);
+    display: flex;
+    padding: 10px;
```
LindvedKrvang: Please don't specify paddings in the CSS (or margins).
We have helper CSS classes for this. `p-1` sets the padding allround the element to what we have defined as "1". `pt-3` sets the top padding of the element to what we have defined as "3" etc. Same for margin.

The reason for this is to try and streamline the look of the application with consistent and identical margins/paddings. It also means that we can change how much "white-space" we have from a single location.

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.scss
```diff
@@ -0,0 +1,40 @@
...
+    height: 100%;
+    align-items: center;
+    cursor: pointer;
+    font-size: 1.2rem;
+
+    sofie-icon-button {
+      color: var(--recall-icon-code-color);
+      transition: color .1s;
+      z-index: 1;
```
LindvedKrvang: z-index is an attribute that we should ideally not try and set.
Unless you have a good reason to set it, we should try and avoid it :)

The reason is, that it is very easy to get into weird UI where all the elements have z-indexes all over the place.

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  @Input() public rundown: Rundown
+
+  constructor(private readonly rundownService: RundownService) {}
+
+  private getToggledTakeMode(): RundownTakeMode {
+    return this.rundown.takeMode === RundownTakeMode.STANDARD ? RundownTakeMode.RECALL : RundownTakeMode.STANDARD
+  }
+
+  public toggleRecallMode(): void {
```
LindvedKrvang: `toggleRecallMode()` is calling `getToggleTakeMode()` so `toggleRecallMode()` should be above `getToggledTakeMode()` to make the file "read like a book".

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.rundownService.setTakeMode(this.rundown.id, this.getToggledTakeMode()).subscribe()
+  }
+
+  protected readonly RundownTakeMode = RundownTakeMode
```
LindvedKrvang: It would be nice if this was moved up next to `Icon = Icon` etc above

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.ts
```diff
@@ -0,0 +1,29 @@
...
+import { RundownTakeMode } from '../../../core/enums/rundown-take-mode'
+import { Rundown } from '../../../core/models/rundown'
+import { RundownService } from '../../../core/abstractions/rundown.service'
+
+@Component({
+  selector: 'sofie-toggle-recall-mode-button',
+  templateUrl: './toggle-recall-mode-button.component.html',
+  styleUrls: ['./toggle-recall-mode-button.component.scss'],
+})
+export class ToggleRecallModeButtonComponent {
```
LindvedKrvang: You have put this component in the `shared` module.
As of right now, this component is only being used in the `rundown-view` module, so it should be placed in that module as well.

The same goes for `mode-activation-bar-component`.

src/app/shared/components/mode-activation-bar/mode-activation-bar.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  templateUrl: 'mode-activation-bar.component.html',
+  styleUrls: ['./mode-activation-bar.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ModeActivationBarComponent {
+  @Input() public takeMode: RundownTakeMode
+
+  public getTakeModeText(): string {
+    if (this.takeMode === RundownTakeMode.RECALL) {
+      return $localize`mode-activation-bar.recall-activated`
```
LindvedKrvang: Minor thing, but it would be nice if you would append `message` to take translation key.
We would like to try and make our keys precise. Imagine you had both a message and a title for this. How would you know which key to use for which place?

src/locale/messages.da.xlf
```diff
@@ -854,6 +854,18 @@
         <source>edit-panel-layout-configuration.modifier-select.placeholder</source>
         <target state="new">Ingen modifikatorer valgt</target>
       </trans-unit>
+      <trans-unit id="3388044901828416266" datatype="html">
+        <source>mode-activation-bar.recall-activated</source>
+        <target state="new">Kildegentagelsestilstand aktiveret. Vil kontinuerligt ombytte PGM og PRV ved Take</target>
```
LindvedKrvang: I'm impressed that you came up with `Kildegentagelsestilstand` :D

LindvedKrvang (commented):
Good work :)

Most of the comments are of the minor nature with only a few comments making it so you don't have my approve quite yet. These comments are mainly that you have placed the components in the wrong module.

src/app/shared/components/mode-activation-bar/mode-activation-bar.component.html
vushu: Ahh I see if it's ok with you Im going to revisit this in the future and make a task for that.

src/app/core/enums/rundown-take-mode.ts
vushu: not really i guess :p

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.html
vushu: Actually i got the inspiration from notification-icon.component.html :d

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.ts
vushu: no clue why it went down there.

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.ts
vushu: I guess Im used to the opposite :p

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.html
LindvedKrvang: Well, eggs in my face then :D

src/app/shared/components/toggle-recall-mode-button/toggle-recall-mode-button.component.ts
LindvedKrvang: When WebStorm auto inserts them (when you choose the auto import while in a html file) it inserts them at the bottom of the ts file.

LindvedKrvang (approved):
Nice work :)

RasmusAlbrektsen (approved):

PR #339, opened 2025-02-18
Added disable panel configuration

vushu:
Added disable for panel configuration, this PR requires this PR [https://github.com/tv2/sofie-tv2-server/pull/93#issuecomment-2663341634](https://github.com/tv2/sofie-tv2-server/pull/93#issuecomment-2663341634)


src/app/settings/panel-settings/components/edit-panel-configuration/edit-panel-configuration.component.html
```diff
@@ -3,6 +3,7 @@
     <sofie-select i18n-label label="edit-panel-configuration.model.label" [formControlName]="'model'" [options]="modelOptions" [isRequired]="true" />
     <sofie-text-input i18n-label label="edit-panel-configuration.hostname.label" [formControlName]="'hostname'" [isRequired]="true" />
     <sofie-select i18n-label label="edit-panel-configuration.panel-layout-configuration.label" [formControlName]="'panelLayoutConfigurationId'" [options]="panelLayoutConfigurationOptions" />
+    <sofie-select i18n-label label="edit-panel-configuration.is-disabled.label" [formControlName]="'isDisabled'" [isRequired]="true" [options]="panelYesNoOptions" />
```
LindvedKrvang: The `isDisabled` is a yes/no question. That means that this should just be a checkbox which mean you should simply use `<sofie-checkbox />` for this.

However, I can see that our SofieCheckbox is missing to be turned into a ControlValueAccessor. So we would need to do that first.

src/app/settings/panel-settings/components/edit-panel-configuration/edit-panel-configuration.component.ts
```diff
@@ -30,6 +30,13 @@ export class EditPanelConfigurationComponent implements OnInit {
 
   public modelOptions: SelectOption<PanelModel>[] = []
   public panelLayoutConfigurationOptions: SelectOption<string>[] = []
+  public panelYesNoOptions: SelectOption<boolean>[] = [
+    {
+      name: $localize`global.yes`,
+      value: true,
+    },
+    { name: $localize`global.no`, value: false },
+  ]
```
LindvedKrvang: This won't be needed when we are using a checkbox

LindvedKrvang (changes requested):
The request changes is because you have used a dropdown instead of a checkbox for a true/false value selection. Refer to my comments

LindvedKrvang:
I have made a PR that makes the Checkbox capable of being used with Forms, so that you can use it for this PR :)
https://github.com/tv2/sofie-web-client/pull/342

vushu:
Added checkbox

image

RasmusAlbrektsen (approved):

src/app/settings/panel-settings/components/panel-configurations/panel-configurations.component.html
```diff
@@ -10,6 +10,7 @@
             <td class="td-width">{{panelConfiguration.type}} - {{panelConfiguration.model}}</td>
             <td class="td-width">{{panelConfiguration.hostname}}</td>
             <td>{{getPanelLayoutConfigurationNameForPanelConfiguration(panelConfiguration)}}</td>
+            <td class="td-width">{{getLocalizedIsDisabledText(panelConfiguration.isDisabled)}}</td>
```
RasmusAlbrektsen, line 13: I'm not too fond of having a column just for this, but since we don't exactly have the ability to simply show a "Status" where this could also be used, then i guess we just go with this for now :-)

RasmusAlbrektsen (approved):

src/app/settings/panel-settings/components/edit-panel-configuration/edit-panel-configuration.component.html
```diff
@@ -1,8 +1,11 @@
-<form [formGroup]="form" class="mb-3">
+<form [formGroup]="form" xmlns="http://www.w3.org/1999/html" class="mb-3">
     <sofie-select i18n-label label="edit-panel-configuration.type.label" [formControlName]="'type'" [options]="typeOptions" [isRequired]="true" />
     <sofie-select i18n-label label="edit-panel-configuration.model.label" [formControlName]="'model'" [options]="modelOptions" [isRequired]="true" />
     <sofie-text-input i18n-label label="edit-panel-configuration.hostname.label" [formControlName]="'hostname'" [isRequired]="true" />
     <sofie-select i18n-label label="edit-panel-configuration.panel-layout-configuration.label" [formControlName]="'panelLayoutConfigurationId'" [options]="panelLayoutConfigurationOptions" />
+    <sofie-checkbox [formControlName]="'isDisabled'" />
+    <label class="ps-2" i18n>edit-panel-configuration.is-disabled.label</label>
```
LindvedKrvang, line 7: It would be very nice if the label was before the checkbox.
It gives a more natural flow of: first you read what you are selecting; then your select it.

LindvedKrvang (approved):

src/app/settings/panel-settings/components/edit-panel-configuration/edit-panel-configuration.component.html
vushu: I'll keep it in mind, we can change it at some point.

PR #341, opened 2025-02-19
feat: Recall mode now designed as a panel item that can be toggled

RasmusAlbrektsen:
This PR is mainly visual changes to the Recall mode, and can be seen in the image below.
Any feedback here is welcome :-) 

UPDATED LOOK:
image


vushu (approved):
Look very nice :D

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.ts
```diff
@@ -118,4 +120,8 @@ export class OnAirDetailsPanelComponent implements OnChanges, OnInit, OnDestroy
     const pieceMetadata: Tv2PieceMetadata | undefined = piece.metadata as Tv2PieceMetadata | undefined
     return pieceMetadata?.type === Tv2PieceType.MANUS
   }
+
+  protected readonly TakeMode = TakeMode
+  protected readonly Icon = Icon
+  protected readonly IconSize = IconSize
```
LindvedKrvang: Would be nice if these were placed at the top of the file :)

src/app/rundown-view/components/take-mode-status-bar/take-mode-status-bar.component.scss
```diff
@@ -1,21 +1,45 @@
 .recall {
-  height: 1.5em;
-  background-color: var(--attention-color);
-  transition: height 0.2s;
+  height: 2em;
+  background-color: var(--recall-mode-color);
+  padding: 4px;
```
LindvedKrvang: You should ideally use our padding classes instead of setting the values as CSS

src/app/rundown-view/components/toggle-recall-mode-button/toggle-recall-mode-button.component.html
```diff
@@ -2,7 +2,7 @@
   [sofieTooltipTemplate]="rundown.takeMode === RundownTakeMode.STANDARD ? activateTooltip : deactivateTooltip"
   [class]="getTakeModeClass()"
   (click)="toggleRecallMode()" >
-  <sofie-icon-button [icon]="Icon.ROTATE" [iconSize]="IconSize.XL" />
+  <mat-slide-toggle color="primary" [checked]="rundown.takeMode === RundownTakeMode.RECALL"/>
```
LindvedKrvang: We are trying to avoid using `mat` components directly.
You should wrap this in a `sofie-slide-toggle`. Then we only have one place to update if/when we no longer use Material Components.

src/scss/_sofie-color-vars.scss
```diff
@@ -13,11 +13,13 @@ body {
   --green-color: green;
   --yellow-color: #ffea00;
   --blue-color: #0f5be9;
+  --purple-color: #5e35b1;
   --attention-color: #ffba00;
   --info-color: var(--blue-color);
   --orangered-color: orangered;
   --transparent-color: #00000000;
 
+  --recall-mode-color: var(--purple-color);
```
LindvedKrvang, line 22: `--recall-mode-color` is referencing how the color is used, not what the color "do" (bad word for this scenario I know).
This form of naming is how we got `--rundown-header-rehearsal-scrollbar-thumb-active-color`, which is not reuseable at all...

We should try and name our colors along the lines of `primary`, `secondary`, `tertiary`, `accent`, etc. This way we get a color scheme/palette that we can use

LindvedKrvang (commented):
It looks good. 
For the design, I have a pet peeve or two, but we can't please everyone ;) 

I haven't approved yet because of your direct usage of a Material component.
Not long ago we removed all of them but the Dialog. We shouldn't start reintroducing them again.

src/scss/_sofie-color-vars.scss
RasmusAlbrektsen: That is why i created the variable for the color itself and then made a new variable that can be used for whenever something has to have change color when going in to "recall" mode. In that way we only change the color one place if we want a different color for the mode entirely. 

But if you still don't see the point i'll get rid of it.

src/app/rundown-view/components/toggle-recall-mode-button/toggle-recall-mode-button.component.html
RasmusAlbrektsen: I'll make a new component for the slide toggle :-)

src/app/rundown-view/components/take-mode-status-bar/take-mode-status-bar.component.scss
RasmusAlbrektsen: Agreed, this one, however, is set directly from the .ts-file so i just wanted to avoid doing that in there. Unless you disagree?

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.ts
RasmusAlbrektsen: Moved!

src/app/rundown-view/components/toggle-recall-mode-button/toggle-recall-mode-button.component.html
RasmusAlbrektsen: It has been made!

src/app/rundown-view/components/take-mode-status-bar/take-mode-status-bar.component.scss
LindvedKrvang: Agree with that it is better to do in the CSS file than in the ts file

LindvedKrvang (approved):

PR #346, opened 2025-02-25
Sof 2487/added macros settings listing macros

vushu:
Added macros to settings, only listing the macros in this PR.


src/app/core/services/macro-event-observer.service.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    logger?: Logger
+  } = {}
+): MacroEventObserver {
+  const mockedEventObserver: EventObserver = parameters.eventObserver ?? instance(mock<EventObserver>())
+  const mockedRundownEventParser: MacroEventParser = parameters.macroEventParser ?? instance(createMockOfMacroEventParser())
+  const mockedLogger: Logger = parameters.logger ?? createLogger()
+  return new MacroEventObserver(mockedEventObserver, mockedRundownEventParser, mockedLogger)
+}
+
+function createMockOfMacroEventParser(): MacroEventParser {
```
LindvedKrvang: Minor thing: You can leave out the `of` so it just becomes `createMockMacroEventParser`

src/app/settings/components/settings-menu/settings-menu.component.ts
```diff
@@ -41,6 +41,10 @@ export class SettingsMenuComponent implements OnInit, OnDestroy {
       name: $localize`settings.sofie-settings.label`,
       routerLink: [SettingsPath.SOFIE_SETTINGS],
     },
+    {
+      name: $localize`settings.macros.label`,
+      routerLink: [SettingsPath.MACROS],
+    },
```
LindvedKrvang, line 43: It would be nice if this was above the "Sofie Settings".
The "Sofie Settings" are the settings from Core and we would like to just keep them at the bottom :)

src/app/settings/macro-settings/components/macro-table/macro-table.component.ts
```diff
@@ -0,0 +1,39 @@
...
+import { SofieTableHeader } from '../../../../shared/components/table/table.component'
+
+@Component({
+  selector: 'sofie-macro-table',
+  templateUrl: './macro-table.component.html',
+  styleUrls: ['./macro-table.component.scss'],
+})
+export class MacroTableComponent implements OnInit {
+  protected readonly Icon = Icon
+  protected readonly IconSize = IconSize
```
LindvedKrvang: These aren't being used (yet) so they should technically not be here.
But since they are going to be used quite soon, I don't think there is a needed to remove them

src/locale/messages.da.xlf
```diff
@@ -874,6 +874,22 @@
         <source>edit-panel-configuration.is-disabled.label</source>
         <target state="new">Slå fra</target>
       </trans-unit>
+      <trans-unit id="2851935488267626084" datatype="html">
+        <source>settings.macros.label</source>
+        <target state="new">Makroer</target>
+      </trans-unit>
+      <trans-unit id="3784074223911232308" datatype="html">
+        <source>settings.macro.name</source>
+        <target state="new">Makronavn</target>
```
LindvedKrvang: This should just be `Navn` since we are in the context of a Macro already where we display it.

src/locale/messages.en-US.xlf
```diff
@@ -874,6 +874,22 @@
         <source>edit-panel-configuration.is-disabled.label</source>
         <target state="new">Disable</target>
       </trans-unit>
+      <trans-unit id="2851935488267626084" datatype="html">
+        <source>settings.macros.label</source>
+        <target state="new">Macros</target>
+      </trans-unit>
+      <trans-unit id="3784074223911232308" datatype="html">
+        <source>settings.macro.name</source>
+        <target state="new">Macro Name</target>
```
LindvedKrvang: Same comment as for the danish translation

LindvedKrvang (approved):
Great work :)

src/app/settings/macro-settings/components/macro-table/macro-table.component.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Subject, takeUntil } from 'rxjs'
+import { MacroStateService } from '../../../../core/services/macro-state.service'
+import { SofieTableHeader } from '../../../../shared/components/table/table.component'
+
+@Component({
+  selector: 'sofie-macro-table',
+  templateUrl: './macro-table.component.html',
+  styleUrls: ['./macro-table.component.scss'],
+})
+export class MacroTableComponent implements OnInit {
```
LindvedKrvang: Small note on the naming of this component.
You have named it Macro _Table_ Component.
The "table" refers to an implementation detail. If someone decides to refactor it to no longer be a table, then we need to rename the component and update _all_ references to it.

If we instead named it something less implementation specific, then a potential refactor would affect this single file.

The component displays Macros, so maybe the component could simple be named something like `Macros`.

src/app/settings/macro-settings/components/macro-table/macro-table.component.ts
vushu: Ah yeah

src/app/settings/macro-settings/components/macro-table/macro-table.component.ts
vushu: Good point!

src/app/settings/macro-settings/components/macro-table/macros.component.ts
```diff
@@ -0,0 +1,39 @@
+import { Component, OnInit } from '@angular/core'
+import { Icon, IconSize } from '../../../../shared/enums/icon'
+import { Macro } from '../../../../shared/models/macro'
+import { Subject, takeUntil } from 'rxjs'
+import { MacroStateService } from '../../../../core/services/macro-state.service'
+import { SofieTableHeader } from '../../../../shared/components/table/table.component'
+
+@Component({
+  selector: 'sofie-macros',
```
RasmusAlbrektsen: Shouldn't this component folder be changed to fit the new name?

RasmusAlbrektsen (approved):

src/app/settings/macro-settings/components/macro-table/macros.component.ts
vushu: ah forgot will change it right away :d

PR #347, opened 2025-02-26
Redirect to StandaloneShelf when not on a desktop device

LindvedKrvang:
Proof of concept - not meant to be merged!

This is a proof of concept of how we can "force" all non-desktop devices to a "Standalone Shelf" while also still being able to navigate there from a desktop if we so desire.

The PoC utilises the library [ngx-device-detector](https://www.npmjs.com/package/ngx-device-detector) to figure out which device we are and [Angular Guards](https://v16.angular.io/api/router/CanActivateFn) to redirect non-desktop users.


vushu:
Cool I just tried it on my iphone :)

LindvedKrvang:
Closing

PR #349, opened 2025-03-04
SOF-2490 - Able to delete macros

RasmusAlbrektsen:
Note that this PR also includes changes to the size of existing icon buttons for the settings. It gave a nicer look and feel that is more in line with the size of e.g. the rows of the tables.


src/app/settings/macro-settings/components/macros/macros.component.ts
```diff
@@ -32,6 +33,17 @@ export class MacrosComponent implements OnInit {
     this.subscribeForMacroObservable()
   }
 
+  public removeMacro(macro: Macro): void {
```
LindvedKrvang: Would be nice if it was called `deleteMacro`

src/locale/messages.da.xlf
```diff
@@ -950,6 +950,22 @@
         <source>edit-macro.operation.invalid.action-arguments.message</source>
         <target state="new">Aktion argument</target>
       </trans-unit>
+      <trans-unit id="9039455515279025293" datatype="html">
+        <source>macros-page.delete-macro</source>
+        <target state="new">Slet makro.</target>
+      </trans-unit>
+      <trans-unit id="4136819087870983707" datatype="html">
+        <source>macro-page.confirm-delete.title</source>
+        <target state="new">Bekræft slettet makro.</target>
```
LindvedKrvang: `Bekræft slettet makro` seems a bit odd. It wants us to confirm something already is deleted?
How about: `Bekræft sletning af makro`?

LindvedKrvang (approved):

vushu (approved):

src/locale/messages.da.xlf
RasmusAlbrektsen: Honestly i just felt so weird saying "sletning" or "fjernelse" so i went with that, but it sounds weird too. I'll switch it to your suggestion :-D

PR #350, opened 2025-03-04
Sof 2523/naviagte to standalone shelf

LindvedKrvang:
Navigate to a "StandaloneShelf" when accessing Alba from a non-desktop device.
When on a desktop it is still possible to navigate to this page, but it requires the user to manually type in the route in the URL.

Disclaimer: I have yet to verify it with an actual device since our tablets is currently being set up.


RasmusAlbrektsen (approved):
This is very cool! Can't wait to try it on a tablet!

vushu (approved):
very nice ☺️

PR #351, opened 2025-03-06
Sof 2524/show active panels for active rundown

LindvedKrvang:
When navigating to the standalone-shelf, we are now presented with our ActionsPanels and StaticButtons and it's possible to call them from here on the active Rundown.

The ActionPanels shows the Actions that belongs to the active Rundown. If there is no Rundown active, the panels disappear and message is displayed saying no Rundon is active.

Also in this PR:
- Refactor the ActionPanels from ProducerShelf to be a reuseable component. The logic is almost 1:1. It has just been moved.
- Refactor the static buttons to be their own component.
- The ActionPanelComponent will try to resize itself if its `scrollable` attribute is set to false.

https://github.com/user-attachments/assets/3d06bc6e-fcac-4544-8f92-b6f6f3da8bc6


vushu:
really cool :D

src/app/producer-shelf/components/action-panels/action-panels.component.ts
```diff
@@ -0,0 +1,237 @@
...
+    private readonly configurationService: ConfigurationService,
+    private readonly configurationEventObserver: ConfigurationEventObserver,
+    private readonly rundownEventObserver: RundownEventObserver,
+    private readonly rundownService: RundownService,
+    private readonly actionRankService: ActionRankService,
+    private readonly pieceEntityService: PieceEntityService,
+    private readonly timeService: TimeService,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('StandaloneActionsShelfComponent')
```
RasmusAlbrektsen: Shouldn't this just be `ActionPanelsComponent` ?

src/app/producer-shelf/components/static-buttons/static-buttons.component.html
```diff
@@ -0,0 +1,7 @@
+<sofie-tv2-action-panel
+    title="STATIC BUTTONS"
```
RasmusAlbrektsen: This is just a random note i'd like an opinion on, i only note it here because we are dealing with it:

Do we like the `STATIC BUTTONS` name? It seems unnecessary to call something "static" but i understand the point. Would something like "Quick Actions"?

src/app/standalone-shelf/components/standalone-action-shelf/standalone-action-shelf.component.html
```diff
@@ -1,2 +1,12 @@
-<h2>standalone-action-shelf works!</h2>
-<a [routerLink]="['/', Paths.RUNDOWNS]">Rundowns</a>
+<div *ngIf="rundown; else noActiveRundown" class="wrapper">
+    <sofie-action-panels class="action-panels" [rundown]="rundown" />
+    <sofie-static-buttons [rundown]="rundown" [centerActions]="true" />
+</div>
+
```
RasmusAlbrektsen, line 5: Extra newline here :-)

RasmusAlbrektsen (approved):
Good stuff!

vushu (approved):

src/app/producer-shelf/components/static-buttons/static-buttons.component.html
LindvedKrvang: Good point. I'm personally not a fan of "Static buttons" because what does that even imply?
We should run your suggestion by Jacob :)

src/app/producer-shelf/components/action-panels/action-panels.component.ts
LindvedKrvang: You are absolutely right.

It's been updated.

src/app/producer-shelf/components/static-buttons/static-buttons.component.html
LindvedKrvang: The title and the component has been updated to "Quick Actions"

src/app/standalone-shelf/components/standalone-action-shelf/standalone-action-shelf.component.html
LindvedKrvang: No more

src/app/producer-shelf/components/static-buttons/static-buttons.component.html
LindvedKrvang: @RasmusAlbrektsen I talked to Jacob and he suggest "Static Actions" and I am inclined to use that instead of "Quick Actions"

vushu (approved):

RasmusAlbrektsen (approved):

PR #354, opened 2025-03-12
SOF-2496 Able to configure InputConfiguration to use Macros

LindvedKrvang:
Able to configue a PanelInput to execute a Macro.

Also in this PR:
- Introduce the MacroPicker, which is heavily inspired by the ActionPicker


RasmusAlbrektsen (commented):
Was the top box (and the one you select when choosing "Modifier" or "Action" before this) supposed to say `Action or Macro` or maybe `Action/Macro` or did we just decide to call it `Action` for now?
image

src/app/settings/panel-settings/components/panel-input-configuration-section/panel-input-configuration-section.component.scss
```diff
@@ -44,3 +44,30 @@
...
+
+.selectable-box-container {
+  display: flex;
+  justify-content: center;
+  gap: 10px;
+
+  .selectable-box {
+    width: 100%;
+    border: 0.1rem solid var(--black-color);
+
```
RasmusAlbrektsen: extra newline

RasmusAlbrektsen (approved):
Aside from the comment i was unclear about earlier it looks good! :-)

LindvedKrvang:
> Was the top box (and the one you select when choosing "Modifier" or "Action" before this) supposed to say `Action or Macro` or maybe `Action/Macro` or did we just decide to call it `Action` for now? image

I vaguely recall that we would just leave it for now

src/app/settings/panel-settings/components/panel-input-configuration-section/panel-input-configuration-section.component.scss
LindvedKrvang: It's gone

vushu (approved):
very nice!

PR #358, opened 2025-03-19
SOF-2540 Default to empty string if text is undefined

LindvedKrvang:
If the input configuration text is undefined when setting up the form, we now put a default empty string in the form.


vushu (approved):
Cool.

RasmusAlbrektsen (approved):

PR #360, opened 2025-03-25
static-actions.component also includes system actions.

vushu:
Included systemactions to static-actions.


src/app/producer-shelf/components/static-buttons/static-actions.component.ts
```diff
@@ -68,6 +71,11 @@ export class StaticActionsComponent implements OnInit, OnDestroy {
     })
   }
 
+  private onActionsChanged(actions: Action[]): void {
```
LindvedKrvang: `onXChanged` doesn't refer to what the method is doing. Instead it refers to where you intent this method to be called from.
Methods are not allowed to know anything about outside of its own scope. This is for the simple reason that there is nothing stopping us from using this method in a context that is not what where you intended the method to be used i.e. there is nothing stopping me from doing something like: `rundownEventObserver.onRundownActivation(() => this.onActionsChanged(this.actions))`

src/app/producer-shelf/components/static-buttons/static-actions.component.ts
```diff
@@ -48,15 +50,16 @@ export class StaticActionsComponent implements OnInit, OnDestroy {
...
       .subscribeToRundownActions(this.rundown.id)
-      .then(actionsObservable =>
-        actionsObservable.subscribe(actions => {
-          this.actions = this.getValidTv2Actions(actions)
-          this.updateStaticActions()
-        })
-      )
+      .then(actionsObservable => actionsObservable.subscribe(actions => this.onActionsChanged(actions)))
       .then(actionsSubscription => (this.actionsSubscription = actionsSubscription))
       .catch(error => this.logger.data(error).error('Failed subscribing to actions'))
```
LindvedKrvang: We should remove the subscription to RundownActions.
This subscription is from before the StaticActionsComponent became its own component and it "shared" the Actions with what was going in to the ActionPanels. 
The ActionPanels required the Rundown specific Actions, but the StaticActionsComponent don't. To prove the point: Only Actions that have been configured to be shown in the StaticActionsPanel will be shown, but when we configured which Actions we don't show Rundown specific Actions. So it should be enough to just fetching the SystemActions.

LindvedKrvang (changes requested):
Good work :)

The request for changes is solely to ensure that we no longer listen for RundownActions (see comment).

src/app/producer-shelf/components/static-buttons/static-actions.component.ts
vushu: Okay 👍

src/app/producer-shelf/components/static-buttons/static-actions.component.ts
vushu: What would be a reasonable name maybe setStaticActions?

RasmusAlbrektsen (approved):

PR #361, opened 2025-03-25
Added shared tabs component

vushu:
I have added a tabs-component:

I tested it in `edit-keyboard-mapping.component.html`
```html
 <sofie-tabs class="ps-5">
    <sofie-tab [tabTitle]="'Action'" [active]="true">
      <ng-template >
        <sofie-action-selector [selectedAction]="selectedAction" (actionSelected)="selectAction($event)" />
      </ng-template>
    </sofie-tab>
    <sofie-tab [tabTitle]="'Macro'" [active]="true">
      <ng-template >
        <h1>My macronies</h1>
        <sofie-action-selector [selectedAction]="selectedAction" (actionSelected)="selectAction($event)" class="ps-5" />
      </ng-template>
    </sofie-tab>
  </sofie-tabs>
  ```
  

https://github.com/user-attachments/assets/7a8d9f92-ed7d-4ff3-b695-cbab6c55d814

### CODE REVIEW

File: src/app/shared/components/tabs/tab.component.ts [hunk 1]
```diff
@@ -0,0 +1,12 @@
...
+
+@Component({
+  selector: 'sofie-tab',
+  template: '',
+})
+export class TabComponent {
+  @Input()
+  public tabTitle!: string
+  @Input() public active: boolean = false
+  @ContentChild(TemplateRef) public template!: TemplateRef<never>
```
LindvedKrvang: Please be consistent with new-lining or not new-lining decorators

src/app/shared/components/tabs/tab.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, ContentChild, Input, TemplateRef } from '@angular/core'
+
+@Component({
+  selector: 'sofie-tab',
+  template: '',
+})
+export class TabComponent {
+  @Input()
+  public tabTitle!: string
```
LindvedKrvang: We are in the context of a `Tab` so we know the title belongs to the Tab, so just call it `title`.
Then it reads `tab.title` instead of `tab.tabTitle`. You would not call it `person.personAddress`, but `person.address`.

Also, why the `!`? Attributes are non-optional be default. There is no need to include the `!`.

src/app/shared/components/tabs/tab.component.ts
LindvedKrvang: No need for the `!`.

src/app/shared/components/tabs/tab.component.ts
```diff
@@ -0,0 +1,12 @@
...
+  @Input() public active: boolean = false
```
LindvedKrvang: This doesn't need to be an `@Input()`. 
As far as I can see, the only purpose of this being an Input is so you can decide which Tab is the first to be active. 
An easy way of doing this is just to have the `Tabs` component select the first Tab as active in its `ngAfterContentInit`.
It still needs to be public, but by not having it as an Input, then it becomes even easier to use the Tabs :D

src/app/shared/components/tabs/tabs.component.html
```diff
@@ -0,0 +1,13 @@
...
+       [class.active]="tab.active"
+       (click)="selectTab(tab)">
+    {{ tab.tabTitle }}
+  </div>
+</div>
+<ng-container *ngFor="let tab of tabs">
+  <div *ngIf="tab.active" class="tab-content">
+    <ng-container *ngTemplateOutlet="tab.template" />
+  </div>
+</ng-container>
```
LindvedKrvang: I haven't tested this, but instead of iterating over all the Tabs you could have an `activeTab` attribute in the .ts file. Then this could be:
```
<div class="tab-content">
  <ng-container *ngTemplateOutlet="activeTab.template" />
</div>
```

Then your .ts file could look something like:
```
@ContentChildren(TabComponent) protected tabs: QueryList<TabComponent>
public activeTab: TabComponent

...

public selectTab(selectedTab: TabComponent): void {
  this.activeTab = selectedTab
}
```
This way you could also get rid of the `active` attribute on `Tab` (which I think we should because the `active` attribute references something outside the scope of the `Tab` component, and methods/components shouldn't know anything about outside of its own scope)

src/app/shared/components/tabs/tab.component.ts
LindvedKrvang: Update: See my other comment about potentially getting rid of the `active` attribute

src/app/shared/components/tabs/tabs.component.scss
```diff
@@ -0,0 +1,18 @@
+.tabs {
+  display: flex;
+  border-bottom: 2px solid var(--gray-color);
```
LindvedKrvang: We would to use `rem` over `px`. 
This comment goes for the entire file

src/app/shared/components/tabs/tabs.component.scss
```diff
@@ -0,0 +1,18 @@
...
+  border: 1px solid var(--gray-color);
+  border-bottom: none;
+  background: var(--light-grey-color);
+}
+.tab.active {
+  background: var(--white-color);
+  font-weight: bold;
+}
+.tab-content {
+  padding: 20px;
```
LindvedKrvang: It wasn't possible to use any of our padding classes for this?

src/app/shared/components/tabs/tabs.component.ts
```diff
@@ -0,0 +1,26 @@
...
+@Component({
+  selector: 'sofie-tabs',
+  templateUrl: './tabs.component.html',
+  styleUrls: ['./tabs.component.scss'],
+})
+export class TabsComponent implements AfterContentInit {
+  @ContentChildren(TabComponent) protected tabs!: QueryList<TabComponent>
+
+  @Input()
+  public tabTitle: string
```
LindvedKrvang: What's the point of this attibute?
It's the individual Tabs that needs an title. This component is the container of the Tabs.

src/app/shared/components/tabs/tabs.component.ts
```diff
@@ -0,0 +1,26 @@
+import { AfterContentInit, Component, ContentChildren, Input, QueryList } from '@angular/core'
+import { TabComponent } from './tab.component'
+
+@Component({
+  selector: 'sofie-tabs',
+  templateUrl: './tabs.component.html',
+  styleUrls: ['./tabs.component.scss'],
+})
+export class TabsComponent implements AfterContentInit {
+  @ContentChildren(TabComponent) protected tabs!: QueryList<TabComponent>
```
LindvedKrvang: No need for the `!`.

src/app/shared/components/tabs/tabs.component.ts
```diff
@@ -0,0 +1,26 @@
...
+    const activeTab = this.tabs.find(tab => tab.active) || this.tabs.first
+    if (activeTab) {
+      this.selectTab(activeTab)
+    }
+  }
+
+  public selectTab(selectedTab: TabComponent): void {
+    this.tabs.forEach(tab => (tab.active = false))
+    selectedTab.active = true
+  }
```
LindvedKrvang: As per my other comments, if you refactor away from using the `active` attribute this could become:
```
public ngAfterContentInit(): void {
  this.selectTab(this.tabs.first)
}

public selectTab(tab: TabComonent): void {
  this.selectedTab = tab
}
```

src/app/shared/components/tabs/tabs.component.ts
```diff
@@ -0,0 +1,26 @@
...
+  styleUrls: ['./tabs.component.scss'],
+})
+export class TabsComponent implements AfterContentInit {
+  @ContentChildren(TabComponent) protected tabs!: QueryList<TabComponent>
+
+  @Input()
+  public tabTitle: string
+
+  public ngAfterContentInit(): void {
+    const activeTab = this.tabs.find(tab => tab.active) || this.tabs.first
```
LindvedKrvang: Please add types :)

LindvedKrvang (commented):
Good work :)

Despite the many comments, they are on the minor side and this is a pretty good implementation for a reuseable component :D

src/app/shared/components/tabs/tabs.component.ts
vushu: whoops

src/app/shared/components/tabs/tabs.component.ts
vushu: Doesn't the ! mean it must be set? :d

LindvedKrvang:
Your file placement is not following our convention.
We have one "component" folder for each component. This folder contains at maximum four files: The .ts file, the .html file, the .scss file and the .spec.ts file.
You have put the .ts file for the `Tab` component inside of the `Tabs` component folder. The `Tab` component should have its own component folder

vushu:
I tried using ng-content, but I had trouble using it, so I kept the ng-template implementation, I have also read that template is recommended for conditional rendering https://v17.angular.io/guide/content-projection#conditional-content-projection

src/app/shared/components/tabs/tabs.component.scss
```diff
@@ -0,0 +1,17 @@
+.tabs {
+  display: flex;
+  border-bottom: 0.2rem solid var(--gray-color);
+}
+.tab {
```
LindvedKrvang: Would be nice with an empty line between the closing bracket of a class and the start of the next one :) 
Goes for the entire file.

src/scss/_sofie-color-vars.scss
```diff
@@ -3,6 +3,7 @@ body {
   --danger-color: #c62828;
   --white-color: #ffffff;
   --light-gray-color: #f6f6f6;
+  --light-grey-color: #f1f1f1;
```
LindvedKrvang: We already have a `light-grey-color`. It's just using the american spelling with an `a` instead of an `e`.
We are writing american english in our repositories.

This new "light gray" attribute should be deleted so we don't have two light gray (that even have different values).

LindvedKrvang (approved):
Very nice and simple Tab implementation. Good job :D 

Do you still need the HTML file for `Tab`? It's empty and I think you introduced it after our conversation of potentially using `ng-content`.

Note: The approve assumes that you delete the `--light-grey-color` attribute. See comment.

src/scss/_sofie-color-vars.scss
vushu: I just like the grey color bit more since it made the text a bit more pronounced, hence why I made it, but I will just delete it.

RasmusAlbrektsen (approved):

PR #362, opened 2025-03-26
SOF-2550 Able to export/import Modifiers

LindvedKrvang:
Able to export and import Modifiers.

When importing, if the Modifier already exist we simply skip it. We can edit Modifiers at this point, so if we want to imort an already existing Modifier then there is a 99% chance that it is the exact same Modifier.

Note: When creating a new Modifier for some reason the `id` had the value of `null`. Now we are setting it to undefined.


vushu (approved):
👍

RasmusAlbrektsen (approved):

PR #363, opened 2025-03-26
SOF-2549 Able to import/export Macros

LindvedKrvang:
Note: Dependent on https://github.com/tv2/alba-web-client/pull/362

Able to export and import Macros.

If a Macro already exist we simply update it. If it doesn't exists, we create a new Macro for it.


vushu (approved):
Nice

RasmusAlbrektsen (approved):

PR #364, opened 2025-03-26
Sof 2551/able to import export panel layouts

LindvedKrvang:
Note: Dependent on https://github.com/tv2/alba-web-client/pull/363

Able to import and export PanelLayoutConfigurations.
If the PanelLayoutConfiguration to be imported doesn't exist we create a new one. If it does already exist we update it.

Also in this PR:
- JsonImportButton now takes an optional `infoMessage`. If this message is provided an info dialog will appear with the message before opening the file input for file selection.


vushu (approved):
Cool!

RasmusAlbrektsen (approved):

PR #365, opened 2025-03-27
fix: No longer limit the static action panel in width

RasmusAlbrektsen:


src/app/producer-shelf/components/static-buttons/static-actions.component.html
```diff
@@ -1,7 +1,8 @@
 <sofie-tv2-action-panel
-    i18n-title title="static-actions.title"
-    [actions]="staticActions"
-    (executeAction)="executeAction($event)"
-    [isScrollable]="false"
-    [centerActions]="centerActions"
```
LindvedKrvang, line 6: We are indenting by two spaces, not four

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -56,26 +56,26 @@
...
+              <sofie-number-input
+                      *ngSwitchCase="ActionArgumentSchemaType.NUMBER"
+                      [label]="selectedAction?.argument?.name!"
+                      [formControlName]="'actionArguments'"
+                      [helpText]="selectedAction?.argument?.description!"
+                      [isRequired]="true"
+                      [errorMessage]="selectedActionErrorMessage" />
+          </ng-container>
       </ng-container>
-    </div>
```
LindvedKrvang: We are indenting by two spaces, not four

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -108,11 +108,14 @@ export class EditKeyboardMappingComponent implements OnInit {
   }
 
   private addActionArgumentsFormControl(): void {
-    this.triggerDataForm.setControl(KEYBOARD_ACTION_ARGUMENTS_CONTROL_ID, this.formBuilder.control(this.keyboardMapping?.trigger.data.actionArguments, Validators.required))
+    if (this.keyboardMapping?.trigger.type === TriggerType.MACRO) {
```
LindvedKrvang: Instead of checking if the type is a Macro, you should check if the type _isn't_ an Action.

These are guard statements which are used to ensure that the data is in the state we want them to be in. We want the data to be an Action so _that's_ the value we check for.

If we add a third thing to the Macro and Actions then this check will no longer be good enough (with no compiler to warn us), but checking that it isn't an Action will still work.

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
```diff
@@ -129,10 +129,14 @@ export class TriggerProducerKeyBindingService implements KeyBindingService {
   }
 
   private findActionForContentPlaceholderAction(placeholderAction: Tv2ContentPlaceholderAction, trigger: Trigger): Tv2Action | undefined {
+    if (trigger.type == TriggerType.MACRO) {
```
LindvedKrvang: Refer to my other comment; We should check that the type _isn't_ an _Action_.

src/app/producer-shelf/components/static-buttons/static-actions.component.scss
```diff
@@ -0,0 +1,9 @@
+:host {
+  .static-action-panel {
+    overflow-y: auto;
+    min-width: 340px;
+    max-width: 340px;
+    max-height: 1200px;
```
LindvedKrvang: We should avoid hardcoding values like these.

I haven't tested this, but I think the issue comes from the use of `window.innerHeight` in Tv2ActionPanelComponnet.
`window.innerHeight` returns the height of the _content area_ of the browser. So when we have two screens stacked ontop of each other showing the same browser, then the height of the content area is large.

I suggest that we try to use `window.screen.availHeight` which should return the height available for content of the current screen. 

Sidenote: We would also no longer need the overfly property, so this CSS class can be removed :)

LindvedKrvang (commented):
Nice to see these things being fixed :)

I haven't approved yet since we shouldn't hardcode height and width values for StaticActions. I have suggested an alternative

vushu:
@RasmusAlbrektsen @LindvedKrvang I made the PR changes. I tried using `window.screen.availHeight`, but the same issue still persist.

vushu (approved):
👌

PR #366, opened 2025-04-04
Sof 2519/implemented macro binding for keyboard

vushu:
Pardon this big PR, this draft enables you to create a macro as keyboard mapping. 

 I haven't renamed all variables yet since I want some feedback first.
I added Tv2Action and Macro into a toplevel interface which I ended up calling Tv2Binding, so that I can differentiate between these 2 types.

- I have created a macro-selector 
- made changes to components to handle this new toplevel interface.

Update:

- changed the common type from TV2Binding to ShortcutBinding and added them to a seperate file.

- renamed variables, functions, etc to reflect the new change.


src/app/shared/models/tv2-action.ts
```diff
@@ -1,8 +1,28 @@
 import { ActionArgumentSchemaType, PartAction, PieceAction, PlaceholderAction, SystemAction } from './action'
 import { PlaceholderActionScope, PlaceholderActionType } from './action-type'
+import { Macro } from './macro'
+
+export type Tv2Binding = Tv2ActionBinding | Tv2MacroBinding
```
LindvedKrvang: We are slowly trying to get rid of the `Tv2` prefix. So you don't need to put in new interace names. We encourage you not to :) 

`Binding` is quite broad/vague. What kind of binding is it? If I am not mistaken (and I might be) then this "binding" is used to bind keyboard shortcuts? If that's the case it could potentially be called `KeyboardShortcutBinding` and then you would have `ActionKeyboardShortcutBinding` and `MacroKeyboardShortcutBinding`.

Sidenote: The file is called `tv2-action.ts`. `binding` does not belong in this file. If this is truly a "KeyboardShortcutBinding", then we shouldn't put it in the shared folder, but instead put it in the module that contains are keyboard :)

src/app/producer-shelf/components/action-panels/action-panels.component.ts
```diff
@@ -155,11 +155,21 @@ export class ActionPanelsComponent implements OnInit, OnChanges, OnDestroy {
       .map(actionPanel => {
         return {
           name: actionPanel.name,
-          actions: this.actions.filter(action => !!action.rundownId && actionPanel.actionFilter.includes(action.metadata.contentType)).sort((a, b) => a.rank - b.rank),
+          actions: this.sortTv2BindingsByRank(actionPanel),
         }
       })
   }
 
+  private sortTv2BindingsByRank(actionPanel: ShelfActionPanelConfiguration): Tv2Action[] {
```
LindvedKrvang: There is a mismatch between method name of implementation.
The name references "binding", but it accepts a `ShelfActionPanelConfiguration` as the parameter and nowhere in the implementation does a "binding" occur. It does even return bindings.

I'm willing to bet that this is mainly because of the missing naming refactor that you have yet to do? :)

Sidenote: Is it you or ts-lint that added the extra curly brackets for the arrow functions?

src/app/producer-shelf/components/tv2-action-panel/tv2-action-panel.component.ts
```diff
@@ -131,4 +131,8 @@ export class Tv2ActionPanelComponent implements OnInit, OnChanges {
   public trackAction(_index: number, action: Tv2Action): string {
     return action.id
   }
+
+  public getActionRank(action: Tv2Action): number {
+    return action.rank
+  }
```
LindvedKrvang: This method seems redundant. 
In my own opinion I think the places it is being used in the html was better without this helper method.

src/app/rundown-view/components/rundown/rundown.component.ts
```diff
@@ -173,7 +173,10 @@ export class RundownComponent implements OnInit, OnDestroy, OnChanges {
   }
 
   private onActionsChanged(actions: Action[]): void {
-    this.videoClipActions = actions.filter((action): action is Tv2VideoClipAction => (action as Tv2Action).metadata?.contentType === Tv2ActionContentType.VIDEO_CLIP)
+    this.videoClipActions = actions.filter((action): action is Tv2VideoClipAction => {
+      const currentAction = action as Tv2Action
+      return currentAction.metadata?.contentType === Tv2ActionContentType.VIDEO_CLIP
+    })
```
LindvedKrvang: What's the reason for adding the curly brackets here?

src/app/rundown-view/services/mini-shelf-cycle.service.ts
```diff
@@ -39,13 +39,13 @@ export class MiniShelfCycleService {
...
-      action => segment.name === action.name && action.metadata.contentType === Tv2ActionContentType.VIDEO_CLIP && (action as Tv2VideoClipAction).metadata.fileName === videoClipFilename
-    )
+    const videoClipAction: Tv2Action | undefined = rundownActions.find(action => {
+      return segment.name === action.name && action.metadata.contentType === Tv2ActionContentType.VIDEO_CLIP && (action as Tv2VideoClipAction).metadata.fileName === videoClipFilename
+    })
     if (!videoClipAction) {
       throw new Error('Unable to find video clip action.')
     }
-    return videoClipAction
+    return videoClipAction as Action
```
LindvedKrvang: This is a dangerous `as` you are doing here.
Why change the the `Action` type to `Tv2Action` on line 42 that then forces you to use this "type by-pass"?

src/app/shared/models/tv2-action.ts
```diff
@@ -1,8 +1,28 @@
...
+}
+
+export interface Tv2ActionBinding {
+  bindingType: Tv2BindingType.ACTION_BINDING
+  binding: Tv2Action
+}
+
+export interface Tv2MacroBinding {
+  bindingType: Tv2BindingType.MACRO_BINDING
+  binding: Tv2Macro
```
LindvedKrvang: Having these `binding` attributes is going to look weird when calling them. It's going to look like:
```
const binding: Binding = // Instantiate a Binding here
console.log(binding.binding.id)
```
What kind of id is it we are dealing with?

Since you know whether it is an Action or a Macro you, then simply call the attribute that. Then the above will look like:
```
const binding: Binding = // Instantiate a Binding here
console.log(binding.action.id)
```
Making it quite clear what kind of id we are dealing with.

src/app/shared/models/tv2-action.ts
```diff
@@ -1,8 +1,28 @@
...
+}
+
+export interface Tv2Macro extends Macro {}
```
LindvedKrvang: This interface is redundant and should be deleted.
As stated in a different comment, we are trying to get away from the `tv2` prefix

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
```diff
@@ -107,12 +107,13 @@ export class TriggerProducerKeyBindingService implements KeyBindingService {
       if (!action) {
         return
       }
-      this.triggersWithAction.set(trigger, action)
+      this.triggersWithAction.set(trigger, { bindingType: Tv2BindingType.ACTION_BINDING, binding: action })
```
LindvedKrvang: Creating the the "binding" on a separate line would be nice.
Right now you are doing two things on line 110: Creating a binding _and_ assigning it to a Trigger.

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
```diff
@@ -186,22 +187,26 @@ export class TriggerProducerKeyBindingService implements KeyBindingService {
...
-    return action.name
+    return action.binding.name
   }
 
-  private getKeyBindingBackgroundColour(trigger: Trigger, action: Tv2Action): string {
-    if (action.metadata.contentType === Tv2ActionContentType.SPLIT_SCREEN) {
+  private getKeyBindingBackgroundColour(trigger: Trigger, action: Tv2Binding): string {
+    if (action.bindingType !== Tv2BindingType.ACTION_BINDING) {
+      return ''
+    }
```
LindvedKrvang: It is also possible for Macros to have a different background color through the `trigger.data.overrideColor`.
This if-statement makes that piece of code unreachable for anything not Actions.

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
```diff
@@ -233,10 +238,14 @@ export class TriggerProducerKeyBindingService implements KeyBindingService {
     }
   }
 
-  private getBackgroundColorForSplitScreen(trigger: Trigger, action: Tv2Action): string {
+  private getBackgroundColorForSplitScreen(trigger: Trigger, action: Tv2Binding): string {
```
LindvedKrvang: We know that a SplitScreen is always an Action and never a Macro, so you could use your `Tv2ActionBinding` as the type for the parameter.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -66,14 +65,25 @@ export class EditKeyboardMappingComponent implements OnInit {
       triggerOn: this.formBuilder.control<KeyEventType>(this.keyboardMapping?.trigger.data.triggerOn ?? KeyEventType.RELEASED, [Validators.required]),
       mappedToKeys: this.formBuilder.control<string[]>(this.keyboardMapping?.trigger.data.mappedToKeys ?? []),
     })
+    this.setActionTriggerForm('')
```
LindvedKrvang, line 68: This won't work with when we edit a Trigger, because it will not set the already selected Action/Macro when opening the sidebar

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -97,18 +107,20 @@ export class EditKeyboardMappingComponent implements OnInit {
...
 
+  public selectMacro(macro: Tv2Binding): void {
+    this.selectedBinding = macro
+    this.setMacroTriggerForm(this.selectedBinding.binding.id)
     this.doesSelectedActionHaveActionArguments() ? this.addActionArgumentsFormControl() : this.removeActionArgumentsFormControl()
   }
 
   private addActionArgumentsFormControl(): void {
-    if (this.keyboardMapping?.trigger.type !== TriggerType.ACTION) {
+    if (this.keyboardMapping?.trigger.type === TriggerType.MACRO) {
```
LindvedKrvang: ActionArguments are only for Actions, so you should continue to check `!== TriggerType.ACTION`.
If/when we add a third TriggerType, this if-statment won't be enough anymore.

The philosophy is that we want to ensure that our data is in a given state, so we check that if the data _is not_ in the desired state, then we stop execution. I.e. if we want a red car, then we check if the car _isn't_ red. If we only checked if the car was blue, then all yellow, green, white etc cars would not be filtered away.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -119,14 +131,20 @@ export class EditKeyboardMappingComponent implements OnInit {
   }
 
   public doesSelectedActionHaveActionArguments(): boolean {
-    return !!this.selectedAction?.argument
+    if (this.selectedBinding?.bindingType !== 'ACTION_BINDING') {
```
LindvedKrvang: You should use the actual enum instead of a magic string

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -119,14 +131,20 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+      return false
+    }
+    return !!this.selectedBinding?.binding.argument
   }
 
   public getSelectedActionArgumentType(): ActionArgumentSchemaType | undefined {
-    if (!this.selectedAction?.argument) {
+    if (this.selectedBinding?.bindingType !== 'ACTION_BINDING') {
```
LindvedKrvang: You should use the actual enum instead of a magic string

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -185,13 +203,42 @@ export class EditKeyboardMappingComponent implements OnInit {
       case KEYBOARD_ACTION_CONTROL_ID:
         return $localize`triggers.action.label`
       case KEYBOARD_ACTION_ARGUMENTS_CONTROL_ID:
-        return this.selectedAction?.argument?.name!
+        if (this.selectedBinding?.bindingType !== 'ACTION_BINDING') {
```
LindvedKrvang: You should use the actual enum instead of a magic string

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -185,13 +203,42 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+
+        return this.selectedBinding?.binding.argument?.name!
       default:
         return controlId
     }
   }
 
   public get selectedActionErrorMessage(): string {
-    return $localize`${this.selectedAction?.argument?.name} action-selector.action-arguments.error`
+    if (this.selectedBinding?.bindingType !== 'ACTION_BINDING') {
```
LindvedKrvang: You should use the actual enum instead of a magic string

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -185,13 +203,42 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+      return ''
+    }
+
+    return $localize`${this.selectedBinding?.binding.argument?.name} action-selector.action-arguments.error`
+  }
+
+  public getSelectedDescription(): string {
+    if (this.selectedBinding?.bindingType !== 'ACTION_BINDING') {
```
LindvedKrvang: You should use the actual enum instead of a magic string

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -185,13 +203,42 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+      return ''
+    }
+    return this.selectedBinding?.binding.description ?? ''
+  }
+
+  public getSelectedArgumentName(): string {
+    if (this.selectedBinding?.bindingType !== 'ACTION_BINDING') {
```
LindvedKrvang: You should use the actual enum instead of a magic string

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -185,13 +203,42 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+      return ''
+    }
+    return this.selectedBinding?.binding.argument?.name ?? ''
+  }
+
+  public getSelectedArgumentDescription(): string {
+    if (this.selectedBinding?.bindingType !== 'ACTION_BINDING') {
```
LindvedKrvang: You should use the actual enum instead of a magic string

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -81,20 +81,33 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
...
-        action: this.actions.find(action => action.id === this.getTriggerTypeId(trigger)) ?? this.createUnknownAction(),
+        action: this.findTv2Binding(trigger),
       } as KeyboardMapping
     })
   }
 
-  private createUnknownAction(): Tv2Action {
-    return {
+  private findTv2Binding(trigger: Trigger): Tv2Binding {
+    const result = this.actions.find(action => action.id === this.getTriggerTypeId(trigger))
```
LindvedKrvang: Missing type for the `result` variable. ´result´ is a broad/non-descript name

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -81,20 +81,33 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
...
         trigger,
-        action: this.actions.find(action => action.id === this.getTriggerTypeId(trigger)) ?? this.createUnknownAction(),
+        action: this.findTv2Binding(trigger),
       } as KeyboardMapping
     })
   }
 
-  private createUnknownAction(): Tv2Action {
-    return {
+  private findTv2Binding(trigger: Trigger): Tv2Binding {
```
LindvedKrvang: The method is called `findTv2Binding` but it only tries to find for actions.

src/app/settings/shelf-settings/components/edit-static-buttons/edit-static-buttons.component.ts
```diff
@@ -41,4 +45,10 @@ export class EditStaticButtonsComponent implements OnInit {
   public cancel(): void {
     this.onCancel.emit()
   }
+
+  public getBindings(): Tv2Binding[] {
+    return this.updatedSelectedStaticActions.map(action => {
+      return { bindingType: Tv2BindingType.ACTION_BINDING, binding: action }
+    })
+  }
```
LindvedKrvang: This method is simply called `getBindings` which indicates that it will just return all bindings that it have, no questions asked. That is _not_ how it is implemented.
The implementations first and foremost only returns "ActionBindings" and only for Actions that have been selected for StaticActions. That is very different than "all bindings".

Reading were this method is being used, it now gives the impression of showing something vastly different, than what it actually does. It used to iterate over `updatedSelectedStatcActions`, now it looks like it is iterating over `all bindings`, which it is in reality not doing.

src/app/settings/shelf-settings/components/static-buttons-configuration-card/static-buttons-configuration-card.component.ts
```diff
@@ -92,4 +92,10 @@ export class StaticButtonsConfigurationCardComponent implements OnInit, OnDestro
   public ngOnDestroy(): void {
     this.shelfConfigurationEventSubscription.unsubscribe()
   }
+
+  public getBindings(): Tv2Binding[] {
```
LindvedKrvang: See my other comment about the vastly different indication that comes from the name `getBindings` in contrast to what is acutally happening here.

src/app/shared/components/action-box/action-box.component.html
```diff
@@ -1,3 +1,11 @@
...
+  <div class="{{ getActionContentType() | actionColorClass }}" [ngClass]="{ highlight }">
+    {{ action.binding.name }}
+  </div>
+</ng-container>
+
+<ng-template #macroBinding>
+  <div class="macro-color" [ngClass]="{ highlight }">
+    {{ action.binding.name }}
+  </div>
+</ng-template>
```
LindvedKrvang: This component is called `action-box.component`. As long as it have that name, it shouldn't be able to show Macros.

src/app/shared/components/action-box/action-box.component.html
```diff
@@ -1,3 +1,11 @@
...
-    {{ action.name }}
-</div>
+<ng-container *ngIf="isAction(); else macroBinding" >
+  <div class="{{ getActionContentType() | actionColorClass }}" [ngClass]="{ highlight }">
+    {{ action.binding.name }}
+  </div>
+</ng-container>
+
+<ng-template #macroBinding>
+  <div class="macro-color" [ngClass]="{ highlight }">
```
LindvedKrvang: It is also possible to override the color of Macros, so it shouldn't have a single color class hardcoded.
The only difference between an ActionTrigger and a MacroTrigger is one call an Action, the other a Macro. All the functionality of being a Trigger still applies to both.

src/app/shared/components/action-box/action-box.component.ts
```diff
@@ -1,12 +1,23 @@
...
   @Input() public highlight: boolean
+
+  public getActionContentType(): Tv2ActionContentType {
+    if (this.action.bindingType !== Tv2BindingType.ACTION_BINDING) {
+      return Tv2ActionContentType.UNKNOWN
+    }
+    return this.action.binding.metadata.contentType
+  }
+
+  public isAction(): boolean {
```
LindvedKrvang: We are in a component called `action-box`. If the thing we are parsing in here isn't an Action, something has gone wrong before we reach this place, meaning this method shouldn't be necessary.

src/app/shared/components/macro-selector/macro-selector.component.html
```diff
@@ -0,0 +1,8 @@
+<sofie-text-input i18n-label label="action-selector.select-action" i18n-placeholder placeholder="action-selector.search-action" [(value)]="macroSearchQuery" />
```
LindvedKrvang: We are in a the MacroSelector. We shouldn't have any references to "action". You are using the translations for actions. (and some CSS classes further down)

src/app/shared/components/macro-selector/macro-selector.component.scss
```diff
@@ -0,0 +1,13 @@
+.actions-container {
```
LindvedKrvang: `.macro-container`

src/app/shared/components/macro-selector/macro-selector.component.scss
```diff
@@ -0,0 +1,13 @@
...
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr 1fr 1fr;
+  grid-template-rows: min-content min-content min-content min-content min-content;
+  width: 680px;
+  height: 650px;
+  overflow-y: auto;
+  overflow-x: hidden;
+}
+
+sofie-action-box {
```
LindvedKrvang: `sofie-macro-box`

src/app/shared/components/macro-selector/macro-selector.component.ts
```diff
@@ -0,0 +1,71 @@
...
+import { MacroStateService } from '../../../core/services/macro-state.service'
+import { Subject, takeUntil } from 'rxjs'
+
+@Component({
+  selector: 'sofie-macro-selector',
+  templateUrl: './macro-selector.component.html',
+  styleUrls: ['./macro-selector.component.scss'],
+})
+export class MacroSelectorComponent implements OnInit, OnDestroy {
+  @Input() public selectedMacro?: Tv2Binding
```
LindvedKrvang: You know this is a Macro. Why not use the specific MacroBinding type?
This seems to be a trend for you in places where we now the specific type, the general type is used forcing us to make extra type checks.

src/app/shared/components/macro-selector/macro-selector.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  constructor(
+    private readonly macroStateService: MacroStateService,
+    logger: Logger
+  ) {
+    this.logger = logger.tag('MacroSelector')
+  }
+
+  public ngOnInit(): void {
+    this.logger.debug('Initialize macro selector')
```
LindvedKrvang: This statement seems redundant.

src/app/shared/components/macro-selector/macro-selector.component.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.macroSelected.emit(this.selectedMacro)
+  }
+
+  public doesMacroMatchSearchQuery(action: Tv2Binding): boolean {
+    if (!this.macroSearchQuery) {
+      return true
+    }
+    const lowerCasedSearchQuery: string = this.macroSearchQuery.toLowerCase()
+    const doesNameMatch: boolean = action.binding.name.toLowerCase().includes(lowerCasedSearchQuery)
+    const isActionSelected: boolean = !!this.selectedMacro && action.binding.id === this.selectedMacro.binding.id
```
LindvedKrvang: `isMacroSelected`

src/app/shared/components/macro-selector/macro-selector.component.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.macros = macros
+      })
+  }
+
+  public selectMacro(action: Tv2Binding): void {
+    this.selectedMacro = action
+    this.macroSelected.emit(this.selectedMacro)
+  }
+
+  public doesMacroMatchSearchQuery(action: Tv2Binding): boolean {
```
LindvedKrvang: `macro`

src/app/shared/components/macro-selector/macro-selector.component.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (!this.macroSearchQuery) {
+      return true
+    }
+    const lowerCasedSearchQuery: string = this.macroSearchQuery.toLowerCase()
+    const doesNameMatch: boolean = action.binding.name.toLowerCase().includes(lowerCasedSearchQuery)
+    const isActionSelected: boolean = !!this.selectedMacro && action.binding.id === this.selectedMacro.binding.id
+    return doesNameMatch || isActionSelected
+  }
+
+  public isSelectedMacro(action: Tv2Binding): boolean {
```
LindvedKrvang: `macro`

LindvedKrvang (commented):
A lot of curly brackets have been added to arrow functions that don't really need curly brackets. Is that you or your IDE?

A general note for you that doesn't just apply for this PR: Slow down on the copy-pasting. You are letting a lot of small details fall through the cracks. The main culprit here is that you are not renaming things when you copy-paste. Feel free to continue to copy-paste, but you should spend more time ensuring that what you have pasted fits the place it has been pasted in.

On the introduction on the `Binding` type:
On an initial glance this seems like an unnecessary complication. We already have the distinction in `Trigger`, `ActionTrigger` and `MacroTrigger`. Why do we need a second abstraction to tell us what we already have?
Also, there are a lot of places where we should know whether we have an `ActionBinding` or a `MacroBinding` where we have just been supplied `Binding` forcing us to make extra unnecessary type checks.

vushu:
@LindvedKrvang I have made it work first with copy pasting, then I will begin the renaming process, so it's intended.

Can't see another solution then introducing another top level class which holds both a macro or an action, since e.g the keyboard-mapping page needs to display all it's binding it needs to handle both types.

src/app/shared/components/macro-selector/macro-selector.component.ts
vushu: This isn't possible to do, since the html handles the public type Tv2Binding.

src/app/shared/models/tv2-action.ts
vushu: nice names I will take it :)

src/app/producer-shelf/components/action-panels/action-panels.component.ts
vushu: Yeah i think my IDE is funky. I will look into it

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vushu: That is true that needs to change when we make the edit task. SOF-2554 I have added this in the description of the task.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vushu: whoops

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
vushu: That weird I thought I fixed that :d

src/app/shared/models/tv2-action.ts
vushu: There is a pro and con in why I have done this.

I will need to do an if to check what kind of Binding it is so that I can access that property

```
if need to do an if (binding.bindingType === ACTION_BINDING) {
     binding.action.id
}
```

whereas if they have the same common property, then I can access the id since that is their commonality.

a macro and an action has 2 properties in common:
name and id
which I will be able to access without needing to check for their underlying type.


the con is that it may be less clear i guess.

src/app/shared/components/action-box/action-box.component.html
vushu: The color you are talking about is for the trigger on the keyboard I think.
whereas the color here is for the box look for when searching for macro.

src/app/shared/components/action-box/action-box.component.html
vushu: I will rename this.

src/app/shared/components/action-box/action-box.component.ts
vushu: I will rename this component and it will be handling macros and actions now.

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
vushu: Yes this I will leave for next task SOF-2520 where we will visualize the macro on the keyboard.

vushu:
I did alot of renaming, I may have missed some.

src/app/shared/models/tv2-action.ts
RasmusAlbrektsen: Is the binding not always have something it should "trigger"? Is it weird to just call it `trigger`?

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
```diff
@@ -129,7 +134,7 @@ export class TriggerProducerKeyBindingService implements KeyBindingService {
   }
 
   private findActionForContentPlaceholderAction(placeholderAction: Tv2ContentPlaceholderAction, trigger: Trigger): Tv2Action | undefined {
-    if (trigger.type !== TriggerType.ACTION) {
+    if (trigger.type === TriggerType.MACRO) {
```
RasmusAlbrektsen: Why did we change this guard away from `if (trigger.type !== TriggerType.ACTION)`?
Shouldn't we only care if we have an action, thereby not care whenever something else reached here?

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
```diff
@@ -186,22 +191,26 @@ export class TriggerProducerKeyBindingService implements KeyBindingService {
     }
   }
 
-  private getTriggerLabel(trigger: Trigger, action: Tv2Action): string {
+  private getTriggerLabel(trigger: Trigger, action: KeyboardShortcutBinding): string {
```
RasmusAlbrektsen: Having the parameter `action` here seems confusing, since the type is a binding.

RasmusAlbrektsen: If we want to change it, remember to do so in the methods on line 201 and 245 too.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -17,10 +17,11 @@ import { DialogColorScheme, DialogSeverity } from 'src/app/shared/components/con
 import { EditKeyboardMappingResponse } from '../edit-keyboard-mapping/edit-keyboard-mapping.component'
 import { PartActionType } from '../../../../shared/models/action-type'
 import { TriggerType } from '../../../../shared/enums/trigger-type'
+import { KeyboardShortcutBinding, KeyboardShortcutBindingType } from '../../../../shared/models/keyboard-shortcut-bindings'
 
 export interface KeyboardMapping {
```
RasmusAlbrektsen, line 22: Does this differ from the information that we already have on a `KeyboardShortcutBinding`? 
I can't figure out if we even need this interface with the new implementation? 
But in any case it should't be done in this PR, in my opinion.
What do you think @vushu @LindvedKrvang?

src/app/settings/shelf-settings/components/edit-static-buttons/edit-static-buttons.component.ts
RasmusAlbrektsen, line 1: This is where the naming can get confusing too, we do not use the keyboard here at all, but we handle types that say we do. The naming in here all refers to `KeyboardShortcutBinding` but we only handle them as the actions they perform and nothing else. They are used for either touch or clicks at the moment.

Calling them `shortcuts` i find easier to understand. since it is a shortcut we can apply somewhere and trigger.

I would like your opinions @vushu @LindvedKrvang.

RasmusAlbrektsen: This also goes for the following:
`edit-static-buttons.component.html`
`edit-static-buttons.component.scss`
`static-buttons-configuration-card.component.html`
`sstatic-buttons-configuration-card.component.ts`

src/locale/messages.da.xlf
```diff
@@ -1016,10 +1016,7 @@
       </trans-unit>
       <trans-unit id="6475995402262064242" datatype="html">
         <source>panel-layout-configuration.import.info-message</source>
-        <target state="new">
-            Panel layout konfigurationer er afhængige af, at de rigtige Panel input modifikatorer og Makroer er tilgængelige i systemet.
-            Sørg for at Modifikatorne og Makroerne er importeret før importering af Layoutsne.
-        </target>
+        <target state="new"> Panel layout konfigurationer er afhængige af, at de rigtige Panel input modifikatorer og Makroer er tilgængelige i systemet. Sørg for at Modifikatorne og Makroerne er importeret før importering af Layoutsne. </target>
```
RasmusAlbrektsen: `Layoutsne` should be `Layouts`

RasmusAlbrektsen (commented):
I have some thing i just want to get an opinion on otherwise it looks good :-)

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
vushu: proably something went wrong when I did the cherry-pick

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
vushu: I will change it to shortcut

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vushu: I don't understand your question :d

src/app/settings/shelf-settings/components/edit-static-buttons/edit-static-buttons.component.ts
vushu: I see, you are right this is a common type that holds action and macro, so calling it without  keyboard would be more appropriate. I will rename them to ShortcutBinding.

vushu: While implementing this, I found out that it really touched many components, so it wasn't as trivial as I hoped it was :d

src/app/shared/models/tv2-action.ts
LindvedKrvang: A Macro and an Action **currently** have two properties in common.
What about when we add the third or fourth type of binding that both has nothing in common with either Macro or Action?
Then all the places where you simply have done `binding.binding.id` no longer compiles thus violating the Open/Closed principle.

src/app/producer-shelf/components/action-panels/action-panels.component.ts
```diff
@@ -155,11 +155,21 @@ export class ActionPanelsComponent implements OnInit, OnChanges, OnDestroy {
       .map(actionPanel => {
         return {
           name: actionPanel.name,
-          actions: this.actions.filter(action => !!action.rundownId && actionPanel.actionFilter.includes(action.metadata.contentType)).sort((a, b) => a.rank - b.rank),
+          actions: this.sortActionsByRankWithRespectToShelfConfiguration(actionPanel),
         }
       })
   }
 
+  private sortActionsByRankWithRespectToShelfConfiguration(actionPanel: ShelfActionPanelConfiguration): Tv2Action[] {
```
LindvedKrvang: You could simply let the name be `sortActionsByRank` and then let the parameter "speak" the `WithRespectToShelfConfiguration`. Then you can remove a lot of clutter from the name of this method.

src/app/rundown-view/services/mini-shelf-cycle.service.ts
```diff
@@ -39,7 +39,7 @@ export class MiniShelfCycleService {
     if (!videoClipFilename) {
       throw new Error('Unable to extract video clip filename.')
     }
-    const videoClipAction: Action | undefined = rundownActions.find(
+    const videoClipAction: Tv2Action | undefined = rundownActions.find(
```
LindvedKrvang: Why change the type from `Action` to `Tv2Action`?

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
LindvedKrvang: That was part of this task

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -81,20 +82,33 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
...
         trigger,
-        action: this.actions.find(action => action.id === this.getTriggerTypeId(trigger)) ?? this.createUnknownAction(),
+        shortcut: this.findTv2ActionPart(trigger),
       } as KeyboardMapping
     })
   }
 
-  private createUnknownAction(): Tv2Action {
-    return {
+  private findTv2ActionPart(trigger: Trigger): ShortcutBinding {
```
LindvedKrvang: This method is called `findTv2ActionPart` indicating that it will be returning a `Part`. Instead it returns a `ShorcutBinding`.

src/app/settings/shelf-settings/components/edit-static-buttons/edit-static-buttons.component.ts
LindvedKrvang: But the KeyboardShortcutBindings aren't just "shortcuts", they are "keyboardShortcuts". 

The real question here is: Why are we trying to use types which purpose is to bind to keyboard shortcuts in a place that don't care about nor uses keybindings?

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
RasmusAlbrektsen: We can disregard this for now :-)

src/app/shared/components/action-selector/action-selector.component.scss
```diff
@@ -8,6 +8,6 @@
   overflow-x: hidden;
 }
 
-sofie-action-box {
```
LindvedKrvang, line 11: We are in the ActionSeletor. This doesn't know anything about any keyboard shortcuts. Nor should it. It should only concern itself about selecting an Action. 
Case and point: We use also use this component when selecting an Action for the Static Actions which has nothing to do with keyboards shortcuts.

src/app/shared/components/action-selector/action-selector.component.ts
```diff
@@ -2,18 +2,19 @@ import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core'
...
 import { Tv2Action } from '../../models/tv2-action'
+import { ShortcutBinding, ShortcutBindingType } from '../../models/shortcut-bindings'
 
 @Component({
   selector: 'sofie-action-selector',
   templateUrl: './action-selector.component.html',
   styleUrls: ['./action-selector.component.scss'],
 })
 export class ActionSelectorComponent implements OnInit {
-  @Input() public selectedAction?: Tv2Action
```
LindvedKrvang, line 12: This goes for the entire file.
We are in the ActionSeletor. This doesn't know anything about any keyboard shortcuts. Nor should it. It should only concern itself about selecting an Action.
Case and point: We use also use this component when selecting an Action for the Static Actions which has nothing to do with keyboards shortcuts.

src/app/shared/components/macro-selector/macro-selector.component.scss
```diff
@@ -0,0 +1,13 @@
...
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr 1fr 1fr;
+  grid-template-rows: min-content min-content min-content min-content min-content;
+  width: 680px;
+  height: 650px;
+  overflow-y: auto;
+  overflow-x: hidden;
+}
+
+sofie-keyboard-shortcut-box {
```
LindvedKrvang, line 11: We are in the MacroSelector, this should know nothing about keyboard shortcuts. Just like the ActionSelector.

LindvedKrvang (changes requested):
I haven't been through it all yet, but there is starting to creep a lot of not-so-good stuff and complication in "in the corners". 
Stuff like renaming the `KeyboardShortcutBinding` to just `ShortcutBinding` to try and make it fit being used in a place where it shouldn't have been in the first place.

We should have a chat about it and possible try and pair programme an implementation.

src/app/shared/components/action-selector/action-selector.component.scss
vushu: that was how it was implemented in the begining, I don't want to change this component as that was how it was implemented.

src/app/shared/components/action-selector/action-selector.component.ts
vushu: Hence the renaming of to ShortcutBinding as @RasmusAlbrektsen pointed out, the problem is that it must know about that type since, the parent component has the action selector and the macro selector as child components. The parent uses The common type ShortcutBinding to pass into the child component. 
So the child components needs  to use this common type in it's public interface. Unless you somehow in the html can cast the common type ShortcutBinding to an Action.

src/app/shared/models/tv2-action.ts
vushu: I don't agree,  taking the asumption that we have more types in the future will often lead to complex code. I would rather we do the change when the change needs to happen, KISS.

src/app/shared/components/macro-selector/macro-selector.component.scss
vushu: that was how it was implemented in the begining, I don't want to change this component as that was how it was implemented action-selector. I wan't to keep this component similar.

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
vushu: I can change it to find findShortcutBinding

vushu:
Frankly this is perfectly fine to merge, so I will keep it open, ofcourse im open to critique and problems that will break the code or things that will cause performance issues. I believe theres a lot of opionated ways to tackle this and it gets very nitpicky now.

This is a good and valid solution and not a wrong one, I stand by my words.

vushu:
I will close this branch since I don't want to keep it hanging and rot.

PR #368, opened 2025-04-11
SOF-2558 - Keyboard shows duplicated content placeholder actions

RasmusAlbrektsen:
Due to the way we updated our array of actions we saw duplicated actions that would interfere with the way we gathered graphic actions from our `OnAirSegment` based off of rank.


vushu (approved):
lgtm

LindvedKrvang (approved):

PR #371, opened 2025-04-29
Sof 2519/able to configure keyboard mapping with macro

LindvedKrvang:
Keyboard mappings are now able to select Macros as well.

Note: I opted to use the `sofie-action-picker` since this encapsulate the logic of ActionArguments, so we don't need to worry about it at all in the `edit-keyboard-mapping.component` (we only need to provide a form control for it).

This PR also show the selected Macro in the KeyboardMappings overview and in the virtual keyboard of the RundownView.


src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -52,32 +52,31 @@
...
+      <sofie-button-group class="action-macro-selector-button-group">
+        <sofie-button (click)="switchToSelectAction()" [type]="keyboardMappingTypeBeingSelected === KeyboardMappingType.ACTION ? 'PRIMARY' : 'STANDARD'">Select Action</sofie-button>
+        <sofie-button (click)="switchToSelectMacro()" [type]="keyboardMappingTypeBeingSelected === KeyboardMappingType.MACRO ? 'PRIMARY' : 'STANDARD'">Select Macro</sofie-button>
+      </sofie-button-group>
+
+      <ng-container [ngSwitch]="keyboardMappingTypeBeingSelected">
+        <sofie-action-picker
+            *ngSwitchCase="KeyboardMappingType.ACTION"
+            label="Select Action"
+            placeholder="No Action selected"
```
vushu: Should this be translated?

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
```diff
@@ -52,32 +52,31 @@
...
-          <sofie-number-input
-            *ngSwitchCase="ActionArgumentSchemaType.NUMBER"
-            [label]="selectedAction?.argument?.name!"
-            [formControlName]="'actionArguments'"
-            [helpText]="selectedAction?.argument?.description!"
+        />
+        <sofie-macro-picker
+            *ngSwitchCase="KeyboardMappingType.MACRO"
+            label="Select Macro"
+            placeholder="No Macro selected"
```
vushu: Also here translation?

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -86,10 +92,22 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
...
+            type: KeyboardMappingType.ACTION,
+            trigger,
+            action: this.actions.find(action => action.id === trigger.actionId) ?? this.createUnknownAction(),
+          } as KeyboardMapping
+        }
+        case TriggerType.MACRO: {
+          return {
+            type: KeyboardMappingType.MACRO,
+            trigger,
+            macro: this.macros.find(macro => macro.id === trigger.macroId) ?? this.createUnknownAction(),
```
vushu: shouldn't this be called `createUnknownMacro`?

vushu (approved):
Cool :), my only concern is we are no longer using the old look and feel, also we don't display the description any longer, should we then just remove the description field since it isn't used?, is that something we are ok with I personally don't mind as long as stakeholders are ok with this change, it's just discussion.

vushu:
Minor bug:
When adding custom label it doesn't get shown on the list, but the macro name.

Screenshot 2025-04-30 at 09 48 16

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
```diff
@@ -172,37 +194,46 @@ export class TriggerProducerKeyBindingService implements KeyBindingService {
   }
 
   private updateTriggerKeybindings(): void {
-    this.triggerKeyBindings = Array.from(this.triggersWithAction, ([trigger, action]) => this.createBinding(action, trigger, this.rundown!.id))
+    const actionTriggerKeybindings: KeyBinding[] = Array.from(this.triggersWithAction, ([trigger, action]) => this.createActionTriggerKeyBinding(action, trigger, this.rundown!.id))
```
RasmusAlbrektsen, line 197: This is no really related to your changes, but still a bit:
Could we not just put a guard on the rundown ID and avoid this assertion on the `this.rundown!.id`?

RasmusAlbrektsen (approved):
I've approved because this is a viable solution and it looks good. My only concern is taking away something that seemingly the users liked, which is the `action-selector` that had more visual appeal, though it was rather large and the Split Screen inputs cluttered it. 

I'd love an alternative solution where we possibly just open that up in a dialog instead, much like an option to open up the entire file system when looking for a file. In any case we just lose the ease of identifying elements by e.g. their color or by being able to separate things into tabs that you can look through in the future (Actions, Macros, Split Screen inputs etc.).
Also, by us not highlighting the search field in the dropdown we are essentially keeping the amount of clicks needed (when comparing to the selector) but making it visually less appealing still, which seems like a downgrade.

But ultimately i can't know if this is important enough, so we can always make that into a new task.

vushu (approved):

vushu:
I approve since its a valid solution and i think we should move on, but there are some UI/UX concerns as @RasmusAlbrektsen mentioned that also plays a role, maybe this needs be discussed by some people with good design eyes?

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
LindvedKrvang: They should. Translations have been added

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.html
LindvedKrvang: They should. Translations have been added

src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
LindvedKrvang: It should. Good catch.

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
LindvedKrvang: We are inside an arrow function, so we lose the context of the method.
This means we would need to do the guard check in both functions. (Which may well be better than the non-null assertions)

PR #372, opened 2025-05-01
SOF-2311 - Spanned piece copies no longer make use of the start property from original piece

RasmusAlbrektsen:
This PR handles the fact that whenever we made copies of a spanning piece (to span for the subsequent parts in a segment) we just pushed a new piece that also included the start time for the original part. We now push the piece without this property carried over to ensure that we span all of the subsequent parts.


vushu (approved):
Nice!

LindvedKrvang (approved):

PR #373, opened 2025-05-02
SOF-2570 Use PlayoutContent instead of PieceType

LindvedKrvang:
Note: Breaking change. Needs to be merged with:
- https://github.com/tv2/sofie-blueprints-inews/pull/261
- https://github.com/tv2/alba-server/pull/343

Introduce PlayoutContent.
PieceType has been replaced by PlayoutContent.


src/app/core/models/tv2-piece.ts
```diff
@@ -1,14 +1,16 @@
 import { Piece } from './piece'
-import { Tv2PieceType } from '../enums/tv2-piece-type'
 import { Tv2OutputLayer } from './tv2-output-layer'
 import { Tv2AudioMode } from '../enums/tv2-audio-mode'
+import { PlayoutContentType } from '../enums/playout-content-type'
 
 export interface Tv2Piece extends Piece {
```
RasmusAlbrektsen, line 6: We should really get rid of the Tv2Piece soon, we could really eliminate some hoops we jump through here and there. Not for this PR, just a note.

src/app/core/parsers/zod-entity-parser.service.ts
```diff
@@ -62,7 +62,9 @@ export class ZodEntityParser implements EntityParser {
     lifespan: zod.nativeEnum(PieceLifespan),
     // TODO: Should this be less TV2 specific.
```
RasmusAlbrektsen, line 63: Could probably remove this comment.

RasmusAlbrektsen (approved):
Good job!

vushu (approved):

PR #376, opened 2025-05-15
SOF-2580 - Hover Scrub for Video Clip Actions

RasmusAlbrektsen:
Our hover scrub no longer worked for actions in our action panels, whether they existed on the producer shelf or the standalone shelf. We want them to exist for only desktop-devices, however, since we cannot distinguish our currently used tablet as "not a desktop"-device, we need to look at the URL to ensure we are using the Standalone shelf. That is what this fix accomplishes.


vushu (approved):
lgtm

LindvedKrvang (approved):

PR #378, opened 2025-05-22
SOF 2582 - Keyboard mappings dont show colors on the settings page

RasmusAlbrektsen:


vushu (approved):
LGTM

src/scss/_sofie-color-vars.scss
```diff
@@ -39,6 +39,7 @@ body {
   --tv2-video-clip-color: #1769ff;
   --tv2-manus-color: #370020;
   --tv2-video-missing-source-gradient-color: #9c2720;
+  --tv2-macro-color: #8a5a45;
```
LindvedKrvang: We should really stop naming colors for where we use them and start naming them for what color they are

LindvedKrvang (approved):

src/scss/_sofie-color-vars.scss
RasmusAlbrektsen: Absolutely right! An oversight on my part, i wanted to make the color and then the macro-reference so it could be easily used in several places! My bad :-)

KvelaGorrrrnio (approved):
Your best work so far 🚀

PR #379, opened 2025-05-30
SOF-2192 Replace references of 'sofie' with 'alba'

LindvedKrvang:
Big PR incoming.
Most of it is the change of the `sofie` prefix to `alba` on all our components.

Note: We have componets/translations that refer to settings in old Sofie. These are still called Sofie since they are referring to Sofie.


RasmusAlbrektsen (approved):

.github/workflows/create-merge-pull-requests.yml
```diff
@@ -13,8 +13,8 @@ jobs:
         - uses: actions/checkout@v3
         - name: Set Git config
           run: |
-            git config --local user.email "Sofie-Automation@users.noreply.github.com"
-            git config --local user.name "GitHub Actions # Sofie-Automation"
+            git config --local user.email "Alba-Automation@users.noreply.github.com"
```
KvelaGorrrrnio, line 16: Shouldn't this be uniform across repos? In alba-tv2-server it is simply `alba@...`. https://github.com/tv2/alba-tv2-server/pull/126#discussion_r2123471439

Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN yarn build
 FROM nginx:1.27-alpine-slim
 WORKDIR /usr/share/nginx/html
 
-COPY --from=BUILD_PHASE /opt/alba-web-client/dist/ng-sofie .
+COPY --from=BUILD_PHASE /opt/alba-web-client/dist/ng-alba .
```
KvelaGorrrrnio, line 12: Unrelated to the PR changes: shouldn't the target be changed to the package name `alba-web-client` instead of `ng-alba`?

KvelaGorrrrnio (approved):

Dockerfile
LindvedKrvang: That would make sense to do :)

PR #383, opened 2025-06-13
SOF-2590 - Sofie header seen in Sofie Settings

RasmusAlbrektsen:
This removes the `dialog` prefix that was mistakenly added to the sofie settings scss-class.


vushu (approved):
Excellent

LindvedKrvang:
As mentioned on the refining, it would be nice if this tasks also removed the other places where the `dialog-` prefix was mistankenly added.

KvelaGorrrrnio (approved):
Same as Lindved. The change is good, but it before the task is closed, there should be looked into whether there are other places that are wrongly affected by the `dialog-` prefix 😄 I don't mind whether that is done in this PR or another.

RasmusAlbrektsen:
@LindvedKrvang @KvelaGorrrrnio I actually did look for this based on the commit in the task, i somehow messed up and didnt get the change for the only other place i could find it. I have changed it now, it was the invalid segment header.

LindvedKrvang (approved):

PR #384, opened 2025-06-19
Sof 2446/implemented sorting shortcuts correctly in keyboard mappings settings

vushu:
I have made an initial implementation of this feature,
I not quite a fan of having an `excludedKeyboardSymbols` to then filter out from found entries via the keyboard api,
alternatively we should just hardcode the single keys.
 
 suggestions are welcome.

Here it is in action.

https://github.com/user-attachments/assets/526d84b2-8dd7-4744-9391-08b9791880b9


src/app/settings/keyboard-mapping-settings/components/keyboard-mapping-settings-page/keyboard-mapping-settings-page.component.ts
```diff
@@ -35,6 +35,7 @@ export class KeyboardMappingSettingsPageComponent implements OnInit, OnDestroy {
   public keyboardMappings: KeyboardMapping[]
 
   public selectedKeyboardMappings: Set<KeyboardMapping> = new Set()
+  public excludedKeyboardSymbols = ['-', '+', '!', '@', '#', '$', '%', '^', '&', '*', '(', ')', '_', '=', '{', '}', '[', ']', ':', ';', '"', "'", '<', '>', '?', '/', '§', '~', '`', ',','.']
```
LindvedKrvang: I'm no Regex guru like Anders, but this seems possible to replace with a Regex :)

src/app/shared/pipes/format-keyboard-keys.pipe.ts
```diff
@@ -13,8 +13,6 @@ export class FormatKeyboardKeysPipe implements PipeTransform {
       return key.replace(/Key/g, '')
     } else if (/Digit/g.test(key)) {
       return `${key.replace(/Digit/g, '')} (Digit)`
-    } else if (/Numpad/g.test(key)) {
-      return `${key.replace(/Numpad/g, '')} (Numpad)`
```
LindvedKrvang: I can't exactly remember why, but I am 85% sure that we deliberately make a distinction betwen numpad numbers and "other" numbers, meaning we should keep this statement.

LindvedKrvang (commented):
What's the motive behind fetching the actual keyboard layout of whatever machine is used by the client?

src/app/shared/pipes/format-keyboard-keys.pipe.ts
vushu: I removed it since it looks like we sort wrongly eg.
If we keep the format the keyboard mapping ui will look like

```
n
add (numpad)
n + shift
.. etc.
```

It is sorted this way since we sort by the key which in this case numpadAdd, hence is why i looks like that is it sorting wrongly.

So that why I removed it, but yeah we need to talk about this.

KvelaGorrrrnio: This pipe should not be used when sorting as it is for converting keys into their display format :D
Instead we should consider adding a new class that compare keyboard keys, where it internally handles key groupings and their individual ranking. E.g.:
```typescript
class KeyboardKeyComparer {
  public compare(key1: string, key2: string) {
    return this.getKeyRank(key1) - this.getKeyRank(key2)
  }

  private getKeyRank(key: string): number {
     // TODO: Implement getting rank for each key grouping type (e.g alphabetic, numeric, special characters, function keys, ...)
    return Infinity
  }
}

```

vushu: But I am not using the pipe to sort :D, let's talk possible implementations? We also need to keep the scope to be within the bug task.

vushu:
The concern is that if the client is using a different OS like macOS which uses another key for eg. CMD, by using the keyboard api we ensure to support this keyboard layout. There are proably other good solutions so any other ideas are welcome :)

LindvedKrvang:
> The concern is that if the client is using a different OS like macOS which uses another key for eg. CMD, by using the keyboard api we ensure to support this keyboard layout. There are proably other good solutions so any other ideas are welcome :)

Maybe try taking a look at `tv2-keyboard-key-label.service`. Perhaps you can utilize it? (However, I admit I haven't taken a close look at it compared to your use-case)

vushu:
Hi I have made some changes after talking with @KvelaGorrrrnio:

1. keys are now grouped by a given rank which is defined in `KeyboardComparerService`, (Not sure if I even need the keyboard api in the service so I haven't used it)
2. Updated the pipe to show the actual key symbol I use the keyboard api to get that. (What do you guys think?)

vushu:
Here is a video of the result (don't mind the lowercaseness I have added uppercase back)


https://github.com/user-attachments/assets/318422c3-a78c-4630-be68-d3326ef8be78

RasmusAlbrektsen (approved):

src/app/shared/pipes/format-keyboard-keys.pipe.ts
```diff
@@ -1,21 +1,28 @@
...
     return keys.map(key => this.formatKey(key)).join(' + ')
   }
 
   private formatKey(key: string): string {
     if (/Key/g.test(key)) {
       return key.replace(/Key/g, '')
-    } else if (/Digit/g.test(key)) {
+    }
+
+    if (/Digit/g.test(key)) {
```
KvelaGorrrrnio, line 18: Nice use of guards instead of `else-if` :+1:

src/app/shared/services/keyboard-key-comparer.service.ts
```diff
@@ -0,0 +1,118 @@
+import { Injectable } from '@angular/core'
+
+enum KeyboardKeyRank {
```
KvelaGorrrrnio: As these are not ranks for individual keys, I suggest that this is called `KeyboardKeyGroupRank`.

src/app/shared/services/keyboard-key.service.ts
```diff
@@ -0,0 +1,6 @@
+import { Injectable } from '@angular/core'
+
+@Injectable()
+export abstract class KeyboardKeyService {
+  public abstract getActualPhysicalKey(keyCode: string): string
```
KvelaGorrrrnio: `Actual` can be removed as it does not add any semantic value to the method name :)

src/app/shared/services/api-keyboard-key.service.ts
```diff
@@ -0,0 +1,24 @@
...
+@Injectable()
+export class ApiKeyboardKeyService implements KeyboardKeyService {
+  private keyboardEntries: KeyboardLayoutMap = new Map<string, string>()
+  constructor(logger: Logger) {
+    this.getKeyboardEntries()
+      .then(() => {})
+      .catch(error => {
+        logger.error('Failed to load keyboard entries:', error)
+      })
+  }
```
KvelaGorrrrnio: As you probably are aware, this may lead to unmapped keys if the `getActualPhysicalKey` is called before the promise is resolved :) The "best" solution I have at the moment is to make the `getActualPhysicalKey` method async as to wait for the resolution this.

src/app/shared/services/api-keyboard-key.service.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Injectable } from '@angular/core'
+@Injectable()
+export class ApiKeyboardKeyService implements KeyboardKeyService {
+  private keyboardEntries: KeyboardLayoutMap = new Map<string, string>()
+  constructor(logger: Logger) {
+    this.getKeyboardEntries()
+      .then(() => {})
+      .catch(error => {
+        logger.error('Failed to load keyboard entries:', error)
+      })
```
KvelaGorrrrnio: The `.then(..)` call does nothing and can be removed.
And the `logger.error(...)` is misused a tiny bit. I suggest that you use the `logger.data(error)` for collecting the error. This way, the error should be printed in an interactive way instead of a string.
```typescript
    this.getKeyboardEntries()
      .catch(error => logger.data(error).error('Failed to load keyboard entries. Using empty key mapping table.'))
```

src/app/shared/services/keyboard-key-comparer.service.ts
```diff
@@ -0,0 +1,118 @@
...
+type GetRankFunction = (key: string) => number | undefined
+
+function getDigitKeyRank(key: string): number | undefined {
+  if (!/^digit/i.test(key)) {
+    return
+  }
+
+  const number = Number(key.replace(/^digit/i, ''))
+  return KeyboardKeyRank.DIGIT + number
+}
```
KvelaGorrrrnio: Just a suggestion: All of these `getXRank` can be moved in to the class. When first looking at the file it is hard to see that there is a class in here :)

You can e.g. in the constructor or directly in the member declaration define these:
```typescript
class KeyboardKeyComparer {
  private readonly getRankFunctions: readonly GetRankFunction[] = [
    function getDigitKeyRank(key: string)): number | undefined {
      // ...
    },
    function getNumpadDigitKeyRank(key: string)): number | undefined {
      // ...
    },
    /...
  ]
}

File: src/app/shared/services/keyboard-key-comparer.service.ts [hunk 9]
```diff
@@ -0,0 +1,118 @@
...
+  const code = key.charCodeAt(0)
+  return KeyboardKeyRank.ALT + code
+}
+
+function getOtherKeyRank(key: string): number {
+  return KeyboardKeyRank.OTHER + key.charCodeAt(0)
+}
+
+@Injectable()
+export class KeyboardKeyComparerService {
```
KvelaGorrrrnio: It would be nice if the `Service` part of the name is removed (the file suffix `.service.ts` should still be there) as it doesn't add any semantic value :) I think that Comparer describes the responsibility of the class clearly enough.

src/app/shared/services/api-keyboard-key.service.ts
```diff
@@ -0,0 +1,24 @@
+import { Logger } from 'src/app/core/abstractions/logger.service'
+import { KeyboardKeyService } from './keyboard-key.service'
+import { Injectable } from '@angular/core'
+@Injectable()
```
LindvedKrvang: Would be nice with a newline between the imports and the rest of the content in the file

KvelaGorrrrnio (approved):
Nice work! :D I have added a few comments on structure, naming and a single concern for a potential race-condition :)

src/app/shared/services/keyboard-key-comparer.service.ts
```diff
@@ -0,0 +1,118 @@
...
+  if (!/^alt/i.test(key)) {
+    return
+  }
+  const code = key.charCodeAt(0)
+  return KeyboardKeyRank.ALT + code
+}
+
+function getOtherKeyRank(key: string): number {
+  return KeyboardKeyRank.OTHER + key.charCodeAt(0)
+}
```
LindvedKrvang: I think I kinda see why the other functions above this one are private static functions. 
However, there is no reason for this specific function to not be a private method of `KeyboardKeyCompareService`.

src/app/shared/services/keyboard-key-comparer.service.ts
```diff
@@ -0,0 +1,118 @@
...
+        return diff
+      }
+    }
+    return aKeys.length - bKeys.length
+  }
+
+  private getKeyRank(key: string): number {
+    for (const rankFunction of this.keyboardRankFunctions) {
+      const rank = rankFunction(key)
+      if (rank !== undefined) {
```
LindvedKrvang: If you want, then you could also use this syntax: `!!rank`.

LindvedKrvang (approved):

src/app/shared/services/keyboard-key-comparer.service.ts
vushu: Will move it to the service :)

src/app/shared/services/keyboard-key-comparer.service.ts
vushu: thanks.

src/app/shared/services/keyboard-key-comparer.service.ts
vushu: Cool, I wasn't sure if this was ok :)

src/app/shared/services/api-keyboard-key.service.ts
vushu: Ah cool thanks, I also wonder what do in this scenario :) Then I guess I will need to change the return value to be a `Promise<string>` :d

src/app/shared/services/api-keyboard-key.service.ts
vushu: alright :)

src/app/shared/services/keyboard-key-comparer.service.ts
vushu: Ah cool :)

src/app/shared/services/api-keyboard-key.service.ts
vushu: @KvelaGorrrrnio  since this is going to affect many places, maybe we could put this as a kaizen thing and frankly I am getting lost in how to do this :D

src/app/shared/pipes/format-keyboard-keys.pipe.ts
vushu: thanks :)

src/app/shared/services/api-keyboard-key.service.ts
vushu: Kaizen task: https://tv2cms.atlassian.net/browse/AG-501

PR #385, opened 2025-07-01
Ag 392/create ingest settings page

LindvedKrvang:
Able to create, edit and delete iNews queues.

https://github.com/user-attachments/assets/d4b6dfb0-34c4-4713-ae53-9c7b8eed3a44

Also in this PR:
- Shows the connection status of the ingest-gateway on the Ingest settings page.


src/locale/messages.en-US.xlf
```diff
@@ -1042,6 +1042,70 @@
...
+        <source>edit-inews-ingest-queue-subscription.queue-id.invalid-message</source>
+        <target state="new">Queue name</target>
+      </trans-unit>
+      <trans-unit id="1755916578227085281" datatype="html">
+        <source>inews-ingest-configuration.card-title</source>
+        <target state="new">iNews ingest configuration</target>
+      </trans-unit>
+      <trans-unit id="1423690841692835990" datatype="html">
+        <source>inews-ingest-configuration.create-button.text</source>
+        <target state="new">Create iNews queue</target>
```
vushu: The queue is already created in iNews right? Wouldn't it make sense to call it something like load or ingest instead of Create? :)

vushu (approved):
Proably it's just me, but I do get confused by angular forms, anyways it looks fine to me.

LindvedKrvang:
> Proably it's just me, but I do get confused by angular forms, anyways it looks fine to me.

@vushu Try take a look at this then: https://angular.dev/guide/forms/reactive-forms :)

src/locale/messages.en-US.xlf
KvelaGorrrrnio: An alternative could also be `Register iNews Queue`. Or `Add iNews Queue` - even though that resembles create a bit.

RasmusAlbrektsen: Or just `Add queue`, i don't see the need to specify iNews in this context. Also a blue-color on the button could possibly better indicate that action.

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.html
```diff
@@ -0,0 +1,16 @@
+<form [formGroup]="form" class="mb-3">
+  <alba-text-input
+      [formControlName]="'queueId'"
+      i18n-label label="edit-inews-ingest-queue-subscription.queue-id.label"
+      i18n-placeholder placeholder="edit-inews-ingest-queue-subscription.queue-id.placeholder"
+      [isRequired]="true"
+  />
+  <alba-checkbox [formControlName]="'isDisabled'" />
```
KvelaGorrrrnio, line 8: I suggest that the dynamic binding is removed as this is just a static value. It can be written as `formControlName="isDisabled"`. By doing so we remove extraneous syntax so there are less layers to understand. It is harder to distinguish dynamic bindings of variables with static values. A counter-example is that we don't write element classes as `[class]="'mb-3'"`.

The suggestion can be applied in other places as well :)

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.saveButtonTooltip = this.getErrorMessagesForForm()
+  }
+
+  private getErrorMessagesForForm(): string | undefined {
+    const invalidControlsLabels: string[] = this.findInvalidControlsLabels()
+    if (invalidControlsLabels.length > 0) {
+      return $localize`common.form-required-field.button-tooltip` + ` ${invalidControlsLabels.join(', ')}`
+    } else {
+      return undefined
+    }
```
KvelaGorrrrnio: The else branch can be dissolved - either by removing it entirely or by reducing it to `return undefined` in the root scope of the method.

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  public onCancel: EventEmitter<void> = new EventEmitter()
+
+  public form: FormGroup
+  public saveButtonTooltip: string | undefined
+
+  constructor(public readonly formBuilder: NonNullableFormBuilder) {}
+
+  public ngOnInit(): void {
+    this.form = this.formBuilder.group({
+      queueId: this.formBuilder.control(this.inewsIngestQueueSubscription?.queueId ?? '', Validators.required),
```
KvelaGorrrrnio: Currently `QUEUE_ID_CONTROL_NAME` is only used in `translateControl(...)`, but is not used otherwise. I assume that the value in the constant refers to this key. If that is the case, then I suggest that we change this to use the constant: `[QUEUE_ID_CONTROL_NAME]: this.formBuilder.control(...),`

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.ts
KvelaGorrrrnio, line 1: Just jotting down my thoughts of the edit/create sidebar.
The experience is a bit cramped.
The input field is not wide enough to contain a typical iNews Queue name.

It would also be nice with a bit more spacing between the controls in the form.

And having a help (?) icon with a tooltip description of what disable means in this context would be preferable.

image

src/locale/messages.da.xlf
```diff
@@ -1042,6 +1042,90 @@
...
+        <source>inews-ingest-settings-configuration.ingest-gateway-connection.not-connected</source>
+        <target state="new">Ikke forbundet</target>
+      </trans-unit>
+      <trans-unit id="286698503445217988" datatype="html">
+        <source>inews-ingest-settings-configuration.ingest-health-status.good.tooltip</source>
+        <target state="new">Ingest-gateway'en er forbundet</target>
+      </trans-unit>
+      <trans-unit id="4815332209398388460" datatype="html">
+        <source>inews-ingest-settings-configuration.ingest-health-status.bad.tooltip</source>
+        <target state="new">Tjeck ingest-gateway'en for at fejlsøge</target>
```
KvelaGorrrrnio: `ingest-gateway` sounds like a referential value to some code construct or enumeration. We could use `Ingest Gateway` as the user facing term for this.

Since this is a sentence, it would make sense to end it with a period.

And `Tjeck` should be `Tjek` :-)

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.scss
```diff
@@ -0,0 +1,3 @@
+.td-width {
+  width: 250px;
+}
```
KvelaGorrrrnio: What is the intention with this styling? I have tried changing and removing it, but I haven't noticed anything different :)
If it is for setting minimum width, then `min-width` can be used instead.

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.html
KvelaGorrrrnio, line 1: The column for whether the queue subscriptions are enabled or disabled shows the information in a technical manner.
image

If the column should still be kept as "Is disabled" I suggest that we use "Yes"/"No" as the column value.

An alternative could be to append the disabled status to the queue name as shown in the image below:
image

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.html
```diff
@@ -0,0 +1,33 @@
+<alba-card>
+  <ng-container cardHeader>
+    <span i18n>inews-ingest-configuration.card-title</span>
+    <alba-spacer />
+    <alba-pill class="me-3" [text]="ingestGatewayStatusText" [tooltip]="ingestGatewayStatusTooltip" [color]="ingestGatewayStatusColor" />
```
KvelaGorrrrnio: I am bit uncertain on whether the overall ingest health status should be placed under the iNews specific portion of the settings page or if it should have its own dedicated card on the page (preferably at the top).

By moving it out to a separate card, we also get more space for listing the messages associated with the status.

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  public get ingestGatewayStatusText(): string {
+    switch (this.ingestHealthStatus) {
+      case IngestHealthStatus.GOOD: {
+        return $localize`inews-ingest-settings-configuration.ingest-gateway-connection.connected`
+      }
+      default: {
+        return $localize`inews-ingest-settings-configuration.ingest-gateway-connection.not-connected`
+      }
+    }
```
KvelaGorrrrnio, line 126: I just want to point out that only differentiating between good and bad/connected and disconnected doesn't model the health state that we are trying to capture. There is both the connected/disconnected to the iNews Gateway and then there is if the iNews gateway is in a bad state - e.g. not having a connection to the iNews server.
I don't know if this is planned to incorporated when we look at data event ingestion or if it should be incorporated now?

src/app/shared/enums/color.ts
```diff
@@ -0,0 +1,7 @@
+export enum Color {
+  SUCCESS = '#00ac17',
+  WARNING = '#ffc302',
+  DANGER = '#ef0000',
+  WHITE = 'white',
+  UNKNOWN = 'grey',
+}
```
KvelaGorrrrnio: These values could also be defined as css variables and referenced here via `var()`. This way we keep the styling information centralized and make it available for CSS classes to use as well.

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.html
KvelaGorrrrnio, line 1: Would the card loose semantic value if we removed the "ingest" from the card title, so it becomes "iNews configuration"?

src/locale/messages.en-US.xlf
```diff
@@ -1042,6 +1042,90 @@
...
+        <source>edit-inews-ingest-queue-subscription-dialog.title.edit</source>
+        <target state="new">Edit iNews queue</target>
+      </trans-unit>
+      <trans-unit id="4151200930341599416" datatype="html">
+        <source>edit-inews-ingest-queue-subscription.queue-id.label</source>
+        <target state="new">Queue name</target>
+      </trans-unit>
+      <trans-unit id="7854902291272113067" datatype="html">
+        <source>edit-inews-ingest-queue-subscription.queue-id.placeholder</source>
+        <target state="new">Write the name of the queue in iNews here</target>
```
KvelaGorrrrnio: The formulation can be a bit clearer by writing it as `Write the name of the iNews queue here`.
Also this is a sentence, so it would be nice if it ends with a period.
An alternative for the placeholder could be an example - e.g. `SOME.TEST.RUNDOWN`

src/locale/messages.en-US.xlf
```diff
@@ -1042,6 +1042,90 @@
...
+        <source>inews-ingest-settings-configuration.ingest-health-status.good.tooltip</source>
+        <target state="new">The ingest-gateway is connected</target>
+      </trans-unit>
+      <trans-unit id="4815332209398388460" datatype="html">
+        <source>inews-ingest-settings-configuration.ingest-health-status.bad.tooltip</source>
+        <target state="new">Check the ingest-gateway to troubleshoot</target>
+      </trans-unit>
+      <trans-unit id="487016061336411097" datatype="html">
+        <source>inews-ingest-settings-configuration.ingest-health-status.unknown.tooltip</source>
+        <target state="new">The ingest-gateway is not connected</target>
```
KvelaGorrrrnio: The `state="new"` can be removed as the translations have been filled out. state="new" means that it has not been translated yet. It goes for the other entries as well.

KvelaGorrrrnio (commented):
Nice work :) I have some suggestions and questions :D

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.scss
LindvedKrvang: Not sure what you are changing, but it does have an affect for me.

https://github.com/user-attachments/assets/483e4c43-25f6-4055-b0ab-8d8fc5ce6124

src/locale/messages.en-US.xlf
LindvedKrvang: I gotta be honest: This is autogenerated and I haven't really looked at it except for changing the value :p

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.html
LindvedKrvang: I have noted that you have suggested this before. 
This seems to come down to personal preference and I think we just have to agree on disagreing ;)

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.ts
LindvedKrvang: I don't really know what we want to show the user about the connection status. Suggestions are welcome.

src/app/shared/enums/color.ts
LindvedKrvang: How would I access that from the Typescript?

src/locale/messages.en-US.xlf
LindvedKrvang: Updated

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.html
LindvedKrvang: Probably not. It's been updated

src/locale/messages.da.xlf
LindvedKrvang: Updated

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.ts
LindvedKrvang: Updated to `30rem`. This should be enough to write `SPORTEN.SPORTSKANAL.AFVIKLINGER.AFV-D.AFVD-ON-AIR`

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.ts
LindvedKrvang: Updated

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.ts
LindvedKrvang: Updated

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.html
LindvedKrvang: That sounds like a splendid idea for a task when we know how we want to display it ;)

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.html
LindvedKrvang: Updated to your last suggestion :)

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.ts
RasmusAlbrektsen: We could possibly have a different task for correctly displaying relevant health status for ingest? I

RasmusAlbrektsen (approved):
Good stuff!

src/locale/messages.da.xlf
```diff
@@ -1042,6 +1042,86 @@
...
+        <source>inews-ingest-configuration.table.queue-id.name</source>
+        <target state="new">Kø</target>
+      </trans-unit>
+      <trans-unit id="3441227564059145484" datatype="html">
+        <source>inews-ingest-configuration.delete-dialog.title</source>
+        <target state="new">Slet iNews kø</target>
+      </trans-unit>
+      <trans-unit id="927174191937804228" datatype="html">
+        <source>inews-ingest-configuration.delete-dialog.message</source>
+        <target state="new">Er du sikker på du vil slette denne kø?</target>
```
vushu: Suggestion:
Er du sikker på, at du vil slette denne kø?

vushu (approved):

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.html
KvelaGorrrrnio: Correct, I have suggested this before 😁 I do not see where the personal preference should come from 😄 I have stated reasons regarding reduced readability due to added lexical complexity and misaligned binding-intent. In addition, dynamic binding has a slightly higher performance overhead than static assignment as stated in [this issue](https://github.com/angular-eslint/angular-eslint/issues/1172#issuecomment-2311875481).
I do not expect that it is changed in this PR, but I will add it as a proposal for our style guide 📝

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.scss
```diff
@@ -0,0 +1,3 @@
+.container {
+  width: 30rem;
```
KvelaGorrrrnio, line 2: Suggestion: Adding `max-width: 100%` will ensure that we get a scalable component for smaller screens :)
Currently it will cut off the content and have horizontal scroll.

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.html
KvelaGorrrrnio: AG-533 mentions this.

src/app/settings/ingest-settings/components/inews-ingest-settings-configuration/inews-ingest-settings-configuration.component.ts
KvelaGorrrrnio: I have created AG-533 for this.

src/app/shared/enums/color.ts
KvelaGorrrrnio: Just use the CSS variable directly in the string value of the enum. E.g.:
```typescript
export enum Color {
  SUCCESS = 'var(--green-color)',
  WARNING = 'var(--attention-color)',
  DANGER = 'var(--danger-color)',
  WHITE = 'var(--white-color)',
  UNKNOWN = 'var(--dark-gray-color)',
}
```

src/locale/messages.en-US.xlf
KvelaGorrrrnio: I have removed the state="new" :)

src/app/settings/ingest-settings/components/edit-inews-ingest-queue-subscription/edit-inews-ingest-queue-subscription.component.scss
KvelaGorrrrnio: I have added `max-width: 100%`.

KvelaGorrrrnio:
It now looks like this:
image

PR #386, opened 2025-07-07
Fix: Icon button does not use user-defined colors.

KvelaGorrrrnio:
Previously setting the CSS `color` attribute would change the color of the icon, but due to default values, this no longer worked. To keep the default values and make it customizable, CSS variables are used with fallback values.

Prior to this change, attention banners looked like this:
image

After it looks like this:
image


vushu (approved):
👍

RasmusAlbrektsen (approved):

PR #388, opened 2025-07-09
AG-488: Off air rundown clock

KvelaGorrrrnio:
Adds animated off air rundown clock view with time information and system name:
image


vushu (approved):
Cool

RasmusAlbrektsen (approved):

PR #389, opened 2025-07-09
AG-487: Adds on air rundown clock

KvelaGorrrrnio:
> [!warning]
> Without https://github.com/tv2/sofie-blueprints-inews/pull/269 and https://github.com/tv2/alba-server/pull/378 you will get weird part names.

The PR adds the on air rundown clock view.
It uses the part name for displaying which part is on air and which is next. Due to improper names on parts from both old and new blueprints, this requires changes both in Sofie blueprints and Alba blueprints.

The part color is based on the first piece with a set playout content type that is not `UNKNOWN` - defaulting to `UNKNOWN`. 

image

The PR also changes the expected end time for untimed rundowns to be computed from the current time and the remaining duration of the rundown, instead of the entire expected duration of the rundown.


vushu (approved):

RasmusAlbrektsen (approved):
Love it!

PR #392, opened 2025-07-11
Refactor/ Removes the tv2piece

RasmusAlbrektsen:
This PR is just an attempt at the initial step to get rid of the `Tv2Piece`. 
It substitutes the use of the `Tv2Piece` to instead use the `Piece` that now has `PieceMetadata` rather than `Tv2PieceMetadata.` 

We no longer need to assume that our pieces are of the kind `Tv2Piece`, but for now we still assume that our metadata can be casted to the type `PieceMetadata`, which isn't pretty but is the next step in line.
We'd probably have to figure out whether we can always assume (or assure) that our pieces have metadata and thereby avoid this optional.


vushu (approved):

src/app/rundown-view/components/follow-playhead-timeline/follow-playhead-timeline.component.ts
```diff
@@ -166,7 +165,7 @@ export class FollowPlayheadTimelineComponent implements OnChanges {
 
   private doesPartContainVideoClipOrVoiceOver(part: Part): boolean {
     const supportedPlayoutContentTypes: (string | undefined)[] = [PlayoutContentType.VIDEO_CLIP, PlayoutContentType.VOICE_OVER]
-    return part.pieces.some(piece => supportedPlayoutContentTypes.includes((piece.metadata as Tv2PieceMetadata | undefined)?.playoutContent.type)) ?? false
+    return part.pieces.some(piece => supportedPlayoutContentTypes.includes((piece.metadata as PieceMetadata | undefined)?.playoutContent.type)) ?? false
```
KvelaGorrrrnio: Can't we just remove the type assertion? :) It seems to have the same type as if there were no assertion.

src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.ts
```diff
@@ -40,17 +39,17 @@ export class PieceTooltipComponent implements OnChanges {
   }
 
   public updateShouldShowHoverScrub(): void {
-    const tv2Piece: Tv2Piece = this.piece as Tv2Piece
-    this.shouldShowHoverScrub = this.hasPieceMedia() && [PlayoutContentType.VIDEO_CLIP, PlayoutContentType.JINGLE].includes(tv2Piece.metadata.playoutContent.type)
+    const pieceMetadata: PieceMetadata = this.piece.metadata as PieceMetadata
+    this.shouldShowHoverScrub = this.hasPieceMedia() && [PlayoutContentType.VIDEO_CLIP, PlayoutContentType.JINGLE].includes(pieceMetadata.playoutContent.type)
   }
```
KvelaGorrrrnio, line 44: Suggestion: Instead of using a type assertion, then a non-null assertion can be used instead:
```typescript
    this.shouldShowHoverScrub = this.hasPieceMedia() && [PlayoutContentType.VIDEO_CLIP, PlayoutContentType.JINGLE].includes(this.piece.metadata!.playoutContent.type)
  }
```

If it is kept with type assertion for the visibility (and searchability) then you can ignore this. Otherwise, we can use non-null assertion for most (if not all) of the type assertions for PieceMetadata.

KvelaGorrrrnio (approved):

src/app/rundown-view/components/rundown-tooltips/piece-tooltip/piece-tooltip.component.ts
RasmusAlbrektsen: I actually did keep it to ensure that if we change the need for asserting metadata it would be easy to locate everywhere we do so :-) 
And i do think we should change the need for asserting this data that we rely so much on and might always be sending with the DTO anyway.

RasmusAlbrektsen: I'm merging this for now, but i have no problem using non-null assertion instead. If you feel strongly about it just let me know and i'll open another PR for it :-)

PR #395, opened 2025-07-17
AG-547: Add es6 rundown clock

KvelaGorrrrnio:
Since we use older versions of CasparCG with Sofie and Alba, they don't
support ES2022. They only support ES6. So until we have upgraded the
CasparCGs to 2.4.X or higher we will use the ES6 version of rundown
clock.


vushu:
You mean CasparCG version 2.4.x :)

vushu (approved):

vushu (approved):

RasmusAlbrektsen (approved):

KvelaGorrrrnio:
> You mean CasparCG version 2.4.x :)

Thanks, I have updated the description 👍

PR #398, opened 2025-07-25
AG-313 Reduce amount of backend calls to the backend

LindvedKrvang:
Currently, every single Media in the RundownView needs to fetch the StudioConfiguration. This means, that since we don't cache the StudioConfiguration in the frontend, we would call the back _a lot_. 
This PR reduces the amount of calls to the backend when we have a lot of simultaneous calls.

https://github.com/user-attachments/assets/65c7c1d3-e53c-494d-bcfc-73b7c2200c8a

I tried see if I could use `ShareReplay` from RXJS, but I was unsuccesfull.
I opted not to make this a StateService, like our other state services, since we are currently not notifying clients about configuration changes (they are still made in old Sofie).


vushu (approved):

KvelaGorrrrnio:
A suggestion for naming: What is happening here is called "[Request coalescing](https://support.bunny.net/hc/en-us/articles/6762047083922-Understanding-Request-Coalescing)", and nothing is really cached in the flow, except for the request references. So a more telling name for the class might be something like "RequestCoalescingConfigurationService".

RasmusAlbrektsen (approved):

LindvedKrvang:
> A suggestion for naming: What is happening here is called "[Request coalescing](https://support.bunny.net/hc/en-us/articles/6762047083922-Understanding-Request-Coalescing)", and nothing is really cached in the flow, except for the request references. So a more telling name for the class might be something like "RequestCoalescingConfigurationService".

That's a way better name.
It has been updated :)

KvelaGorrrrnio (approved):

PR #401, opened 2025-08-04
AG-570 - Split screen inputs colors on virtual keyboard

RasmusAlbrektsen:
We didn't type check for split screen inputs. Most likely we lost this during the switch to `PlayoutContentType`.


NikolajDN (approved):

src/app/rundown-view/services/trigger-producer-key-binding.service.ts
```diff
@@ -295,7 +296,7 @@ export class TriggerProducerKeyBindingService implements KeyBindingService {
   }
 
   private isInsertIntoSplitScreenAction(action: Tv2Action): action is Tv2InsertIntoSplitScreenAction {
-    return action.metadata.playoutContent.type === PlayoutContentType.SPLIT_SCREEN && 'insertedContentType' in action.metadata
+    return action.metadata.playoutContent.type === PlayoutContentType.SPLIT_SCREEN_INPUT && 'insertedContentType' in action.metadata
```
LindvedKrvang, line 299: Did you verify this?
`insertedContentType` only exist on one interface and that interface explicitly states the `playoutContent.type` is `SPLIT_SCREEN`.
image

RasmusAlbrektsen: That's very interesting. The change works, but i wonder if it due to a side effect rather than the intension. 
I'll dig deeper once im through with the other PR's :-)

KvelaGorrrrnio: Is it correctly inferred that we don't need the `SPLIT_SCREEN_INPUT` playout content type?
Can we derive that it is an input from the `ìnsertedContentType` property?

RasmusAlbrektsen: Well ideally i guess it shouldn't be? 
However, until i made this change none of the input actions were caught and colored on the keyboard.

LindvedKrvang (approved):

PR #402, opened 2025-08-11
Upgrade angular to 18

KvelaGorrrrnio:
An attempt to upgrade our Angular version, in order to keep up with newer versions.
I hit a wall towards 19, where we need to decide if we are going to use standalone components for all our components or how we will approach it.


KvelaGorrrrnio:
Closing in favor of #422 and #423.

PR #403, opened 2025-08-15
Ag 631/show media countdown on rundown view

LindvedKrvang:
The RundownView now shows the remaining time of a Media being played.
It is looking for Media being played in either: The active Part or; on an auxiliary.

Once the Media reaches zero, the countdown turns red.

If no Media is being played, the countdown does not show a value.


src/app/core/services/rundown-timing.service.ts
```diff
@@ -122,4 +132,39 @@ export class RundownTimingService {
...
+      return this.getRemainingDurationOfMediaFromPiece(onAirVideoClipPiece, currentEpochTime)
+    }
+
+    const auxiliaryVideoClipPiece: Piece | undefined = this.findVideoClipPieceOnOutputLayer(rundown.infinitePieces, Tv2OutputLayer.AUXILIARY)
+    if (auxiliaryVideoClipPiece) {
+      return this.getRemainingDurationOfMediaFromPiece(auxiliaryVideoClipPiece, currentEpochTime)
+    }
+
+    return
+  }
```
KvelaGorrrrnio, line 158: The method uses multiple levels of abstraction for finding the piece in question and getting the remaining duration hereof.
We can consider to split this up into two distinct processes, such that we first find the piece, and then compute the remaining duration for the media.
E.g.:

```typescript
  public getRemainingDurationOfActiveMedia(rundown: Rundown, currentEpochTime: number): number | undefined {
    const onAirVideoClipPiece = this.getMostImportantActiveVideoClipPiece(rundown)
    if (!onAirVideoClipPiece) {
      return
    }
    return this.getRemainingDurationOfMediaFromPiece(onAirVideoClipPiece, currentEpochTime)
  }
  
  private getMostImportantActiveVideoClipPiece(rundown: Rundown): Piece | undefined {
    const onAirPieces: Piece[] = this.getOnAirPieces(rundown)

    const programVideoClipPiece: Piece | undefined = this.findVideoClipPieceOnOutputLayer(onAirPart.pieces, Tv2OutputLayer.PROGRAM)
    if (programVideoClipPiece) {
      return programVideoClipPiece
    }

    return this.findVideoClipPieceOnOutputLayer(onAirPart.pieces, Tv2OutputLayer.AUXILIARY) 
  }

  private getOnAirPieces(rundown: Rundown): Piece[] {
    const onAirSegment: Segment | undefined = rundown.segments.find(segment => segment.isOnAir)
    if (!onAirSegment) {
      return []
    }

    const onAirPart: Part | undefined = onAirSegment.parts.find(part => part.isOnAir)
    if (!onAirPart) {
      return []
    }
    return onAirPart.pieces
  }
```

src/app/core/services/rundown-timing.service.ts
```diff
@@ -122,4 +132,39 @@ export class RundownTimingService {
...
+  private findVideoClipPieceOnOutputLayer(pieces: Piece[], outputLayer: Tv2OutputLayer): Piece | undefined {
+    return pieces.find(piece => piece.metadata?.outputLayer === outputLayer && piece.metadata.playoutContent.type === PlayoutContentType.VIDEO_CLIP)
+  }
+
+  private getRemainingDurationOfMediaFromPiece(piece: Piece, currentEpochTime: number): number {
+    const media: Media | undefined = this.mediaStateService.getMedia(piece?.metadata?.sourceName + '')
+    const timeElapsedSinceExecutionOfPiece: number = currentEpochTime - (piece?.executedAt ?? 0)
+    const remainingDurationOfMedia: number = (media?.duration ?? 0) - timeElapsedSinceExecutionOfPiece - VIDEO_CLIP_DURATION_CUT_OFF
+    return Math.max(remainingDurationOfMedia, 0)
+  }
```
KvelaGorrrrnio: If `media` is made a parameter, then this method doesn't need to do "repository" work, nor handle the edge case when there is no media.

```typescript
  private getRemainingDurationOfMediaFromPiece(piece: Piece, media: Media, currentEpochTime: number): number {
    const timeElapsedSinceExecutionOfPiece: number = currentEpochTime - piece.executedAt
    const remainingDurationOfMedia: number = media.duration - timeElapsedSinceExecutionOfPiece - VIDEO_CLIP_DURATION_CUT_OFF
    return Math.max(remainingDurationOfMedia, 0)
  }
```

Ideally, the `MediaStateService` should be moved out of this class, as it is pure logic in here and `MediaStateService` introduces a slippery slope for state management to creep in :-)

My suggestion would be to move the `MediaStateService`  dependency to the `RundownTimingStateService` and split the flow up in finding the most important on air video clip piece and getting the remaining duration for media.
So from calling side, the flow would be:

1. Find the most important piece.
2. Get the media object based on the most important piece.
3. Compute the remaining duration for the media object.

Doing so will also remove the need for verifying class dependencies in the test :)

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.ts
```diff
@@ -87,6 +89,14 @@ export class OnAirDetailsPanelComponent implements OnChanges, OnInit, OnDestroy
     this.remainingDurationInMsForOnAirSegment = undefined
   }
 
+  private updateRemainingMediaDuration(rundownTimingContext: RundownTimingContext): void {
+    this.remainingDurationInMsForActiveMedia = rundownTimingContext.remainingDurationInMsForActiveMedia
+    if (this.remainingDurationInMsForActiveMedia && this.remainingDurationInMsForActiveMedia > 0) {
+      // Due to over Timer.pipe we need to negate the value to display the correct value
```
KvelaGorrrrnio: `our` instead of `over`?

src/app/core/services/rundown-timing.service.ts
```diff
@@ -4,10 +4,20 @@ import { Segment } from '../models/segment'
 import { PartEntityService } from './models/part-entity.service'
 import { Part } from '../models/part'
 import { Injectable } from '@angular/core'
+import { MediaStateService } from '../../shared/services/media-state.service'
+import { Media } from '../../shared/services/media'
+import { Piece } from '../models/piece'
+import { Tv2OutputLayer } from '../models/tv2-output-layer'
+import { PlayoutContentType } from '../enums/playout-content-type'
+
+const VIDEO_CLIP_DURATION_CUT_OFF: number = 5000 // This is a hardcoded value decided by TV2 to subtract from the remaining duration of video clips.
```
KvelaGorrrrnio: Should we add the reason why it is hardcoded? That it can be found in Sofie settings, but that we don't have the event structure set in place for ensuring the correct value.

KvelaGorrrrnio (approved):
Nice work :rocket: I think we should move the dependency on the `MediaStateService` out of `RundownTimingService` as it adds a dependency on out-of-process functionality and adds a backdoor for state management into a class of pure logic.
I am fine with the refactor happening in another tech-debt task, so we can test this in PCU-F. I will be happy to do a coop-session on it too 😃

src/app/core/services/rundown-timing.service.ts
RasmusAlbrektsen: We usually add `IN_MS` to these :-)

RasmusAlbrektsen (approved):

src/app/core/services/rundown-timing.service.ts
LindvedKrvang: Updated

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.ts
LindvedKrvang: Updated

PR #405, opened 2025-08-26
AG-632: Add media countdown in clock view

KvelaGorrrrnio:
Adds media countdown to rundown clock and legacy rundown clock.
image

The PR also:
- Makes media countdown "overflow" (counts how much overtime we are in).
- Increases part countdown font-size a bit and allows empty part names.
- Adds `/native` to the path for the angular version of the rundown clock, so both version are available.


src/app/core/services/rundown-timing.service.spec.ts
```diff
@@ -896,12 +897,13 @@ describe(RundownTimingService.name, () => {
...
           it('returns zero', () => {
-            const currentEpochTime: number = videoClipPiece.executedAt + media.duration + 10 * 1000
+            const tenSeconds: number = 10 * 1000
+            const currentEpochTime: number = videoClipPiece.executedAt + media.duration + tenSeconds
 
             const testee: RundownTimingService = createTestee({ mediaStateService })
             const result: number | undefined = testee.getRemainingDurationOfActiveMedia(rundown, currentEpochTime)
 
-            expect(result).toBe(0)
+            expect(result).toBe(-1 * tenSeconds - fiveSecondsCutOff)
```
LindvedKrvang, line 906: The name of the test says it is testing that it is returning zero, but it looks like the implementation tests that it returns minus five?
The name and implementation should be in line :)

src/app/standalone-shelf/components/on-air-rundown-clock/on-air-rundown-clock.component.html
```diff
@@ -26,15 +26,21 @@
...
-      <div class="label">Segment count down</div>
-      <div class="value" [class.overtime]="remainingDurationInMsForOnAirSegment > 0">{{ remainingDurationInMsForOnAirSegment | timer: '#+HH?:mm:ss' }}</div>
+    <div class="row">
+      <div class="box filled large bordered">
+        <div class="label">Segment count down</div>
+        <div class="value" [class.overtime]="remainingDurationInMsForOnAirSegment > 0">{{ remainingDurationInMsForOnAirSegment | timer: '#+HH?:mm:ss' }}</div>
+      </div>
+
+      <div class="box filled large bordered">
+        <div class="label">Media count down</div>
```
LindvedKrvang: Missing translation.
And if you want to be the real MVP, then you could also add the missing translations for the rest of this file :)

src/app/standalone-shelf/components/on-air-rundown-clock/on-air-rundown-clock.component.ts
```diff
@@ -68,7 +69,9 @@ export class OnAirRundownClockComponent implements OnInit {
     this.remainingDurationInMsForRundown = rundownTimingContext.currentEpochTime + rundownTimingContext.remainingDurationInMsForRundown - rundownTimingContext.expectedEndEpochTimeForRundown
     this.localEpochTime = rundownTimingContext.currentEpochTime
     this.expectedEndEpochTime = rundownTimingContext.expectedEndEpochTimeForRundown
+    this.negatedRemainingDurationInMsForActiveMedia = rundownTimingContext.remainingDurationInMsForActiveMedia
```
LindvedKrvang: This line seems redundant.
First off, it doesn't negate the value.
Second off, you call `updateNegatedRemainingDurationForActiveMediaInMs()` right below this line which also sets the `negatedRemaningDurationInMsForActiveMedia`.

src/standalone/rundown-clock/index.html
```diff
@@ -152,14 +152,25 @@
...
   window['remove'] = () => {}
 
   const host = window.location.host
+  const apiBaseUrl = host + '/api'
+  const websocketBaseUrl = host + '/events'
+
+  // For local development:
+  // const host = window.location.hostname
+  // const apiBaseUrl = host + ':3005/api'
+  // const websocketBaseUrl = host + ':3006'
```
LindvedKrvang, line 169: Is this something we want in our production code?

LindvedKrvang (approved):

src/app/standalone-shelf/components/on-air-rundown-clock/on-air-rundown-clock.component.ts
KvelaGorrrrnio: Thanks! It is now gone 😃

src/standalone/rundown-clock/index.html
KvelaGorrrrnio: Since this is a temporary file, that is removed when we upgrade CasparCG - I will say yes.
It makes it easier to test local, as both path and port changes between production and local development :)

src/app/standalone-shelf/components/on-air-rundown-clock/on-air-rundown-clock.component.html
KvelaGorrrrnio: Translations are added for all labels in the rundown clock view.

src/app/core/services/rundown-timing.service.spec.ts
KvelaGorrrrnio: Nice catch! I forgot to update the text, but I will not change it as the test case is removed in #406.

PR #406, opened 2025-08-26
AG-647: Splits flow for getting remaining duration of active video clips. 

KvelaGorrrrnio:
The new flow consists of 3 flows:
1. Find active video clip piece (via RundownEntityService).
2. Get Media object (via MediaStateService).
3. Compute remaining duration (via RundownTimingService).

This refactor is only applied in the angular project - not the legacy rundown clock.


src/app/core/services/models/rundown-entity.service.spec.ts
```diff
@@ -374,6 +377,179 @@ describe(RundownEntityService.name, () => {
...
+
+        expect(result).toBeUndefined()
+      })
+    })
+
+    describe('when the active part has one video clip piece', () => {
+      const media: Media = TestEntityFactory.createMedia({
+        sourceName: 'media-id',
+        duration: 60_000,
+      })
```
LindvedKrvang, line 468: FYI: This is created "outside the scope" of the tests i.e. it is the exact same instance that is being provided to all of the tests it is used in.
This means that if one test alters someting on the Media, then that is persistet to the other tests. This could be a potential source for randomly failing tests.
Not saying we shouldn't do this, but we should be aware of the possible implications.

src/app/core/services/models/rundown-entity.service.spec.ts
```diff
@@ -374,6 +377,179 @@ describe(RundownEntityService.name, () => {
       })
     })
   })
+
+  describe(RundownEntityService.prototype.getActiveVideoClipPiece.name, () => {
```
LindvedKrvang, line 382: Missing a test case for when there is a VideoClipPiece in the InfinitePieces, but the Active Part does _not_ have a VideoClipPiece i.e. all tests that currently tests that we get a VideoClipPiece has a VideoClipPiece on the Active Part :)

src/app/core/services/models/rundown-entity.service.spec.ts
```diff
@@ -374,6 +377,179 @@ describe(RundownEntityService.name, () => {
...
+          const testee: RundownEntityService = createTestee()
+
+          const result: Piece | undefined = testee.getActiveVideoClipPiece(rundown)
+
+          expect(result).toBe(videoClipPiece)
+        })
+      })
+
+      describe('when the video clip piece is on the auxiliary output layer and is an infinite piece', () => {
+        it('returns the video clip piece', () => {
```
LindvedKrvang, line 508: It would be nice if there was a distinction between the VideoClipPiece on the ActivePart and the VideoClipPiece in InfinitePieces.
The logic should be that if there is both a VideoClipPiece on the ActivePart and in InfinitePieces, then it should return the one from the ActivePart.
With the current test implementation, it would still parse if we returned the VideoClipPiece from InfinitePieces, which is not the logic we want :)

LindvedKrvang (approved):

src/app/core/services/models/rundown-entity.service.spec.ts
KvelaGorrrrnio: Good point. I have marked the attributes of Media readonly, as to discourage mutations on the object.

src/app/core/services/models/rundown-entity.service.spec.ts
KvelaGorrrrnio: I have added a test case for this :)

src/app/core/services/models/rundown-entity.service.spec.ts
KvelaGorrrrnio: I am not entirely sure what you mean with this - perhaps we should take it face-to-face? :)

PR #428, opened 2025-09-16
AG-717: Improve performance of rundown view

KvelaGorrrrnio:
> [!note]
> The reference point is the CPU usage prior to v19. Testing happens on a MacBook Pro M1 Max, with the browser window being rendered on the laptop screen (120FPS support).

This PR aims to improve the performance (primarily CPU usage) of the rundown view.

**Performance improvements:**
- Changes change detection strategy to `OnPush` for:
  - `VirtualKeyboardComponent` - it heavily reduced the CPU utilization.
  - `ProducerShelfComponent`.
  - `SegmentComponent`.
  - `OffsetablePieceComponent`.
  - `StickySegmentComponent`.
- Time reference updating (with `requestAnimationFrame`) is moved from SegmentComponent to `FollowPlayheadTimline` as it is the only consumer of it.
- Rendering timeline markers from time reference updates, is triggered by a public method on the TimelineMarkersComponent. By avoiding the Angular `@Input` for the time reference, a lot of change detection triggers is avoided - it heavily reduced CPU utilization and increased stability of the timeline markers rendering.

**Other improvements:**

- Corrects size of sidebar in producer shelf to match the size of the static actions panel. This is a fix and the static action panel sizing should be re-evaluated as there are some competing concepts.
- Removes `trackX` methods in favor of the `track <expr>` introduced with Control Flow syntax.

## Summary
In my testing with the applied changes, my CPU utilization went from around 98%-100% down to 20%-35% on average, when rendering an active rundown in the rundown view. Prior to the Angular migration, the utilization for a similar use case was around 45%-60%. So we actually have a better performing system now, than prior to the Angular version migrations.

## Potential future improvements
- Re-engineer how we render parts and pieces on the timeline.
  - One way is to animate by translation (`transform: translateX()`) rather than resizing the boxes. Resizing is a heavier operation to calculate and render than translation.
  - Another option that doesn't require a total re-engineering, is to see if we can bypass the Angular change detection (`@Input`) for time updates to parts and pieces.
- Look into how we pass time data down from the RundownComponent to reduce the how many changes Angular detects.


LindvedKrvang (approved):
Nice work

rana-tv2 (approved):

PR #432, opened 2025-09-24
AG-316 timings in rundown header do not have fixed position when e.g. showstyle variant changes

rana-tv2:
This PR is a refactor For HTML and CSS in the following components:
-  rundown-header-panel
- on-air-details-panel

the effort is to have the timings centred in the middle of the screen and not affected when SCHEMA, DESING, SHOW STILE VARIANT  changes in a rundown.


src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.scss
```diff
@@ -53,3 +58,21 @@
...
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+}
+
+.on-air-details-panel-right {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+}
```
KvelaGorrrrnio: You can make use of the fact that we write SCSS, and reduce the redeclaration of the flex property and flex direction. E.g.
```scss
.subpanel-group {
  display: flex;
  flex-direction: row;

  &.left {
    justify-content: flex-start;
  }

  &.center {
    justify-content: center;
  }

  &.right {
    justify-content: flex-end;
  }
}
```

That way we can combine the classes in template e.g. `<div class="subpanel-group left">`.

src/app/rundown-view/components/rundown-header/rundown-header.component.scss
```diff
@@ -22,8 +22,26 @@
...
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+  }
+
+  &--right {
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+  }
```
KvelaGorrrrnio: In BEMIT `--xyz` is used a modifier for changing some properties on the base class. As this is used for subelements (and that we want to abandon BEMIT in this project), I suggest that we make these classes under the `.c-rundown-header` class. We can use the same approach as I outlined for the on-air-details-panel, by utilizing SCSS hierarchy:
```scss
.panel-group {
  display: flex;
  flex-direction: row;

  &.left {
    justify-content: flex-start;
  }

  &.center {
    justify-content: center;
  }

  &.right {
    justify-content: flex-end;
  }
}
```
Then we can use it as `<div class="panel-group left">` in the template.

KvelaGorrrrnio (approved):
Nice work :)

src/app/rundown-view/components/on-air-details-panel/on-air-details-panel.component.scss
rana-tv2: implemented the suggestion to use `.subpanel-group`

src/app/rundown-view/components/rundown-header/rundown-header.component.scss
rana-tv2: implemented the suggestion and added `.panel-group` class under `.c-rundown-header`

RasmusAlbrektsen (approved):
Nice :-)

KvelaGorrrrnio (approved):

PR #435, opened 2025-09-25
AG-708 keyboard settings virtual keyboard is too small

rana-tv2:
This PR concerns two point:
- adjusting the virtual keyboard to a usable size 
- adding the logic required to allow hiding the keyboard when clicking outside of the keyboard.
One issue that seems to not quite work is that when hiding the keyboard by clicking the keyboard button the first click to reopen it doesn't do anything, but the second click opens it again. 

https://github.com/user-attachments/assets/e8e782aa-304d-46dc-a874-4fa2c6d44b1b


src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
```diff
@@ -138,8 +143,40 @@ export class EditKeyboardMappingComponent implements OnInit {
...
+    this.removeClickOutsideListener()
+    this.clickOutsideListener = (): void => {
+      document.addEventListener('click', this.handleClickOutside.bind(this), { once: true })
+    }
+    setTimeout(() => {
+      if (this.clickOutsideListener) {
+        this.clickOutsideListener()
+      }
+    })
   }
```
KvelaGorrrrnio, line 157: If I understand correct you are binding a lambda function to `clickOutsideListener`, such that you can unregister it at a later point. If that understanding is correct, then it looks like you are binding the wrong lambda. The method you want to unregister is `this.handleClickOutside.bind(this)`. So I would suggest that you assign this method to `clickOutsideListener`, so that this becomes:
```typescript
class EditKeyboardMappingComponent implements ... {
  private addClickOutsideListener(): void {
    this.removeOutsideListener()
    this.clickOutsideListener = (event: Event): void => this.handleClickOutside(event)
    setTimeout(() => {
      if (this.clickOutsideListener) {
        document.addEventListener('click', this.clickOutsideListener, { once: true })
      }
    })
  }
}
```
I don't think the setTimeout is necessary if you stop the immediate propagation of the original event. If we stop the immediate propagation of the event that triggers `toggleKeyboardVisibility` (can be done with `$event.stopImmedaitePropgation(); toggleKeyboardVisibility` in the template), then we can remove the setTimeout and this can be simplified a bit. By doing this we also remove the need for checking for existence of the clickOutsideListener:
```typescript
class EditKeyboardMappingComponent implements ... {
  private addClickOutsideListener(): void {
    this.removeOutsideListener()
    this.clickOutsideListener = (event: Event): void => this.handleClickOutside(event)
    document.addEventListener('click', this.clickOutsideListener, { once: true })
  }
}
```
These changes will incidentally fix the issue that after closing the keyboard shortcut a double-click is needed to open it again.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
```diff
@@ -36,14 +37,13 @@
...
   width: 0;
-  border-bottom: 80px solid black;
-  border-left: 20px solid transparent;
-  border-right: 20px solid transparent;
-  border-top: none;
+  border-bottom: 60px solid var(--rundown-keyboard-background-color);
+  border-left: 40px solid transparent;
+  border-right: 0 solid transparent;
   cursor: auto;
   z-index: 101;
```
KvelaGorrrrnio: As we looked at together, we can simplify this by making this a pseudo element (`::before`) on the actual keyboard element (in this component). This way we can position the triangle relative to the position of the virtual keyboard, instead of relative to the document.

Another minor point of improvement: There is a redundant div-element (the one that is marked with `#floatingKeyboard`).

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
```diff
@@ -18,10 +18,11 @@
...
 .floating-keyboard {
-  position: absolute;
-  top: 250px;
-  left: 25%;
-  width: 74%;
+  position: fixed;
+  top: 210px;
+  right: 20px;
+  width: 100%;
+  max-width: 1300px;
```
KvelaGorrrrnio: The current styling doesn't completely scale for smaller screen sizes, as it is translated 20px left from the right edge of the screen, but still aims for 100% of the width. The simple fix would be to clamp the max-width to `calc(100% - 40px)`, so we ensure a balanced margin on both sides of the virtual keyboard.

Image

The next comment has nothing to do with your solution: I am not fond of the fact that the virtual keyboard is positioned with relation to the document instead of the anchor that summons it.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
rana-tv2: Updated `addClickOutsideListener` as suggested such that it has the correct binding.

the setTimeut is necessary as the `floatingKeyboardElement` `ViewChild` is not static and may not be resolved until after change detection, although setting the timeout to 0 seems to solve the clicking twice issue.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
rana-tv2: refactored to remove `.triangle` class and made it pseudo element instead.
And removed to redundant div-element.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
rana-tv2: refactored such that the `.floating-keyborad` now has 
`
  width: 100%;
  max-width: cal(100% - 20px);
`


regarding the last comment: I couldn't find a way to fix the size issue without having `position: fixed` on the `.floating-keyboard` any suggestions for a better solution is very welcome 😊

RasmusAlbrektsen (approved):
Good stuff :-)

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
```diff
@@ -18,34 +18,31 @@
...
-  width: 74%;
+  position: fixed;
+  top: 2;
+  left: 10px;
+  right: 0;
+  width: 100%;
+  max-width: calc(100% - 20px);
   display: flex;
-  flex-direction: column;
+  flex-direction: row;
```
KvelaGorrrrnio: Since `.floating-keyboard` is used on the `<alba-virtual-keyboard />` element, the display and flex configuration can be removed as these are implicitly set in the VirtualKeyboardComponent. Setting them here is a potential cause for issues if the layout is changed within the component. Same goes for the `gap` attribute.

KvelaGorrrrnio (approved):

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.scss
KvelaGorrrrnio: Already fixed.

src/app/settings/keyboard-mapping-settings/components/edit-keyboard-mapping/edit-keyboard-mapping.component.ts
KvelaGorrrrnio: The setTimeout is avoided by stopping the click events propagation with `$event.stopImmedaitePropagation`.

PR #442, opened 2025-10-09
feat: Enables reordering of static actions using drag and drop

RasmusAlbrektsen:
This PR enables users to reorder their static actions using drag and drop in the edit-sidebar. 
It works by simply reordering the array of action ID's and therefore requires no backend changes.

This PR also includes a reordering of the configuration cards that makes it look more like the way our action panels show in the producer shelf of the rundown view. 

On a small screen, the static actions will move below the action panel table instead. 

image


RasmusAlbrektsen:
Setting the PR as a draft because i'm looking into the responsiveness when resizing and letting the cards span the space.

src/app/settings/shelf-settings/components/static-buttons-configuration-card/static-buttons-configuration-card.component.html
```diff
@@ -1,7 +1,7 @@
 <alba-card class="pt-2 pl-0">
   <ng-container cardHeader>
     <span i18n>static-buttons-configuration.card-header</span>
-    <alba-icon-button [icon]="Icon.PEN" [iconSize]="IconSize.S" tooltip="Edit Static Buttons" (click)="openEditStaticButtons()" />
+    <alba-icon-button [icon]="Icon.PEN" [iconSize]="IconSize.S" tooltip="Edit Static Actions" (click)="openEditStaticButtons()" />
```
LindvedKrvang, line 4: Would be awesome if the tooltip was translated :)

src/app/settings/shelf-settings/components/shelf-action-panel-settings-page/shelf-settings-page.component.html
```diff
@@ -1,18 +1,20 @@
...
   @if (shelfConfiguration) {
-    <alba-shelf-action-panels-configuration [shelfConfiguration]="shelfConfiguration" />
-    <div class="mt-5">
-      <alba-static-buttons-configuration-card />
-    </div>
-    <alba-card class="mt-5 pb-3">
-      <ng-container cardHeader i18n>
-        shelf-settings.misc.title
-      </ng-container>
+    <div class="sub-grouping">
```
LindvedKrvang: You can't use our `.row` css class?

LindvedKrvang (approved):
Looks good :)

Personal preference: I'm not fan of having the configuration cards next to each other. But since we don't have an "official guidelines" yet, this is simple one vote for a potential outcome ;)

src/app/settings/shelf-settings/components/shelf-action-panel-settings-page/shelf-settings-page.component.html
RasmusAlbrektsen: I could, but as we are discussing in DM's, this is not responsible for keeping things on a row now. This changes depending on the media query for the width and can just as easily depict the components below in a column.

src/app/settings/shelf-settings/components/static-buttons-configuration-card/static-buttons-configuration-card.component.html
RasmusAlbrektsen: It really would. Seems that our translations broke with the Angular-update. I haven't been able to fix it, but i will try here until there are two approves. Otherwise i'll try in a different PR :-)

rana-tv2 (approved):
Neat work 👍

KvelaGorrrrnio (approved):

PR #443, opened 2025-10-14
Use Angular-based rundown clock over legacy version

KvelaGorrrrnio:
> [!warning]
> This change breaks rundown clock rendering on older CasparCG versions (at least 2.1).

I have added AG-746 as the reference ticket, as the current hypothesis is that the issue lies in the legacy-version of the rundown clock - so we expect that the task is will resolve itself if we use the Angular-based rundown clock. This has a major drawback that the rundown clock only renders on newer versions of CasparCG. So it entails that we upgrade Alba galleries to CasparCG 2.4.

This is a team-decision, and it is dependent on the Alba graphics role being tested with CasparCG 2.4 and deployed to production, which is why this PR is marked as _draft_.

## Change description

The PR removes the legacy version of the rundown clock and changes the route for the Angular-based one from `/standalone/rundown-clock/native` to `/standalone/rundown-clock`.

In addition, the PR also adds the CasparCG event callbacks when loading the rundown clock to lessen the number of warnings in the CasparCG log.

## Checklist

- [X] The changes are documented.
         _No documentation changes are necessary for the change._
- [X] Core logic is covered by tests
         _No logic is changed._


RasmusAlbrektsen:
I think we are alright to include this PR in general. In both Gallery D and in OB6, they never mentioned that we were missing the clock view, so the fact that they will go without until we streamline CasparCG versions will probably be fine :-)

RasmusAlbrektsen (approved):

LindvedKrvang (approved):

PR #444, opened 2025-10-17
The PanlInputId is now validated in its entirety instead of just checking the first character

LindvedKrvang:
The validator logic for our PanelInputId was simply to check whether the first character is a digit or not. This was chosen since the id of a PanelInput is simply the index given by the manufacturer.

When we are creating the id for an input using modifiers, we join the UUIDs of the modifiers with the PanelId. This gives us an id following the pattern: `{UUDI_1};{UUDI_2};{UUDI_3};{panelId}`.

The UUIDs of the modifiers are created by the backend. Since they are UUIDs, they are allowed to have a letter as the first character.
This gave us an edgecase where, if a Modifier had gotten an id that started with a letter, then our validation for the PanelInputId would fail. This makes the user unable to ever save the PanelLayout.

This PR updates the validation by introducing a new validator for the PanelInputId.
This validator checks if the id follows our rules for the PanelInputId.


rana-tv2 (approved):

src/app/settings/panel-settings/services/panel-input-id-validator.service.ts
```diff
@@ -0,0 +1,51 @@
...
+@Injectable()
+export class PanelInputIdValidator {
+  /**
+   * This validator validates whether the id of a PanelInput is valid.
+   * The id can be valid in two forms:
+   * - The id can consist of a number e.g. "21".
+   * - The id can consist of UUIDs followed by a number all separated by ';' e.g. "e90f88f8-ddef-4ef9-a975-1bb0bcef3909;af5fedad-c743-41e7-8da6-353d521a38ba;21".
+   * The id is not allowed to start with a semicolon.
+   */
+  public static validate(): ValidatorFn {
```
KvelaGorrrrnio: There is a mismatch between function name and return type. An alternative could be to call it `createAngularFormsValidator`, as it makes it clear what it does and allows for other types of validators to coexist.
The class itself could also be renamed to `PanelInputIdValidatorFactory` to exhibit the intent more clearly.

src/app/settings/panel-settings/services/panel-input-id-validator.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    return (control: AbstractControl): ValidationErrors | null => {
+      const value = control.value
+      if (typeof value !== 'string') {
+        throw new Error(`Invalid type - value must be a string - ${value}`)
+      }
```
KvelaGorrrrnio: Question: How does throwing an error behave in contrast to returning a validation error?

src/app/settings/panel-settings/services/panel-input-id-validator.service.ts
```diff
@@ -0,0 +1,51 @@
...
+
+      const idSegments: string[] = id.split(';')
+      const isValidPanelInputId: boolean = idSegments.every((idSegment: string, index: number) => {
+        if (index === idSegments.length - 1) {
+          return IS_ONLY_DIGITS_REGEX.test(idSegment)
+        }
+
+        return UUID_REGEX.test(idSegment)
+      })
```
KvelaGorrrrnio: The method has multiple responsibilities, namely: Validating an input and composing an Angular Forms validator. The logic for actually validating the panel input id, could be extracted to a separated method. This allows for reuse internally, easy extraction if/when it is needed elsewhere, and can make it easier to test.

KvelaGorrrrnio (commented):
Nice work - I have some semantic and structural comments.

RasmusAlbrektsen (approved):

rana-tv2 (approved):
🚀

src/app/settings/panel-settings/services/panel-input-id-validator.service.ts
KvelaGorrrrnio: I have updated the class to `PanelInputIdValidatorFactory` and method to `createAngularFormsValidator` in bd39a5c92b3fdb01a9daecc16655efc013534575.

src/app/settings/panel-settings/services/panel-input-id-validator.service.ts
KvelaGorrrrnio: I have converted this to result in an invalid id control in bd39a5c92b3fdb01a9daecc16655efc013534575, to make it visible to the user that something is wrong. (I know that it is not the best UX)

src/app/settings/panel-settings/services/panel-input-id-validator.service.ts
KvelaGorrrrnio: I have split up the validation logic from the Angular Forms response composition in Tbd39a5c92b3fdb01a9daecc16655efc013534575.
The checks prior to the split are removed, as they are also handled in the `idSegments.every(...)` block.

PR #445, opened 2025-10-17
AG-773 - Update shelf configuration in the producer shelf

RasmusAlbrektsen:
## Change description

This PR is a fix for the fact that we did not update the view whenever our `shelfConfiguration` received updates. To ensure this happens, i moved the `shelfConfiguration` to the `ProducerShelfCompoent` instead, and then gave it as an input to child components, in aligment  with the "On Push" update strategy.


src/app/producer-shelf/components/action-panels/action-panels.component.ts
```diff
@@ -38,18 +35,17 @@ const CHECK_ON_AIR_PIECES_INTERVAL_IN_MS: number = 200
 })
 export class ActionPanelsComponent implements OnInit, OnChanges, OnDestroy {
   @Input() public rundown: Rundown
+  @Input() public shelfConfiguration?: ShelfConfiguration
```
KvelaGorrrrnio, line 38: Nice work on the PR! :) One thing to consider is where we need the optionality of the shelf configuration - and where we don't need it. For static actions and action panels, I will argue that they do not have a merit for being displayed if we don't have a shelf configuration.

It would force the parent components to model what e.g. loading states mean for layout and rendering.

Just food for thought – I'll gladly have a chat about it :)

KvelaGorrrrnio (approved):
Nice work :)

rana-tv2 (approved):

src/app/producer-shelf/components/action-panels/action-panels.component.ts
RasmusAlbrektsen: @KvelaGorrrrnio i like the sound of this. Would you consider this to be a seperate PR?
Otherwise i'll keep this PR up for now and then likely bring it up with you tomorrow for a chat :-)

PR #447, opened 2025-10-23
AG-746: Only show part countdown for video clips in rundown clock

KvelaGorrrrnio:
Prior to the change, the part countdown was not shown uniformly across the rundown view and the rundown clock.
It was always shown for the on air part in the rundown clock, but only for video clips (and voice overs) in the rundown view.
Keeping it consistent was a reported desire by our super users.

## Change description

The part countdown is only shown for video clips (and voice overs) in the rundown clock.

It is worth looking into joining the timings that we display - and re-evaluate how they behave and are displayed.

## Checklist

- [X] The changes are documented.
   _No documentation is written as the cost of fixating the current logic is higher than being able to manipulate the logics (co)location and behavior._
- [X] **Core logic is covered by tests.**
  _No tests are written as the cost of fixating the current logic is higher than being able to manipulate the logics (co)location and behavior._


rana-tv2 (approved):
looks good 👍

RasmusAlbrektsen (approved):

PR #454, opened 2025-11-06
AG-835 - Shelf panels only show when a shelfConfiguration is loaded

RasmusAlbrektsen:
## Change description

This change is to account for the fact that we do not want to show UI elements that have no relevance. If no shelf configuration has been provided.


rana-tv2 (approved):
looks good 👍

RasmusAlbrektsen:
I converted this to a draft since i'd like to rethink the approach and maybe change the input for the `StaticActionsComponent` to be the `staticActionsIds`. This way, we can only render the static actions if these are provided. I will request reviews once this is done

PR #455, opened 2025-11-10
AG-837 - Fix change detection for shelf action panels

RasmusAlbrektsen:
## Change description

This PR is a fix to the fact that when entering a rundown from the overview-page, we would end up in a state where the view doesn't update once we have both the shelf configuration and the actions we need. 

This has been ensure by enforcing OnPush-strategy for detection and also manually marking the detection in the components in question.

PR #456, opened 2025-11-11
AG-835 - Load shelf configuration components only when relevant

RasmusAlbrektsen:
> [!warning]
> This PR is based on #455 .

## Change description

This PR changes the `ProducerShelfComponent` to instead use only the parts of the `ShelfConfiguration` that has relevance. 

Since we still account for the possibility of an undefined `shelfConfiguration` in the rundown view (we should look into changing this), we still have to make som if-statements in the HTML that i would like to do without.

For the standalone-shelf we are not able to use the same approach since this isn't a component created with inputs.


src/app/producer-shelf/components/action-panels/action-panels.component.ts
```diff
@@ -136,11 +136,11 @@ export class ActionPanelsComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   private updateActionPanels(): void {
-    if (!this.shelfConfiguration || !this.rundown) {
+    if (!this.actionPanelConfigurations || !this.rundown) {
```
KvelaGorrrrnio: An empty list is a truthy value, so we can remove this check as it will just result in an empty list for `this.resolvedActionPanels`.

And should we set `this.resolvedActionPanels` to an empty list in this case?

```suggestion
    if (!this.rundown) {
```

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
```diff
@@ -17,5 +17,6 @@ export class ProducerShelfComponent {
   @Input() public keyBindings: KeyBinding[]
   @Input() public keystrokes: string[]
   @Input() public rundown: Rundown
-  @Input() public shelfConfiguration?: ShelfConfiguration
+  @Input() public staticActionIds?: string[]
+  @Input() public actionPanelConfigurations?: ShelfActionPanelConfiguration[]
```
KvelaGorrrrnio: Both `staticActionIds` and `actionPanelConfigurations` should not be optional. They are lists - an empty list can signal the same state as `undefined` :) It also makes it the checks and subsequent handling easier.

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
```diff
@@ -1,10 +1,14 @@
 <div class="wrapper">
     <div class="main">
-      <alba-action-panels [rundown]="rundown" [shelfConfiguration]="shelfConfiguration" />
+      @if (actionPanelConfigurations && actionPanelConfigurations.length > 0) {
```
KvelaGorrrrnio: If `actionPanelConfigurations` become non-nullable, then this can be simplified to just checking the length directly:

```suggestion
      @if (actionPanelConfigurations.length > 0) {
```

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
```diff
@@ -1,10 +1,14 @@
...
+        <alba-action-panels [rundown]="rundown" [actionPanelConfigurations]="actionPanelConfigurations" />
+      }
       <alba-virtual-keyboard [keyBindings]="keyBindings" [keystrokes]="keystrokes" />
     </div>
 
-    <div class="sidebar">
-        <alba-static-actions [rundown]="rundown" [shelfConfiguration]="shelfConfiguration" [centerActions]="true" />
-    </div>
+      @if (staticActionIds && staticActionIds.length > 0) {
```
KvelaGorrrrnio: If `actionPanelConfigurations` become non-nullable, then this can be simplified to just checking the length directly:

```suggestion
      @if (staticActionIds.length > 0) {
```

src/app/rundown-view/components/rundown-view/rundown-view.component.html
```diff
@@ -26,7 +26,8 @@
           [keyBindings]="keyBindings"
           [keystrokes]="keystrokes"
           [rundown]="rundown"
-          [shelfConfiguration]="shelfConfiguration"
+          [actionPanelConfigurations]="shelfConfiguration ? shelfConfiguration.actionPanelConfigurations : []"
+          [staticActionIds]="shelfConfiguration ? shelfConfiguration.staticActionIds : []"
```
KvelaGorrrrnio: Alternative syntax: Using optional chaining can make it a bit more concise:

```typescript
          [actionPanelConfigurations]="shelfConfiguration?.actionPanelConfigurations ?? []"
          [staticActionIds]="shelfConfiguration?.staticActionIds ?? []"
```

src/app/standalone-shelf/components/standalone-action-shelf/standalone-action-shelf.component.html
```diff
@@ -1,7 +1,13 @@
...
-    <alba-action-panels class="action-panels" [rundown]="rundown" [shelfConfiguration]="shelfConfiguration" />
-    <alba-static-actions [rundown]="rundown" [centerActions]="true" [shelfConfiguration]="shelfConfiguration" />
+    @if (shelfConfiguration) {
+      @if (shelfConfiguration.actionPanelConfigurations.length > 0) {
+        <alba-action-panels class="action-panels" [rundown]="rundown" [actionPanelConfigurations]="shelfConfiguration.actionPanelConfigurations" />
+      }
+      @if (shelfConfiguration.staticActionIds.length > 0) {
+        <alba-static-actions [rundown]="rundown" [centerActions]="true" [staticActionIds]="shelfConfiguration.staticActionIds" />
+      }
+    }
```
KvelaGorrrrnio, line 10: Are the nested checks necessary?
The only downside I see is that we can have an empty static actions panel.

src/app/producer-shelf/components/static-buttons/static-actions.component.ts
```diff
@@ -20,7 +19,7 @@ import { Tv2ActionPanelComponent } from '../tv2-action-panel/tv2-action-panel.co
 export class StaticActionsComponent implements OnInit, OnChanges, OnDestroy {
   @Input() public rundown: Rundown
   @Input() public centerActions: boolean = false
-  @Input() public shelfConfiguration?: ShelfConfiguration
+  @Input() public staticActionIds: string[]
```
KvelaGorrrrnio: Nitpick: We can add the default value of `[]` to ensure that we don't reference an `undefined` value. That is true for other components as well. I don't know if we will ever encounter an issue with it - so treat this as food for thought :D

src/app/producer-shelf/components/action-panels/action-panels.component.ts
RasmusAlbrektsen: Changed in: https://github.com/tv2/alba-web-client/pull/456/commits/24e20c1a51870fe40957803e82936792d9ce8c29 to fit the suggestion and i set the list to empty since this is the same approach as for static action and i agree with the approach.

src/app/standalone-shelf/components/standalone-action-shelf/standalone-action-shelf.component.html
RasmusAlbrektsen: It was to avoid the empty static actions panel. I worked from the approach that if something isn't configured in the settings, we shouldn't take space from the other components. In this instance we can let the action panels span the rest of the right side of the screen. 

I could change it to:
```TypeScript
    @if (shelfConfiguration && shelfConfiguration.actionPanelConfigurations.length > 0) {
      <alba-action-panels class="action-panels" [rundown]="rundown" [actionPanelConfigurations]="shelfConfiguration.actionPanelConfigurations" />
    }
    @if (shelfConfiguration && shelfConfiguration.staticActionIds.length > 0) {
      <alba-static-actions [rundown]="rundown" [centerActions]="true" [staticActionIds]="shelfConfiguration.staticActionIds" />
    }
```

Still not pretty, but it removes the nesting. What do you think?

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
RasmusAlbrektsen: Changed to the suggestion in https://github.com/tv2/alba-web-client/pull/456/commits/4fb832f0f5435f487dde1da27200b507226eb9cc

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.html
RasmusAlbrektsen: Changed to the suggestion in https://github.com/tv2/alba-web-client/pull/456/commits/4fb832f0f5435f487dde1da27200b507226eb9cc

src/app/producer-shelf/components/static-buttons/static-actions.component.ts
RasmusAlbrektsen: Changed in https://github.com/tv2/alba-web-client/pull/456/commits/4fb832f0f5435f487dde1da27200b507226eb9cc

src/app/rundown-view/components/rundown-view/rundown-view.component.html
RasmusAlbrektsen: Changed in https://github.com/tv2/alba-web-client/pull/456/commits/68549060568d0157743d150cccaeee1f0d081535

src/app/producer-shelf/components/producer-shelf/producer-shelf.component.ts
RasmusAlbrektsen: They are no longer optional!

LindvedKrvang (approved):

src/app/standalone-shelf/components/standalone-action-shelf/standalone-action-shelf.component.html
KvelaGorrrrnio: It is fine to keep it as is :)

KvelaGorrrrnio (approved):

PR #457, opened 2025-11-11
Refactor - Change routing to standalone components

RasmusAlbrektsen:
> [!warning]
> If we accept this suggestion and merge the PR, we need to make a note in the post deploy configuration to ensure that we change the shortcut used for tablet in OB6. It will also break in the staging-environment.
> It now has to route to the path: `/standalone/rundown/actions` instead.
> This also requires us to make a blueprints change that now points to the new path for the rundown clock.

## Change description

This PR is a suggestion as to how we can rethink the routing to our standalone components based on their context. The `RundownClockComponent` and now `RundownActionOverviewComponent` (priorly `StandaloneActionShelfComponent`) are only relevant in the context of a rundown and can therefore be found under the `standalone/rundown`-path, e.g:
`/standalone/rundown/clock`

I renamed the `StandaloneActionShelfComponent` since i feel the naming is redundant in the context of the route it is used from. 

## Checklist


src/app/standalone/routes.ts
```diff
@@ -0,0 +1,28 @@
...
+import { RundownClockComponent } from './components/rundown-clock/rundown-clock.component'
+
+export enum StandalonePaths {
+  RUNDOWN = 'rundown',
+}
+
+export enum RundownPaths {
+  CLOCK = 'clock',
+  ACTION_OVERVIEW = 'action-overview',
+}
```
KvelaGorrrrnio: We can consider whether this should be in its own module to keep it isolated from future standalone modules (e.g. developer or support standalones) :D

src/app/standalone/routes.ts
```diff
@@ -0,0 +1,28 @@
+import { RundownActionOverviewComponent } from './components/action-overview/rundown-action-overview.component'
+import { Routes } from '@angular/router'
+import { RundownClockComponent } from './components/rundown-clock/rundown-clock.component'
+
+export enum StandalonePaths {
+  RUNDOWN = 'rundown',
+}
+
+export enum RundownPaths {
+  CLOCK = 'clock',
```
KvelaGorrrrnio: The change for the rundown clock requires a blueprints change :)

src/app/standalone/rundown/components/rundown-actions/rundown-actions.component.ts
KvelaGorrrrnio, line 1: `action-overview` does not feel right to me. The purpose of it is more for controlling actions than looking at them :D 
Perhaps we can do a brainstorm on possible variations :grin:

KvelaGorrrrnio (commented):
Nice proposal! I like that we get rid of the shelf association for the standalone interfaces :+1: I haven't approved since I would like to have a chat/brainstorm on possible variations - or be convinced that `overview` is the best option :)
A note on the PR description - the use of actions on the tablet is described, but the URL referenced points to the rundown clock - just an inconsistency.

src/app/standalone/routes.ts
RasmusAlbrektsen: Added to the description, i seem to have mixed some things together by mistake :-D

KvelaGorrrrnio:
We could add some information on the "action overview" about what installation and rundown it is associated with.

src/app/standalone/rundown/components/rundown-actions/rundown-actions.component.ts
KvelaGorrrrnio: `/standalone/rundown/actions` and `RundownActionsComponent`.

KvelaGorrrrnio (approved):

src/app/standalone/routes.ts
RasmusAlbrektsen: Implemented in: https://github.com/tv2/alba-web-client/pull/457/commits/3dda0b522545fdc7adc71d6d50f5ed6ef10f67e0

src/app/standalone/rundown/components/rundown-actions/rundown-actions.component.ts
RasmusAlbrektsen: Implemented in: https://github.com/tv2/alba-web-client/pull/457/commits/3dda0b522545fdc7adc71d6d50f5ed6ef10f67e0

rana-tv2 (approved):
🚀

PR #468, opened 2025-12-19
replace deprecated ngx-colors package with custom color picker implementation and removed dependency to angular-animations

rana-tv2:
## Change description
This PR is an effort for replacing the deprecated `ngx-colors` package with a self-rolled color picker implementation and removing the dependency on `@angular/animations`.

1. Replaced ngx-colors in ColorPickerComponent:
    * Removed `NgxColorsModule` and replaced it with `@angular/cdk/overlay` (OverlayModule) to manage the color picker dropdown.
    * Implemented a custom dropdown UI that displays the predefined color palette and a custom color picker using the native `<input type="color">`.
    * Updated the component logic to handle opening/closing the overlay and selecting colors.
    * Updated the component styles to style the new dropdown panel where I tried to replicate the `ngx-colors` style.
    * after running into the error `src/app/shared/components/color-picker/color-picker.component.scss exceeded maximum budget. Budget 4.00 kB was not met by 1.34 kB with a total of 5.34 kB` I had to move the delay calculation form the scss file into the html file.
    
2. Removed Dependencies:
    * Removed ngx-colors and `@angular/animations` from `package.json`
3. Updated Application Bootstrap:
    * Removed `provideAnimations()` from `main.ts` to completely remove the Angular animations dependency from the application startup.

_side note: I'm not the most seasoned front-end developer so all inputs, comments and critics are welcome_ 😄 

https://github.com/user-attachments/assets/1344fbc7-4be4-4cb8-9a68-5e45c30330c5

## Checklist


src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+@keyframes slideInItem {
+  from {
+    opacity: 0;
+    transform: translateY(-10px);
+  }
+  to {
+    opacity: 1;
+    transform: translateY(0);
+  }
+}
```
KvelaGorrrrnio: These two animations are near identical - the only thing separating them are the starting position for translateY. You could combine these and control the starting position with a CSS variable.
It could look like:
```scss
  @keyframes verticalSlideInWithFadeIn {
    from {
      opacity: 0;
      transform: translateY(var(--slide-down-fade-in-translate-y));
    }
    to {
      opacity: 1;
      transform: translateY(0);
    }
  }
}
```
Then the animatable item can use it like:
```scss
.fade-in-downward {
  --slide-down-fade-in-translate-y: -20px;
  animation: verticalSlideInWithFadeIn 0.3s ease-out;
}

.fade-in-upward {
  --slide-down-fade-in-translate-y: 100px;
  animation: verticalSlideInWithFadeIn 0.6s ease-out;
}

File: src/app/shared/components/color-picker/color-picker.component.scss [hunk 2]
```diff
@@ -56,3 +56,159 @@
...
+
+::ng-deep .color-picker-panel {
```
KvelaGorrrrnio: Is `::ng-deep` necessary for this selector? I tried removing it and did not see anything change :)

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+          box-shadow: 0 0 2px rgba(0,0,0,0.5);
+        }
+      }
+
+
+      .circle.wrapper.color {
+        background-image: linear-gradient(45deg, #ccc 25%, transparent 25%),
+          linear-gradient(-45deg, #ccc 25%, transparent 25%),
+          linear-gradient(45deg, transparent 75%, #ccc 75%),
+          linear-gradient(-45deg, transparent 75%, #ccc 75%);
```
KvelaGorrrrnio: What does this do? 😃

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+  to {
+    opacity: 1;
+    transform: translateY(0);
+  }
+}
+
+::ng-deep .color-picker-panel {
+  $radius: 5px;
+  
+  &.opened {
```
KvelaGorrrrnio: As the color-picker modal is inserted into the DOM with the `opened` class, then the class can be removed and this styling can be moved to the parent selector.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+      }
+      
+      .circle.wrapper {
+        position: relative;
+        margin: 0 5px;
+        margin-bottom: 5px;
+        flex: 34px 0 0;
+        animation: slideInItem 0.3s ease-out backwards;
+      }
+    }
```
KvelaGorrrrnio, line 126: From what I can see, then there are more classes and variations than there are permutations. I think this is partially caused by how the preset colors are highlighted with help from extra HTML tags.
By utilizing pseudo elements, we can simplify both this SCSS and the HTML that it styles.
The intended approach is to create a pseudo element `::before`, that makes the visual representation of "selected circle" (the white border with an inset. Alternatively it can also be achieved by insetting a box-shadow. In the snippets below I am using the pseudo element approach.

The HTML at https://github.com/tv2/alba-web-client/pull/468/changes#diff-ba27cc7ee31dca242512f6b55c61598058788fd14f60676ec93e037b9306662bR47-R57 can be reduced to:
```html
<div
  class="color-circle"
  [style.animation-delay]="(i + 1) * 0.03 + 's'"
  [style.background]="color"
  [class.selected]="value === color"
  (click)="selectColor(color); close()"
>
</div>
```

The important thing here is the class `color-circle` which we will use as the base for styling the color circle and the pseudo-element for indicating color selection.
We use the conditional class `selected` evaluated by `value === color`, to activate the pseudo element. This is done with the selector `&.selected::before` in the SCSS snippet below.

```scss
$color-circle-diameter: 34px;
$selected-color-circle-inset: 4px;
$selected-color-circle-thickness: 2px;

.color-circle {
  position: relative;
  height: $color-circle-diameter;
  width: $color-circle-diameter;
  border-radius: 100%;
  cursor: pointer;

  --slide-down-fade-in-translate-y: -10px;
  animation: verticalSlideInWithFadeIn 0.3s ease-out backwards;

  &.selected::before {
    position: absolute;
    display: block;
    top: $selected-color-circle-inset;
    right: $selected-color-circle-inset;
    bottom: $selected-color-circle-inset;
    left: $selected-color-circle-inset;
    content: '';
    border: $selected-color-circle-thickness solid white;
    border-radius: 100%;
  }
}
```

The SCSS variables at the top of the snippet is positioned in the parent selector `.colors` but is omitted in the snippet for simplicity.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+
+    .manual-input-wrapper {
+      $height: 45px;
+      display: flex;
+      margin: 15px;
+      font-family: sans-serif;
+
+      p {
```
KvelaGorrrrnio: Paragraph (`p`) tags are used for paragraphs of text, where here it is used for a label. If smaller text sequences should be styled, then `span` is ideal. But in this case, as it is a label, the tag `label` can be used as well :)

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+        margin: 0;
+        margin-right: 5px;
+        text-align: center;
+        font-size: 10px;
+        letter-spacing: 1.5px;
+        text-transform: uppercase;
```
KvelaGorrrrnio: To keep the label styling consistent in the application, I suggest that the `text-transform: uppercase;` statements are removed.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+      $height: 45px;
+      display: flex;
+      margin: 15px;
+      font-family: sans-serif;
+
+      p {
+        margin: 0;
+        margin-right: 5px;
+        text-align: center;
+        font-size: 10px;
```
KvelaGorrrrnio: A convention we try to enforce is for text-sizing and input fields to use `rem` units instead of pixels, as this will allow for better scaling across devices and user settings. We often use pixels in the rundown view, as we have more strict requirements on resolution and information density.

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+        letter-spacing: 1.5px;
+        text-transform: uppercase;
+        line-height: 48px;
+        width: 30px;
+        user-select: none;
+        color: #595b65;
```
KvelaGorrrrnio: A thing that we strive for as a team, as much as it is possible, is to use the defined CSS color variables defined in the project. The motivation for it is to keep the color scheme consistent and add a bit of friction for when we need to new colors.
This comment applies to the other color values added in this PR :)

src/app/shared/components/color-picker/color-picker.component.scss
```diff
@@ -56,3 +56,159 @@
...
+      }
+
+      .g-input {
```
KvelaGorrrrnio: What does g-input mean/stand for? :)

KvelaGorrrrnio (approved):
Nice work! It is a nice and functional solution :+1: I have a suggestion for changing the UX and UI a bit to overall match other Alba components a bit more.

I have made a demo which includes most of my comments: https://github.com/tv2/alba-web-client/pull/469.
The PR does not look at using CSS variables for color codes and using `rem` over `px`.

RasmusAlbrektsen:
What is the status here? 
If we are not close to merging this i suggest we just take it down, but if you feel like making the changes suggested you are more than welcome :-)

KvelaGorrrrnio (approved):
Looks good :rocket:

PR #469, opened 2026-01-02
Suggestion: Moves manual color selection to top of color picker and the presets to the bottom

KvelaGorrrrnio:
#468 removes the (deprecated) Angular animations library and the color picker that depended on it with a homemade one.
This PR tries to make some UI/UX changes that takes the color picker from the previous library style to the current Alba settings style and improve on some of the aspects that the previous color picker library had.

One of the things I wanted to achieve with this suggestion is to make the manual color selection and the current color in the color picker modal stand out as the primary element and make the presets secondary. A part of achieving this is to fixate the color selection positions in logical positions for easy and fast operation.

## Change description

- Simplifies the HTML and CSS for the color picker presets, by using a pseudo element, instead of an element, for indicating color preset selection.
- Renames CSS classes to have more context - e.g.:
  - `.color-picker` &rarr; `.color-picker-input-field`
  - `.color.circle`&rarr; `.color-circle`
  - `.color-picker-panel.opened` &rarr; `.color-picker-modal`
- Removes duplicate preset color (`#370020`).
- Scales color presets on hover and selection.
- Combines CSS key frame animations into one, where the start position is controlled with the CSS variable `--slide-down-fade-in-translate-y`.

https://github.com/user-attachments/assets/1707b0a7-4ee8-41c3-9707-e9717b07b765

> [!note]
> This PR has not addressed streamlining of CSS units and usage of color codes and color variables.

## Checklist


rana-tv2 (approved):
🪨

RasmusAlbrektsen (approved):

